PR #211, opened 2014-11-11
Added `scoop update --preview` and created `scoop info`..

kodybrown:
I don't know what you think about `scoop info`.. I like it - especially as more and more cool new apps are added to the main bucket. I've found myself searching the tubes many times to figure out what an app is!

If you like it, I would _really_ appreciate any suggestions on the output format/style/layout of `scoop info app`.


deevus:
After checking this out and having a play I think `scoop info` is cool. I think it could be useful however it feels non-essential if we're trying to keep the main repo simple.

I'm not convinced about `scoop update --preview <app>` though, since running `scoop status` tells you the same thing.

However, I do think this raises a point about the extensibility of scoop. I have touched on this before when I made `scoop create` but I shall digress anyway. Scoop runs any commands called `scoop-*` in the libexec folder. I think that it would be cool if scoop could be extended in such a way that users could install new commands as they please. 

I have created an issue here to discuss: #304

deevus:
It's been 5 months with no more replies from @kodybrown so I'm leaning towards closing this.

kodybrown:
I'm sorry.. Easter weekend was busy with church and family activities and
I'm just now checking my personal email from mid-last week.. closing it is
fine btw.

I really wanted a way to search what each package/app is from the
commandline. What I'd really like is to see a short blurb next to the
search results, for instance (sorry, from memory, on my phone):

````
    PS> scoop search

      rsync (1.2)                 remote sync tool blah blah..
      ...
      sys internals (20150320)   sysinternals suite of awesome utilities.
      xylophone (3.21)       whatever this does..
      ...
      neko (x.y)                  I still can't remember this..
      haxe (a.b.c)               some kind of language or other..
````

Maybe it's just me, but I can't remember all these cool utils and what they
do. What i stated before was a start in this direction...

Issue #192, opened 2014-11-03
OpenJDK vs OracleJDK


scottwillmoore:
Both the openJDK and OracleJDK are both similar in features, however many developers lean towards the closed source OracleJDK for numerous reasons.
- It is commercially tested by Oracle
- Most home computers use Oracle's JVM, and while even the openJDK can compile and run applications on Oracles JVM, using the OracleJDK with the OracleJVM may reduce bugs?
- It has several tools and libraries not availiable in the openJDK

It is for these reasons most developers choose the OracleJDK over the open source cousin.  Unfortunately scoop does not offer the OracleJDK due to its policy of open source projects, however many of the open source projects depend on a JDK.
- Groovy, Scala, Kotlin (language)
- Ant, Maven, Gradle (build tool)
- Probably many more ?

It would be good if we could include the languages and build tools above, without forcing developers into using the openJDK or manual configuration/scripts to hold scoop apps together.

What is the thoughts of others on this issue?  Could we have conditional dependencies perhaps, with a choice between the two.  The OracleJDK manifest could be put in another repository so that scoops policy still applies?

PR #237, opened 2014-12-18
Remove openjdk dependancy for ant

sestegra:
In fact, when installing oraclejdk, openjdk conflicts with oraclejdk

### DIFF
--- a/bucket/ant.json
+++ b/bucket/ant.json
@@ -8,6 +8,5 @@
     "env_add_path": "bin",
     "env_set": {
         "ANT_HOME": "$dir"
-    },
-    "depends": "openjdk"
+    }
 }


deevus:
@lukesampson Possible feature of ORing dependencies? 

Something like `"depends": "openjdk|oraclejdk"`.

scottwillmoore:
We've already discussed this possibility, however decided against it because of a few issues. First the `oraclejdk` is located in the `scoop-extras` bucket which is not installed with scoop, which I guess would add complexity to the scoop code in order to handle multiple dependencies when they may not even exist in the main bucket. Therefore something such as `"depends: "openjdk|oraclejdk"` may have to become `"depends": "openjdk|scoop-extras:oraclejdk"`. So now the issue not only becomes version ORing, but also checking different buckets for dependencies and it may get messy later on (I am not against this idea however).

Also I believe @lukesampson also mentioned that the idea of scoop is so that scoop will never stop when installing multiple packages to ask a user a question. If a list of packages is supplied as arguments it should just work through all the packages until the process has been completed. I believe this could also be solved by potentially including arguments such as `scoop install ant:oraclejdk` which could also be end up complicating scoop. (Again I am not against the idea however).

I believe @lukesampson and I talked about the dependencies while working on `oraclejdk` in #192. So you may be able to find some extra information there. I guess we will have to decide whether an actual software solution should be devised or whether we can work around it?

deevus:
I think using `scoop reset` as discussed in #192 is workable as long as proper documentation is available.

PR #241, opened 2014-12-20
Global reset

sestegra:
Add global option to reset command


gutierri:
@lukesampson ?

lukesampson:
To me, reset doesn't make much sense for global apps. Global apps are for every user on the system. The reset command is for switching between different versions of an app, and that could cause chaos if an administrator is switching versions on everyone. I think this is the sort of thing you want to control just for yourself as a developer.

deevus:
I'd be interested to know the use case in which wanting to reset a global app comes up. Perhaps @sestegra can shed some light on this?

c33s:
should be handled some way, currently installing ruby & ruby19 globally works but switching wont work. you should either allow installing both and allow global switching or disallow install both versions

PR #261, opened 2015-01-18
Go v1.4.1

dre1080:


lukesampson:
It looks like you started with an old branch where the Go version was 1.3.2—can you please pull the latest changes with git and then redo this PR? The current version of Go that we have is 1.4.

PR #263, opened 2015-01-23
update-nim

scottwillmoore:
Updated nim to the latest version. Also under recent updates, 'nimrod' was renamed to 'nim'.


lukesampson:
Awesome :+1:

PR #268, opened 2015-02-01
Updated groovy.json for the groovy 2.4.0 release

jlchoike:
FYI - I used http://hash.online-convert.com/sha256-generator to generated the SHA 256 hash for groovy-2.4.0.zip. I think this should be correct, because I generated the same hash value for 2.3.7.zip using this site. Please let me know if you have any questions. Thanks!

### DIFF
--- a/bucket/groovy.json
+++ b/bucket/groovy.json
@@ -1,10 +1,10 @@
 {
     "homepage": "http://groovy.codehaus.org/",
-    "version": "2.3.7",
+    "version": "2.4.0",
     "license": "Apache 2.0",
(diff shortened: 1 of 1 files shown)


scottwillmoore:
Thanks! Another way you can generate a hash is by installing the scoop package `shasum`. Also you can easily test your packages by copying the `package.json` into `~\AppData\Local\scoop\apps\scoop\bucket\` folder and running `scoop install package`, if it fails you can run `scoop uninstall package` and repeat this proccess until it is working.

PR #290, opened 2015-03-11
scoop-create command to help with creating app manifests

deevus:
In the same vein of the Homebrew `brew create` command I was entertaining the idea of something similar for scoop. 

I wouldn't consider this complete or foolproof. Just would like to discuss the idea.

### DIFF
--- a/libexec/scoop-create.ps1
+++ b/libexec/scoop-create.ps1
@@ -1,0 +1,55 @@
+# Usage: scoop create <url>
+# Summary: Create a custom app manifest
+# Help: Create your own custom app manifest
+param($url)
+
+function create_manifest($url) {
+    $manifest = new_manifest
+
+    $manifest.url = $url
+    $url_parts = parse_url $url
+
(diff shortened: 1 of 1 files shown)


lukesampson:
I like the idea. It's a nice simple initial implementation too, and I don't think it can hurt to have a guided way to put together a manifest.

deevus:
Well there you go :)

deevus:
In the same vein as git-* commands, what do you think of scoop being extended so that scoop-*.ps1 files on the path would be picked up as scoop commands?

lukesampson:
I'm not sure. It might add vulnerabilities and degrade performance if the path has to be searched. Also the maintenance cost of supporting plugins like this probably isn't negligible.

Do you have a good example scenario that would make it worthwhile?

deevus:
I don't currently have a specific example other than scoop-create but you were happy to merge that in.

lukesampson:
OK cool.

ntwb:
Sweet :+1:

PR #292, opened 2015-03-12
BFG Repo-Cleaner

deevus:
> Removes large or troublesome blobs like git-filter-branch does, but faster. And written in Scala

BFG is the first manifest I've started using scoop-create :)


lukesampson:
Very cool :+1:

Issue #304, opened 2015-03-26
Extending scoop with custom commands


deevus:
I propose that scoop be extended so that can run any commands called `scoop-*`. 

I first thought that scoop could just walk `$env:PATH` checking for commands matching `scoop-*` but I'm afraid that this might affect performance too much. My next thought is that instead we could just walk the `$env:scoop/shims` path which will be much quicker. 

Additionally scoop commands could be installed using scoop itself eg. `scoop install scoop-foo` by having a manifest with `"bin": "scoop-foo.ps1"`. Like magic!

I'm going to riff on this a bit this afternoon.

Thoughts?

lukesampson:
I agree it does seem cool, but I'm not sure what the use case is for extensibility. I think we should be careful here not to add extensibility features and lose clear focus and simplicity. I don't really see the benefit of people independently implementing commands of varying quality, perhaps with overlapping functionality.

Also, the `scoop info` example doesn't seem very useful to me, but if we were to thrash it out and decide it was useful, then it would seem to me that we should add this as a core command. If it can't prove it's usefulness then it's a win for simplicity by keeping it out.

Maybe there's a better example of a command that would make extensibility really useful, but I can't think of one.

deevus:
The point I think is that this way keeps the core of scoop simple but allows people to go crazy if they want. @kodybrown for example appears to be running scoop from his own private fork. 

Personally I don't want fragmentation like that to become commonplace, which is the main reason I'm putting this forward.

deevus:
I have a bucket with a test command at https://github.com/deevus/scoop-commands

lukesampson:
OK. I'm still not sure that this is a good direction, but your implementation looks nice and simple and I guess letting people do their own thing isn't so bad, if they really want to. So yeah, let's do it then.

PR #305, opened 2015-03-26
Added support for custom commands

deevus:
Custom commands `scoop-*` that are in the `$env:scoop\shims` directory
can now be run using scoop as well as the built-in commands

### DIFF
--- a/lib/commands.ps1
+++ b/lib/commands.ps1
@@ -1,5 +1,7 @@
 function command_files {
-	gci (relpath '..\libexec') | where { $_.name -match 'scoop-.*?\.ps1$' }
+	(gci (relpath '..\libexec')) `
+        + (gci "$env:scoop\shims") `
+        | where { $_.name -match 'scoop-.*?\.ps1$' }
 }
 
 function commands {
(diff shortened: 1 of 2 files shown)


deevus:
Fixes #304

deevus:
I have a bucket with a test command at https://github.com/deevus/scoop-commands

lukesampson:
OK cool. Is this good to go, or does it need some testing still?

deevus:
If not ready it's close. I tested everything I could think of and couldn't
break it.

lukesampson:
That's good enough for me.

lukesampson:
Sorry, had to revery—I got lots of these errors when I type `scoop help`:

```
gci : Cannot find path 'C:\shims' because it does not exist.
At C:\Users\luke\appdata\local\scoop\apps\scoop\current\lib\commands.ps1:3
char:12
+         + (gci "$env:scoop\shims") `
+            ~~~~~~~~~~~~~~~~~~~~~~
```

lukesampson:
I think you just need to use `$scoopdir\shim` instead of `$env:scoop\shims`.

deevus:
Sure thing I'll look at it later tonight

PR #314, opened 2015-03-30
Added simplistic nightly support

### LINKED ISSUE #181: Nightly Support for Rust (and possible others)
Rust is an experimental programming language and as such is iterating very quickly, meaning that the recommended way to consume it on Windows is through the nightly mingw32 downloads. 

This can be seen here: http://www.rust-lang.org/install.html

As far as I can tell there doesn't seem to be anything stopping me from changing the rust manifest to use a nightly version. However the hash will be different every day, which makes setting the hash useless. 

Would there be room for a nightly flag (or some other flag) that would suppress the hash warning?

Here is my updated manifest:
```
{
	"homepage": "http://www.rust-lang.org",
	"version": "nightly",
	"license": "MIT/Apache 2.0",
	"depends": "gcc45",
	"url": "https://static.rust-lang.org/dist/rust-nightly-i686-w64-mingw32.exe",
	"innosetup": true,
	"bin": [
		"bin\\rustc.exe",
		"bin\\rustdoc.exe"
	],
}
```

Suppressing the hash warning might be a simple solution like `"nightly": true` or perhaps a more general solution such as `"hash_suppress_warning": true`

I'd be happy to look at it and make a PR if you'd be willing to go down that path.

deevus:
When an app versioned as "nightly" is installed (or updated),
the current date is appended to the version which changes the
version directory to nightly-yyyymmdd.

Fixes #181

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -1,5 +1,13 @@
+function nightly_version($date) {
+	$date_str = $date.tostring("yyyyMMdd") 
+	warn "this is a nightly version: downloaded files won't be verified"
+	"nightly-$date_str"
+}
+
 function install_app($app, $architecture, $global) {
 	$app, $manifest, $bucket, $url = locate $app
+	$use_cache = $true
+	$check_hash = $true
 
 	if(!$manifest) {
 		abort "couldn't find manifest for $app$(if($url) { " at the URL $url" })"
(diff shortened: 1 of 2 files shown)


lukesampson:
Looks good to me!

PR #320, opened 2015-04-04
Scoop update git

deevus:
Uses git to update scoop instead of downloading the zip every time

Issues
---------

* Changing repos/branch is not possible after initial update

Usage
---------

If you want to test this out do the following:

```powershell
# this is a custom install script - check the source out if you're worried
# all i changed is where the zip is downloaded from
iex (new-object net.webclient).downloadstring("https://gist.githubusercontent.com/deevus/864ce870042ae1235fe2/raw/9eae811a59c070a9f384b3cfb6b54f8f329be02c/scoop-install.ps1")

# update/status will check these settings
scoop config SCOOP_REPO "https://github.com/deevus/scoop"
scoop config SCOOP_BRANCH "scoop-update-git"

# will now use git to update
scoop update
```

Let me know if I've missed anything

### DIFF
--- a/bin/install.ps1
+++ b/bin/install.ps1
@@ -29,8 +29,6 @@ cp "$dir\_scoop_extract\scoop-master\*" $dir -r -force
 rm "$dir\_scoop_extract" -r -force
 rm $zipfile
 
-$null > "$dir\last_updated" # save install timestamp
-
 echo 'creating shim...'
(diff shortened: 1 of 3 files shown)


lukesampson:
Nice one! I tested it with a fresh install and it worked great. The code looks good, I can't see anything that you've missed.

deevus:
I'm glad :)

PR #331, opened 2015-04-09
Appveyor CI

deevus:
Continuous integration using [Invoke-Build](https://github.com/nightroman/Invoke-Build) and [AppVeyor](http://www.appveyor.com/)

The motivation is trying to automate the testing process so we can catch problems early and often.

### DIFF
--- a/.build.ps1
+++ b/.build.ps1
@@ -1,0 +1,43 @@
+. "lib/core.ps1"
+
+function Run-Test($test) {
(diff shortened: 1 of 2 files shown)


deevus:
This is not quite ready yet. I'm still working out the configuration to get the CI working properly. I have to push commits to get it to rebuild :(

lukesampson:
No problem.

PR #334, opened 2015-04-10
Add ninja build system

scottwillmoore:
Another build system for low level applications. Ninja is more like make instead of cmake, designed for lightning fast C/C++ builds.

### DIFF
--- a/bucket/ninja.json
+++ b/bucket/ninja.json
@@ -1,0 +1,10 @@
+{
+    "homepage": "http://martine.github.io/ninja/",
+    "version": "1.5.3",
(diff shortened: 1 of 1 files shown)


deevus:
This is already in lukesampson/scoop-extras

I don't see any reason why it can't be in the main bucket though. So perhaps we remove it from extras.

scottwillmoore:
Oh I didn't see it in the extras bucket. Thank you for pointing that out. Yep it should be removed from scoop-extras as this meets all the requirements to fit into the main bucket.

PR #336, opened 2015-04-11
Migrate testing to Pester

deevus:
I've started migrating the tests to use Pester, which is a proper testing framework written in Powershell. It can output in NUnit XML format which is compatible with AppVeyor. 

If you check the build results for this PR you can see passing/failing tests in the Tests tab. 

I've migrated several of the old tests to Pester and I'll try to complete them all before merging this in. Thoughts?


deevus:
@lukesampson Any idea why [unzip_old with small.zip](https://github.com/lukesampson/scoop/blob/pester-tests/test/Scoop-Core.Tests.ps1#L64) would fail?

[AppVeyor console](https://ci.appveyor.com/project/deevus/scoop/build/48-pester-tests)

deevus:
@lukesampson The leftover test files `install.ps1`, `msi.ps1`, `proxy.ps1`, `rawargs.ps1`, `vimrc.ps1` and `wrap.ps1` I haven't converted to Pester as they aren't using the `tests.ps1` library. 

Do you think they could be converted?

lukesampson:
Yeah don't worry about converting those leftover files, I think they were just to find out how out how other programs and OS work.

PR #337, opened 2015-04-12
Add manifest validation tests

deevus:
The manifest-validation test iterates through each json in the main
bucket to ensure that required fields (url, version) are not empty

### DIFF
--- a/test/Scoop-Manifest.Tests.ps1
+++ b/test/Scoop-Manifest.Tests.ps1
@@ -21,3 +21,15 @@ describe "manifest-validation" {
         }
     }
 }
+
+describe "parse_json" {
+    beforeall {
+        $working_dir = setup_working "parse_json"
+    }
+
+    context "json is invalid" {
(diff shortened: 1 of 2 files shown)


scottwillmoore:
Awesome, I remember several times in the past when scoop would produce errors when using commands such as `scoop search` which iterate over all the JSON manifests if a manifest was syntactically inccorect because someone had submitted a manifest missing a comma or quotation mark in a JSON manifest. Hopefully this picks up on manifest errors before they are published. :)

deevus:
Thanks. I'm not sure how much more I should add to this without complicating it too much.

scottwillmoore:
If it works and checks all the *required* attributes of a manifest as well as the syntax of a manifest (which I believe is tested by `$manifest | should not benullorempty`) then that is all it needs to do. Maybe just check that it works by removing a comma, or quotation mark from a json manifest to see whether it detects that it is broken.

PR #342, opened 2015-04-23
Renames function in lib\install.ps1 to not collide with other packages.

nengberg:
If you have another package installed, for example this well known package https://gallery.technet.microsoft.com/scriptcenter/Invoke-Locate-PowerShell-0aa2673a it wont work to install packages via scoop. 

The reason why it is not working is because the command locate seems to be reserved for that package and therefore collides with the function.

A fix for this is to rename a function called locate. What do you think?

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -5,7 +5,7 @@ function nightly_version($date) {
 }
 
 function install_app($app, $architecture, $global) {
-	$app, $manifest, $bucket, $url = locate $app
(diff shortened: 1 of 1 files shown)


deevus:
Not sure how I feel about this. It's a simple fix however should we need to rename all our functions with conflicts in future?

deevus:
I'm thinking we might be able to resolve this using Scope.

nengberg:
I agree with you, ignore this PR. :)

PR #344, opened 2015-04-24
Added JSON manifest for Grails SDK

jlchoike:


jlchoike:
@lukesampson How do you go about testing new manifests? I would like to add others to enable scoop to have the same functionality as GVM or Posh-GVM, because Posh-GVM does not work correctly (and there is not much documentation to explain common errors (and GVM does not support Windows).

lukesampson:
@jlchoike If you run `scoop help install` there's some info about how you can test the install from your file system.

```text
To install an app from a manifest on your computer
     scoop install \path\to\app.json

When installing from your computer, you can leave the .json extension off if you like.

Options:
  -a, --arch <32bit|64bit>  use the specified architecture, if the app supports it
  -g, --global              install the app globally
```

So if you're testing `grails.json` and it's in the working directory, you can just type `scoop install grails.json`. The `.json` on the end will just make sure it doesn't install `grails` from the official repo (once it's published here).

Also note you can use the `--architecture` switch if you have a manifest that supports 32-bit and 64-bit settings.

lukesampson:
This looks good to me. Thanks :+1:

PR #346, opened 2015-04-26
invoke-build.json: Invoke-Build 2.12.0

nightroman:
Invoke-Build 2.12.0 contains a number of fixes and improvements since 2.9.14. Also, some scripts are excluded from the package as redundant. Thus, "bin" in json is updated accordingly.


lukesampson:
Thanks Roman :smile:

PR #352, opened 2015-05-04
Docker Machine 0.2.0

sakai135:


lukesampson:
Cool, I haven't tried this but it looks good.

PR #356, opened 2015-05-09
Update nim to 0.11.2

moigagoo:
Removed nim_debug.exe (doesn't exist anymore), added nimble.exe (nim's package manager).


deevus:
Thanks

PR #359, opened 2015-05-10
Replace short function names with aliases

### LINKED ISSUE #351: Short internal function names may clash with aliases defined in a session
Original case. A build script for Invoke-Build contains the task *update*:

    task update {
        scoop update
    }

This task fails with an error

    Invoke-BuildExec : Cannot process argument transformation on parameter
    'Command'. Cannot convert the "update" value of type "System.String" to
    type "System.Management.Automation.ScriptBlock".

    At ...\scoop\apps\scoop\current\bin\scoop.ps1:12 char:42
    + elseif ($commands -contains $cmd) { exec $cmd $args }

It fails due to the conflict of `exec` commands. Invoke-Build defines an alias
`exec` for its internal function `Invoke-BuildExec`. On command name resolution
PowerShell searches aliases first. When scoop invokes `exec` then an existing
alias is called, not the internal scoop function `exec`.

Contrived problem case, easy to test right in the PowerShell console:

    function Invoke-Exec {"Invoke-Exec with '$args'"}
    Set-Alias exec Invoke-Exec
    scoop update

It works fine but it outputs

    Invoke-Exec with 'update System.Object[]'

That is scoop calls `Invoke-Exec` via the alias `exec`, not its own function as
it expects.

---

How Invoke-Build deals with the same issue (and got punished...). It uses classic
Verb-PrefixNoun for internal function names and defines short aliases to these
names for actual use. E.g.

    function Invoke-BuildExec ...
    Set-Alias exec Invoke-BuildExec

And then it uses `exec` everywhere. A user may have its own alias `exec`
defined in parent scopes or even globally, conflicts will not happen.

Verb-Noun names will unlikely conflict with user aliases. I know some people
dislike Verb-Noun names. Any names which unlikely suitable for aliases will do.
E.g. scoop internal functions `command_files`, `command_name`, `command_path`
are probably named safely. But functions `commands` and `exec` in the same
script are probably not, such aliases may be defined in a user profile or by
another tool which uses scoop (e.g. Invoke-Build via user tasks).

---

The suggestion for scoop is to review such short function names like `exec`.
Names with non-word characters (underscore, dash) seem to be less likely to
conflict.

scoop may continue to use short names as aliases, so massive renaming is not
necessary. Example, instead of:

    function exec($cmd, $arguments) {
        ...
    }

this may be used

    function exec_command($cmd, $arguments) {
        ...
    }
    Set-Alias exec exec_command
    
Then the rest of scoop code should work not changed. At the same time the
potential conflict with an existing alias `exec` is resolved.

P.S. I also explained this "trap" as [Too-simple-function-names](https://github.com/nightroman/PowerShellTraps/tree/master/Basic/Too-simple-function-names). I collect such cases there :)

deevus:
All one word function names have been renamed to avoid clashes
with aliases. Those functions are then aliased to their original
name.

Fixes #351

### DIFF
--- a/lib/commands.ps1
+++ b/lib/commands.ps1
@@ -4,9 +4,10 @@ function command_files {
         | where { $_.name -match 'scoop-.*?\.ps1$' }
 }
 
-function commands {
+function scoop_commands {
 	command_files | % { command_name $_ }
 }
+set-alias commands scoop_commands
 
 function command_name($filename) {
 	$filename.name | sls 'scoop-(.*?)\.ps1$' | % { $_.matches[0].groups[1].value }
@@ -27,8 +28,9 @@ function command_path($cmd) {
     $cmd_path
 }
 
-function exec($cmd, $arguments) {
+function command_exec($cmd, $arguments) {
     $cmd_path = command_path $cmd
 
 	& $cmd_path @arguments
 }
+set-alias exec command_exec
--- a/lib/config.ps1
+++ b/lib/config.ps1
@@ -1,12 +1,13 @@
 $cfgpath = "~/.scoop"
 
-function hashtable($obj) {
+function scoop_hashtable($obj) {
 	$h = @{ }
 	$obj.psobject.properties | % {
 		$h[$_.name] = hashtable_val $_.value		
 	}
 	return $h
 }
+set-alias hashtable scoop_hashtable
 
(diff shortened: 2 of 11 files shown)


deevus:
@lukesampson Please review these changes before merging them in. Just want to make sure you're happy with how I've named internal commands.

PR #360, opened 2015-05-11
A brute force fix for powershell alias problems

lukesampson:
This is a possible fix for #351. What it does:

1. Resets common Powershell aliases to their defaults in Scoop's script scope (e.g. ls, gcm). This was already being done with script-scoped functions, but this change does it with script-scoped aliases instead.
2. Prioritises local functions over user-defined aliases with the same name.
3. Doesn't require renaming functions to make them more obscure in the hope that they won't collide with user's aliases.
4. Gives more protection against invasive aliases.

On my machine this adds about 30ms execution time to each scoop command issued.

I'm going to think about this one before committing. Comments welcome.

### DIFF
--- a/bin/scoop.ps1
+++ b/bin/scoop.ps1
@@ -6,6 +6,8 @@ set-strictmode -off
 . "$psscriptroot\..\lib\core.ps1"
 . (relpath '..\lib\commands')
 
+reset_aliases
+
 $commands = commands
 
 if (@($null, '-h', '--help', '/?') -contains $cmd) { exec 'help' $args }
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -222,12 +222,11 @@ function reset_alias($name, $value) {
 		if($existing.definition -ne $value) {
 			write-host "alias $name is read-only; can't reset it" -f darkyellow
 		}
-		
-		return;
+		return # already set
 	}
 	if($value -is [scriptblock]) {
 		new-item -path function: -name "script:$name" -value $value | out-null
-		$value = "script:$name"
+		return
 	}
 	
 	set-alias $name $value -scope script -option allscope
@@ -235,7 +234,7 @@ function reset_alias($name, $value) {
 
 function reset_aliases() {
 	# for aliases where there's a local function, re-alias so the function takes precedence
-	$aliases = get-alias -scope global |? { $_.options -notmatch 'readonly' } |% { $_.name }
+	$aliases = get-alias |? { $_.options -notmatch 'readonly' } |% { $_.name }
 	get-childitem function: | % {
 		$fn = $_.name
 		if($aliases -contains $fn) {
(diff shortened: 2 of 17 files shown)


nightroman:
It's just a quick comment. I did not think of the `reset_aliases` approach carefully yet.

Is it going to resolve the original problem case of #351? I can see that `reset_aliases` deals with `-scope global`. In the problem #351 Invoke-Build defines the aliases not in the global scope. Also Invoke-Build calls may be nested so that aliases may be defined in several scopes.

lukesampson:
@nightroman it does solve the simpler test case you provided in #351 with `set alias exec Invoke-Exec`. It seems that `get-alias -scope global` matches aliases defined in the parent scope and above—I'm assuming that includes any aliases set by Invoke-Build that leak into Scoop's scope.

nightroman:
This test fails

```powershell
function test2 {
	Get-Alias test -Scope global
}
function test1 {
	Set-Alias test bar
	test2
}
test1
```

> Set-Alias : This command cannot find a matching alias because an alias with the name 'test' does not exist.

It works if `-Scope global` is removed.

lukesampson:
@nightroman yep, you need to execute a script to see aliases that were defined without `-scope global` become "global" from the point of view of the script.

```powershell
"get-alias test -scope global" > test.ps1
set-alias test write-host
./test.ps1
```

nightroman:
Yes, if you execute this code in the global scope, i.e. from interactive
console or using `powershell -File`.

If we put these three lines in a new script `test2.ps1` and invoke it without
dot sourcing then `test` is not found.

```powershell
@'
"get-alias test -scope global" > test.ps1
set-alias test write-host
./test.ps1
'@ > test2.ps1

# works
powershell -file test2.ps1
powershell -command '. .\test2.ps1'

# fails
powershell -command .\test2.ps1
```

It's because in first two cases the code is invoked in the global scope, so
that the alias *test* is created in the global scope and `get-alias -global`
finds it there.

In the 3rd case the code is invoked in a script scope and `get-alias -global`
cannot find it.

Perhaps I miss something obvious. Please ignore my worries if Invoke-Build case
from #351 works fine with this change. It's my problem to understand how it works.

lukesampson:
If you can provide a complete test case for Invoke-Build I'd be happy to run it and see if it works.

nightroman:
Please get the script [Invoke-Build.ps1](https://github.com/nightroman/Invoke-Build/blob/master/Invoke-Build.ps1).
It can be used alone without the package.

Create this build script *.build.ps1*, i.e. a standard PowerShell script with a task:

```powershell
task update {
    scoop update
}
```

Invoke it by InvokeBuild in PowerShell

```powershell
<path1>\InvokeBuild update <path2>\.build.ps1
```

Or, if InvokeBuild is in the path and *.build.ps1* is in the current directory, then simply

```powershell
InvokeBuild
```

lukesampson:
Thanks. The test did fail, so I've updated this to override all aliases, not just global ones.

deevus:
@lukesampson @nightroman I'm thinking we merge this in as the solution for now.

lukesampson:
@deevus thanks, done.

PR #370, opened 2015-05-25
Updated CMAKE to version 3.2.2

azabujuban:
I just changed the `filename` and the `checksum`, however I was wondering what's the proper testing procedure?


lukesampson:
Thanks for this. You can run `scoop install path/to/cmake.json` to test an install.

azabujuban:
Got it
Thanks!

PR #374, opened 2015-05-25
add dart dev channel

guillermooo:
I've just realized that it's also easy to add the dev version of dart, so here it is.


lukesampson:
Maybe this would be better in scoop-extras—the main bucket is just for current stable versions.

guillermooo:
Ah, makes sense. Will fix tomorrow.

PR #379, opened 2015-05-30
Fix #376 Robocopy is not always in PATH

### LINKED ISSUE #376: Robocopy is not always in PATH
Having been a personal advocate of Scoop to several of my friends, I have been told this issue multiple times, even though I haven't experienced this myself.

After installing Scoop and trying to `scoop install` something, Robocopy can't be found. 

I'll see if I can get them to come and give more information in this thread.

deevus:
Add ensure_robocopy_in_path method to core.ps1 which checks if
robocopy is available, then shims it if not.

### DIFF
--- a/bin/install.ps1
+++ b/bin/install.ps1
@@ -32,6 +32,7 @@ rm $zipfile
 echo 'creating shim...'
 shim "$dir\bin\scoop.ps1" $false
 
+ensure_robocopy_in_path
 ensure_scoop_in_path
 success 'scoop was installed successfully!'
 echo "type 'scoop help' for instructions"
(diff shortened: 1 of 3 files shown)


deevus:
@lukesampson Can you verify that this works? I don't have a clean VM in my current location to do so.

deevus:
@lukesampson Build is passing now :)

lukesampson:
I don't have a clean VM either, but I tested with a fresh user account with `C:\Windows\System32` removed from the path and it worked great. I'm going to merge this, but do you think it's worth adding  check to make sure that the robocopy.exe file exists, and aborting otherwise?

deevus:
It should definitely exist. I don't think we need to worry about checking for it at this point.

lukesampson:
OK cool.

PR #380, opened 2015-05-31
Add scoop-alias command

### LINKED ISSUE #377: Scoop Reinstall Software
After having an issue with spacemacs, I uninstalled and reinstalled emacs and it worked

It would be a good feature to have a simple script to reinstall something (unlike reset) using `scoop reinstall foo` which calls

```
scoop uninstall foo
scoop install foo
```

Which would be a snap since scoop could then install from cache

deevus:
Fixes #377

### DIFF
--- a/lib/commands.ps1
+++ b/lib/commands.ps1
@@ -20,8 +20,11 @@ function command_path($cmd) {
         # get path from shim
         $shim_path = "$scoopdir\shims\scoop-$cmd.ps1"
         $line = ((gc $shim_path) | where { $_.startswith('$path') })
-        iex -command "$line"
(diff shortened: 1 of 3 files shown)


deevus:
@lukesampson I opted to just put the ps1 file directly in shims. Do you see a problem with this? An alternative would be to create a `/scoop/aliases` directory and put the scripts there, then shim them.

lukesampson:
No, this looks good to me—I don't see a problem with putting the ps1 file directly in shims. Do you want to merge it in when you're ready?

deevus:
Let's do this!

PR #381, opened 2015-05-31
Added mediainfo-cli 0.7.73

azabujuban:
There are two versions: a GUI and a CLI version. The CLI one seemed more appropriate.

thanks
--oleg


lukesampson:
Thanks for this. I've just added the homepage and license info to the manifest.

PR #387, opened 2015-06-07
added ffmpeg 2015-06-07

azabujuban:


lukesampson:
Thanks :) I just modified this to fix the 32-bit install and correct the white space formatting.

azabujuban:
Great! Sorry about the 32 bit install - I could only test the 64 bit version on my machine. Is it possible to force a specific architecture?

lukesampson:
No problem. Yes you can do `scoop install bucket/ffmpeg -a 32bit` to force the 32-bit install.

PR #395, opened 2015-06-14
Fix broken Sourceforge URL

nilkesede:


deevus:
The old address still works. Is there a reason we should migrate to this new one?

lukesampson:
Yeah, merging this request would probably fix the problem for some people and break it for others.

Sourceforge mirrors are getting really flaky now, and they've disabled the download URLs that auto-redirect to the closest mirror.

We could look at writing a special case for SF urls to load the HTML and parse it to find the closest mirror , e.g. loading http://sourceforge.net/projects/sevenzip/files/7-Zip/9.20/7z920-x64.msi/download?nodownload=1 and looking for the `href` of the `<a class="direct-download">`.

deevus:
Is there any reason we shouldn't use the main download links from http://www.7-zip.org/download.html to move away from Sourceforge?

lukesampson:
I can't see any reason not to. They must have added those new links recently—nicely spotted.

deevus:
@nilkesede Are you able to update and rebase your PR to use the main download links from http://www.7-zip.org/download.html ?

deevus:
Fixed in e689df24dbfafc1f0137699c186fdf5b2332567a

PR #396, opened 2015-06-16
Let PyWin32 extension installer find this python

chidea:
These will fix the issue that pywin32 extension installer being unable to find python installed with scoop.
pywin32 installer : http://sourceforge.net/projects/pywin32/files/pywin32/Build%20219/
Reference : http://tech.valgog.com/2010/01/after-installing-64-bit-windows-7-at.html


lukesampson:
Thanks :+1:

PR #414, opened 2015-07-07
Update youtube-dl.json

nishanthkarthik:
Updated version to 2015.07.07

### DIFF
--- a/bucket/youtube-dl.json
+++ b/bucket/youtube-dl.json
@@ -1,9 +1,9 @@
 {
     	"homepage": "http://rg3.github.io/youtube-dl/",
     	"license": "Unlicense",
-    	"version": "2015.04.03",
-    	"url": "https://yt-dl.org/downloads/2015.04.03/youtube-dl.exe",
-    	"hash": "md5:1eb53c85ced7e75ac7a3c9edc5cedd5e",
+    	"version": "2015.07.07",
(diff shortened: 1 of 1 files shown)


lukesampson:
Cool :+1:

nishanthkarthik:
What if we write something that scrapes the version number from respective websites automatically? There is manual work involved for every program under scoop, but it would keep all software up to date. Can I start writing?

lukesampson:
There's already a [checkver](https://github.com/lukesampson/scoop/blob/master/bin/checkver.ps1) script that does what I think you want to do. There's some usage info [here](https://github.com/lukesampson/scoop/issues/97), and the [app manifest reference](https://github.com/lukesampson/scoop/wiki/App-Manifests) has some more info about how to add this to your manifest.

nishanthkarthik:
What I meant was,
say `gcc`, the current url in manifest is,
http://ufpr.dl.sourceforge.net/project/mingwbuilds/host-windows/releases/4.8.1/64-bit/threads-posix/seh/x64-4.8.1-release-posix-seh-rev5.7z
This 4.8.1 version number affects the url once a new update for gcc comes out. 

Why did I create this pull request to update the manifest? What if all manifests can be updated automatically?

nishanthkarthik:
Thanks, found it.

lukesampson:
Oh also, did you see @guillermooo's [script](https://github.com/lukesampson/scoop/issues/405) to update Dart manifest's automatically?

PR #426, opened 2015-07-16
upgrade to dart 1.11.1

devoncarew:
Upgrade to dart 1.11.1. This PR also changes the refs from /latest (had been 1.11.0) to a specific version. The contents of /latest will change every time there's a new release, and the checksum will no longer be valid. We now point to a specific version and use that version's checksum.

### DIFF
--- a/bucket/dart.json
+++ b/bucket/dart.json
@@ -1,5 +1,5 @@
 {
-    "version":  "1.11.0",
+    "version":  "1.11.1",
     "license":  "BSD",
     "homepage":  "https://www.dartlang.org/",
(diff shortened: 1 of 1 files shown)


devoncarew:
cc @guillermooo and @lukesampson

guillermooo:
Cool! LGTM. I'll try to update https://gist.github.com/guillermooo/f579fdb84aab4bc488da so that generated manifests point to specific versions from now on. That should make it easier to generate the manifests in the future.

lukesampson:
Sounds good, thanks for this :+1:

PR #431, opened 2015-07-20
Remove unused checkvers

iakio:


lukesampson:
Oh—why are the checkvers not being used any more?

iakio:
PHP5.2 and 5.3 are EOL, No more new release.
These versions are not on http://windows.php.net/download/ .

lukesampson:
OK, thanks :)

PR #437, opened 2015-07-25
Change encoding of generated shim files to ASCII

rivy:
* change file encoding of the basic shim to be in a normal ASCII character set

.# DISCUSSION

Within PowerShell, `echo STRING > FILE` generates a file encoded in UCS-2-BOM format (or "Unicode", in PowerShell terminology). UCS-2 later morphed into the similar UTF-16 encoding format, and, within PowerShell, is treated as a "fuzzy equivalent" encoding. Problematically, these UCS-2 files are misinterpreted by CMD, which expects files using the ASCII (or "extended-ASCII") character encoding. And, in general, the UCS-2 format has many drawbacks when interpreted outside of a narrow PowerShell window. So, in general, it's best to stick to pure ASCII (or UTF-8, *without BOM*) encoded files, if at all possible.

PowerShell is perfectly comfortable interpreting ASCII / UTF-8 (no-BOM) files for scripts or for other general input files. So, in this case, we now use `STRING | out-file FILE -encoding oem` instead of `echo STRING > FILE` and `STRING | out-file FILE -encoding oem -append` instead of `echo STRING >> FILE`.

NOTE: ASCII (not "extended-ASCII") and UTF-8 (non-BOM) encoded files are exactly equivalent.

NOTE: "Extended-ASCII" is poorly defined, requiring an associated Windows Code Page designation to firmly establish the encoding.

ref: [Unicode UTF and BOM] http://www.unicode.org/faq/utf_bom.html @@ https://archive.is/TKnyk
ref: [Unicode and NET] http://csharpindepth.com/Articles/General/Unicode.aspx @@ https://archive.is/czs89


rivy:
@lukesampson , I've used "OEM" encoding in this patch to match the encoding used elsewhere within the code. But I'm unsure as to why "OEM" would be a better choice than just "Default" (which uses the default ANSI code page). Since the code is dealing with only ASCII characters everywhere that the encoding is used, I don't think that there's any difference anyway. But I'm still left wondering why "OEM" was chosen over "Default".

deevus:
I do recall there being an issue when `Default` encoding was used, but I can't remember where at this point.

rivy:
If you remember why `Default` is an issue, let me know. I'm still curious what it could be.

This code shouldn't be an issue anyway, given that it's using `OEM`, and that it's not functionally different from the original. It only increases the readability of the resulting shims.

deevus:
Are you able to fix the conflicts, then I'll merge it in?

rivy:
Done.

deevus:
:+1:

PR #438, opened 2015-07-25
Add `.editorconfig` style rules

rivy:
* re-enforce requested project file formats and basic character-level coding style
* emphasize the Win/DOS requirement that BAT/CMD files *must* have CRLF line endings

.# Discussion

Using [EditorConfig](http://editorconfig.org) style rules helps devs to maintain the file formatting and style of the project, even while using different editors. This minimizes the semi-invisible, but, paradoxically, visually noisy, whitespace differences between commits. It can also help enforce file format requirements (such as the Win/DOS CMD shell requiring CRLF line endings for correct functioning of BAT/CMD files).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+# EditorConfig (is awesome): http://EditorConfig.org
+
+# * top-most EditorConfig file
+root = true
+
+# default style settings
+[*]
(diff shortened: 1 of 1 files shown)


deevus:
Good stuff :+1:

Issue #427, opened 2015-07-17
Can Scoop (or at least the apps) be made portable?


derrikcurran:
I'd like to be able to move the entire Scoop directory (including apps) to another machine without needing to worry about the shims pointing to the wrong locations. I have a development environment set up in a folder synced by Dropbox. It would be really nice to have every app I install via Scoop available on every computer I use just by adding the shims folder to my path.

lukesampson:
You can set the scoop install directory with `$env:SCOOP` [see here](https://github.com/lukesampson/scoop/issues/249).

If you can handle re-installing everything, you'd just uninstall scoop (`scoop uninstall scoop`), set `$env:scoop` to the new path, then re-install scoop.

If you've already installed apps under the default directory and you want to keep them, you'll need to update the shims after you move the directory. These are the steps I'd take:

1. Move `~/appdata/local/scoop` to the new location
3. Set $env:SCOOP to point to the new location (see #249 on how to persist this across sessions).
2. Add the new `shim` path to your path (and remove the old one)
3. Update the scoop.ps1 shim to point to the correct location
4. For each installed app (you can use `scoop list`) you need to run `scoop reset [app]` to fix the shim.

derrikcurran:
Thanks. I did get that far (though `scoop reset [app]`) hadn't occurred to me.) I suppose I can just write a script to do all that each time I switch machines. Would it be feasible to just generate the shims based on $env:SCOOP?

lukesampson:
I was thinking you'd make your dropbox folder path the same, so that you'd only have to do this once. It's safer and better performance to have the shims not use the environment variable.

derrikcurran:
Ah ok. The problem is that I'm forced to put Dropbox on different drive letters depending on the machine. Now that I think of it, this can be solved with symbolic links. Maybe it's too narrow a use case to mess with the shims. That said, It might be worth noting that using relative paths in the shims seems to work well. e.g. `$path = $PSScriptRoot + "\..\apps\vim\7.4\vim.exe"`.

deevus:
I've personally looked at this with a version of Scoop running on a flash drive. If shims were relative, it would be much easier. 

@lukesampson Do you see any problem with shims being relative when generated by Scoop?

lukesampson:
I think that would be ok if shims are relative.

lukesampson:
Would probably need to update `scoop which` as well though.

thoughton:
[...]

PR #443, opened 2015-07-27
Make shim paths relative

deevus:
Shim paths are now relative when created, and `scoop-which` was modified to work within this as well.

/cc #427


nightroman:
I am not sure it is related but would it be possible to allow manifests with relative paths to sources? Right now, if I am not mistaken, only full paths are supported.

Use case: I have some local tools with known locations and all I want is to install and uninstall them easily with scoop. It would be nice if I can keep a scoop manifest together with a tool archive (hence a relative path is needed).

PR #475, opened 2015-08-26
fix/improve Scoop-Alias and Scoop-Core tests

rivy:
* Scoop-Alias: remove erroneous output from dot-sourcing the source file
* Scoop-Alias: using "TestDrive:\" for mocked shimdir
  - reduced cleanup
  - improved robustness; excution is now independent of CWD
* Scoop-Core: corrected directory location code
  - improved robustness; execution is now independent of CWD


deevus:
Thanks :+1:

PR #479, opened 2015-08-28
add simple syntax checking for all PowerShell script/module files

rivy:
* checks all .ps1 and .psm1 files in the Project repository
* NOTE: this is just simple syntax checking, but it can catch a lot of simple mistakes


deevus:
This is great :+1:

Issue #451, opened 2015-08-08
Clear criteria for main bucket


deevus:
As mentioned in lukesampson/scoop-extras#112 ragel was added to extras since phantomjs and latex were already here. 

I thought this might be a good chance to reconsider any apps that could be better off in the main bucket, but I think we first need to be clear about what the criteria for the main bucket is.

Here's what I gather apps need to be:
- Open source (and have a license)
- Command line (not mutually exclusive)
- Portable (no installer required)
- No registry (generally no exceptions)

Anything I missed?

lukesampson:
Stable versions only

deevus:
@lukesampson You mentioned loosening up on the criteria for the main bucket. Should we update the documentation?

lukesampson:
Yes—I'm just not sure where the documentation is that you're referring to.

These are the latest requirements that I used in the [scoop-extras README](https://github.com/lukesampson/scoop-extras):

1. a reasonably well-known and widely used developer tool
1. the latest stable version of the program
1. a fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

I think that is relaxed enough.

Requirement 1 is a bit of a judgement call. I just don't think we should include stuff you whipped up on the weekend and you're still trying to find an audience for.

Requirement 3 kind of implies that portable installs are desirable. We have to allow for Python which only distribute an installer.

deevus:
I can perhaps go in scoop under a contributions section of the wiki, or in
the README

ShanConn:
I was browsing the winget repo and yep, msft is still as awful as always.

EqualDust:
love scoop, and msft even with a fairly standard install of calibre portable manages to get the install/uninstall borked. Gotta love that we as a community manage to ship better tools than msft paid professionals. https://github.com/microsoft/winget-cli/issues/1750

PR #482, opened 2015-08-30
add `pcregrep` to main bucket

rivy:
* command-line perl compatible grep
* see [pcre.org](http://pcre.org)
* BSD license; meets all criteria in issue #451


deevus:
Good stuff :+1:

Issue #481, opened 2015-08-30
`scoop update` causes errors if the scoop CMD shim is modified


rivy:
When `scoop update` is executed from the CMD line, the scoop CMD shim (`scoop.cmd`) is used to initiate execution, calling the scoop application. The scoop application then performs the update (which regenerates the CMD shim) and terminates. Upon exit from the scoop application, control flow returns to the newly regenerated CMD shim and execution continues to a final termination.

This process is most often completely benign because the CMD shim is usually regenerated with an identical copy. So, when control flow returns to the CMD shim and execution resumes, it continues along the known, expected path. However, if the regenerated CMD shim is *not* identical, control flow returns to an alternate location within the script, leading to a high likelihood of an unplanned and error-prone execution path on the way to final termination.

Control flow is well-defined when CMD shells out to another executable (.COM or .EXE) or calls another BAT/CMD script. The CMD shell reads and parses the hand-off statement, saving a pointer to the character location just past the end of the statement. Then the CMD shell executes the hand-off statement, transferring flow of control. When the called executable/script exits, control flow returns to the original script, reading and executing from the saved character location within the script file. Testing shows that if the hand-off statement is enclosed within a parenthesis statement block, that block is fully read, parsed, and kept in memory for execution upon return of control flow; flow of control is returned to the character location at the end of that block. Testing also shows that the location "end of statement" includes any single CRLF trailing the statement/block, if present.

Given this process by the CMD shell, it is extremely difficult to plan a well defined execution path after modifications are done to an executing script, since it is dependent on the modification and the original source file execution path. Creating a generic, well-functioning update method is essentially impossible unless we simplify the issue by disallowing return of execution to any newly modified code path. This can be accomplished in one of two ways: 1. handing off flow of control with no return by executing another BAT/CMD script without using `call` (aka, proxy execution); or 2. constructing a parenthesis-block which contains the hand-off and ends the script within the block itself (eg, via `goto :EOF`; this is the only alternative possible for hand-off to executables which are not BAT/CMD scripts). Going forward, using either of these two methods will allow any in-use modification of the `scoop` BAT/CMD shim, without limitation.

For past versions of `scoop`, we don't have either of the two simplifications in place. But we can use the fact that past shims were all constructed to hand-off to `scoop.ps1` using the last command in the script to give us a known control flow return location, eg, the character position one past the original CMD shim file size.

[...]

PR #486, opened 2015-09-01
add immediate in-process environment updates for CMD shell

### LINKED ISSUE #485: `scoop install ...` from the CMD shell requires a process restart for finalization
When running `scoop install ...` (or `scoop update ...`), environment changes are performed and persisted to the USER or MACHINE store. Within a PowerShell shell, the PROCESS environment is also updated, allowing for immediate use of the installed application (from the same shell command line). But, when used from the CMD shell, `scoop install ...` does not update the PROCESS environment. Because of this, a new CMD shell must be executed to finalize the environment changes and allow use of the installed application.

However, BAT/CMD scripts can persist environment changes to the CMD shell PROCESS environment. By leveraging that ability, it is possible to pass environment updates back up to a calling CMD process via "environment piping". To use "environment piping", a temporary file is created and passed to its child process. The child process adds the environment changes to the temporary file and then exits, returning control to the parent process. The parent process then processes the temporary file, adding the changes to the current CMD shell PROCESS environment.

This does mean that the parent process (the `scoop.cmd` shim) must execute significant code after the child process (the main scoop application) returns. This impacts upon the method used to allow safe modification of the scoop CMD shim during a `scoop update` (see Issue #481 and the solution in commit c86cb30ec2b851967c9d44aa88202b3bf64bda8f). It is difficult, if not impossible, to construct a parenthesis-block statement which can load the environment changes into the PROCESS environment and then exit cleanly. Because of this, it is necessary to switch to the proxy execution method mentioned in Issue #481 for safe CMD shim updates.

rivy:
**note:** Without PR #484, for this update only, must run `scoop update` twice to get the final working product.

* for CMD shells, during install or update, pass environment changes back up the call stack into the calling CMD shell process
* adds feature parity with the PowerShell shell to the CMD shell
  - allows immediate use of newly installed packages
  - starting a new CMD shell is *no longer required* prior to use of an installed/updated application
  - eg, `scoop install xxx` can be immediately followed by using `xxx`
* uses an environment variable pipe

* fixes #485

* note: variables set within the adjunctive install script are not captured unless the install script uses the env() subroutine
* note: due to increased code complexity, proxy execution is now required for safe CMD shim updating

.# Discussion

When running `scoop install ...` (or `scoop update ...`), environment changes are performed and persisted to the USER or MACHINE store. Within a PowerShell shell, the PROCESS environment is also updated, allowing for immediate use of the installed application (from the same shell command line). But, when used from the CMD shell, `scoop install ...` does not update the PROCESS environment. Because of this, a new CMD shell must be executed to finalize the environment changes and allow use of the installed application.

However, BAT/CMD scripts can persist environment changes to the CMD shell PROCESS environment. By leveraging that ability, it is possible to pass environment updates back up to a calling CMD process via "environment piping". To use "environment piping", a temporary file is created and passed to its child process. The child process adds the environment changes to the temporary file and then exits, returning control to the parent process. The parent process then processes the temporary file, adding the changes to the current CMD shell PROCESS environment.

This does mean that the parent process (the `scoop.cmd` shim) must execute significant code after the child process (the main scoop application) returns. This impacts upon the method used to allow safe modification of the scoop CMD shim during a `scoop update` (see Issue #481 and the solution in commit c86cb30). It is difficult, if not impossible, to construct a parenthesis-block statement which can load the environment changes into the PROCESS environment and then exit cleanly. Because of this, it is necessary to switch to the proxy execution method mentioned in Issue #481 for safe CMD shim updates.

### DIFF
--- a/bin/scoop.ps1
+++ b/bin/scoop.ps1
@@ -1,15 +1,21 @@
 #requires -v 3
-param($cmd)
+param(
+    [parameter(mandatory=$false)][string] $__CMDenvpipe = $null,
+    [parameter(mandatory=$false,position=0)] $__cmd,
+    [parameter(ValueFromRemainingArguments=$true)][array] $__args = @()
+    )
 
 set-strictmode -off
 
 . "$psscriptroot\..\lib\core.ps1"
 . (relpath '..\lib\commands')
 
+$env:SCOOP__CMDenvpipe = $__CMDenvpipe
+
 reset_aliases
 
 $commands = commands
 
-if (@($null, '-h', '--help', '/?') -contains $cmd) { exec 'help' $args }
-elseif ($commands -contains $cmd) { exec $cmd $args }
-else { "scoop: '$cmd' isn't a scoop command. See 'scoop help'"; exit 1 }
+if (@($null, '-h', '--help', '/?') -contains $__cmd) { exec 'help' $__args }
+elseif ($commands -contains $__cmd) { exec $__cmd $__args }
+else { "scoop: '$__cmd' isn't a scoop command. See 'scoop help'"; exit 1 }
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -2,6 +2,8 @@ $scoopdir = $env:SCOOP, "~\appdata\local\scoop" | select -first 1
 $globaldir = $env:SCOOP_GLOBAL, "$($env:programdata.tolower())\scoop" | select -first 1
 $cachedir = "$scoopdir\cache" # always local
 
+$CMDenvpipe = $env:SCOOP__CMDenvpipe
+
 # helper functions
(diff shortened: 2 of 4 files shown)


lukesampson:
I'm not opposed to the idea of the .cmd shim refreshing the environment variables after execution, I just don't think this is the way to do it.

Better would be for the .cmd shim to call something like [this script from Chocolatey](https://github.com/chocolatey/chocolatey/blob/master/src/redirects/RefreshEnv.cmd) to reload environment variables after the shimmed program has been called.

I think this would achieve the same result, but it's much less complicated and doesn't require such broad and sweeping changes to the powershell code.

Or we could try to work out a way to call `SendMessageTimeout` (something like [this](http://stackoverflow.com/a/2479615/87453)) to refresh the environment. I had a go at this but couldn't get it to work.

PR #492, opened 2015-09-10
Corrected r.json and updated from R 3.1.1 to 3.1.3

alapala:
The R 3.1 binaries were moved to an `old` directory on CRAN mirrors, so I updated the URL. I also updated `r.json` to use 3.1.3 instead of 3.1.1.

Note that R 3.2 packages now seem to have `bin/R,1.exe` and `bin/R,2.exe` instead of `bin/R.exe` and `bin/Rscript.exe`, which I am not sure how to handle. Might be a good idea to shim the architecture-dependent executables instead.

### DIFF
--- a/bucket/r.json
+++ b/bucket/r.json
@@ -1,9 +1,9 @@
 {
     "homepage": "https://www.r-project.org",
-    "version": "3.1.1",
+    "version": "3.1.3",
     "license": "GPL2",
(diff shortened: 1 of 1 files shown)


lukesampson:
Thanks. Do packages still work after this change, or do we need to sort that out?

alapala:
Yup, R 3.1 still installs and is functional after this change

We do need to work out the installation of R 3.2, however. The executables come with different filenames that I don't quite understand.

PR #512, opened 2015-09-29
Shim askpass.exe

bcomnes:
Is there anything else required to create a shim?


lukesampson:
That should be all, thanks :+1:

PR #518, opened 2015-10-06
Create minisign.json

reelsense:
Minisign is a dead simple tool to sign files and verify signatures. Signature written by minisign can be verified using OpenBSD's signify tool.


lukesampson:
Cool, thanks :)

PR #521, opened 2015-10-11
Fix hashes for git and ruby

EmberQuill:
Scoop failed to update git and ruby for me due to hash mismatches. I manually downloaded the files that failed the hash check and checked them myself, then corrected the hashes in the JSON files.


lukesampson:
Thanks :)

PR #525, opened 2015-10-13
Updates NuGet to 3.2.0

berwyn:
NuGet also changed the location where they publish their binaries (and they're just a straight binary now), so this PR accounts for that change as well.


lukesampson:
Thanks :+1:

PR #546, opened 2015-10-29
Updates Rust to 1.4.0, adds MSVC variant

berwyn:
My one concern about this at the moment is that the MSVC version of the compiler only supports x86_64 in this release, is there anything that should be done about this?


berwyn:
My other question, would Rust's beta releases belong here or in the extras bucket?

lukesampson:
Thanks. Yes the extras bucket is the place for non-release versions.

PR #563, opened 2015-11-09
Create php56.json

zQueal:
Added `5.6.15 x86 VC11`.


zQueal:
This could have been avoided if I simply read. Sorry!

PR #564, opened 2015-11-09
Update Pshazz app to latest commit

hansek:
I'm just not sure about *hash* because as is written in https://github.com/lukesampson/scoop/wiki/App-Manifests I made SHA256 hash of URL, but as I tried SHA256 on original URL from app manifest then result was not `b8ee4ccffa719bbe97b47856457592f13ce4103e92b6bb41ff421a00f21bb08a`.

Did I do something wrong?

Thanks!

### DIFF
--- a/bucket/pshazz.json
+++ b/bucket/pshazz.json
@@ -1,8 +1,8 @@
 {
-    "version": "0.2014.11.25",
-    "url": "https://github.com/lukesampson/pshazz/archive/3b7b426320e42718b7ef3b0ebf9bbe06c6ee0430.zip",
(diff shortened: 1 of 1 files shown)


lukesampson:
Sorry, the documentation is a bit unclear—the SHA 256 hash needs to be over the entire downloaded file, not just the URL. I get 71825ce737c857c52203696a01f26776e8715014a392cb1a497961f7051d4da3 when I try it. Do you want to update the pull request?

hansek:
Updated, makes sense that the hash is CRC of file(s).

lukesampson:
Cool :+1:

hansek:
Thanks :+1:

PR #578, opened 2015-11-28
Added GNU which

RobBlackwell:
Please consider adding GNU which to the main Scoop bucket.


lukesampson:
It looks like the AppVeyor build is failing because there's no newline at the end of the file.

PR #584, opened 2015-12-03
Move PHP 5.6 to new bucket

belisarh:
Create new bucket for PHP 5.6 at current stable version PHP 5.6.16 to upgrade PHP main bucket to PHP 7.0.0


lukesampson:
Thanks for these :+1:

Issue #589, opened 2015-12-07
Issue while installing R 3.2.2


Cyianor:
The shims supplied in the manifest do not work.

    ~ $ scoop install r
    installing r (3.2.2)
    downloading https://cran.rstudio.com/bin/windows/base/R-3.2.2-win.exe...done
    checking hash...ok 
    unpacking innosetup...done
    creating shim for r
    can't shim bin\r.exe: file doesn't exist

There are two binaries, `R,1.exe` and `R,2.exe` in the `bin` directory. It seems like `R,1.exe` represents the 32bit version of `Rterm.exe` and `R,2.exe` the 64bit version. 

The other parts of R like `Rgui.exe` or `Rscript.exe` are in the folders `bin/i386` respectively `bin/x64`. Is there some way to create different shims depending on architecture? Or maybe better create two shims?

    bin/i386/Rterm.exe -> Rx32
    bin/x64/Rterm.exe -> Rx64

lukesampson:
You could add a `pre_install` step to the manifest that runs some powershell to copy `R,1.exe` or `R,2.exe` to `r.exe` based on the architecture (`[intptr]::size -eq 8` means 64-bit). Similar for the other files like `Rterm.exe`—you could copy the right version to `bin` so that the shim step can find them.

There's an example of a `pre_install` in the [python](https://github.com/lukesampson/scoop/blob/master/bucket/python.json) manifest.

Cyianor:
Works now.

PR #590, opened 2015-12-09
Fixed an issue with the 32/64 bit versions of R

Cyianor:
Addresses issue [#589](https://github.com/lukesampson/scoop/issues/589).

The R installer delivers the 32 as well as the 64 bit versions of R. I
added a pre_install to r.json to choose the correct version depending on
the user's architecture.

Additionally I added a new shim for 'rterm.exe' to have an alternative
way to start R without removing the 'r' alias in powershell.


Cyianor:
It's unclear to me why the `ensure_robocopy_in_path` test fails...

lukesampson:
Yeah I'm not sure why the test failed. Thanks for this :smile:

PR #610, opened 2015-12-23
Update to OpenSSL 1.0.2e.

thomaskonrad:
Updated the OpenSSL bucket to the current version. The old one is not available any more from the download site (it leads to an HTTP 404 error).


lukesampson:
Thanks :smile:

Issue #538, opened 2015-10-22
Install specific version


vpratfr:
Hi,

I ran `scoop install rsync` which install version 3.1.1. On my servers running under Debian stable, I have rsync 3.0.9 (latest available there) which has a compatibility issue with 3.1.1. 

How can I install 3.0.9 on my local computer ? Is there something like `scoop install rsync@3.0.9`?

deevus:
Hi @vpratfr 

There is not currently any way to install a specific version of a manifest in scoop. The current closest solution is to have a manifest specifically for that version. E.g. manifests for `git` and `git19`

You could create a `rsync30` manifest in this case to solve your problem

deevus:
@vpratfr Did you end up resolving your issue?

vpratfr:
I did that manually. I still think there should be a way to specify a version number for a package, this would be helpful to solve cases like that.

PR #614, opened 2016-01-03
Add node-v4.2.4

pongstr:
Since `node-5.x.x` is a bleeding edge, I think it will be great if scoop 
would also support an LTS version `node-4.x.x`.

#538


lukesampson:
Sounds good, thanks!

PR #640, opened 2016-01-16
add git-with-openssh

aaronzs:
add `git-with-openssh.json`. Modified from `git.json`. Just make the OpenSSH tools in the `git-for-windows` package can be searched in PATH.


lukesampson:
It looks like tests are failing because the file doesn't end with a newline—are you okay to fix that?

lukesampson:
Thanks!

PR #642, opened 2016-01-18
Start implementing support for installing specific versions

### LINKED ISSUE #538: Install specific version

deevus:
By leveraging `git rev-list` and `git show` I discovered that it was
possible to crawl the git history and get the json manifests for the
entire history of an app.

From there it was just a matter of parsing the json into an array of
manifests and patching support into `scoop install`. I also added a
rudimentary `scoop versions <app>` which will list all the versions of
the particular app from git history

I haven't tested all the edge cases yet so this is strictly WIP. 

Fixes #538

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -6,8 +6,43 @@ function nightly_version($date, $quiet = $false) {
     "nightly-$date_str"
 }
 
-function install_app($app, $architecture, $global) {
+function app_history($app) {
     $app, $manifest, $bucket, $url = locate $app
+    $manifest_path = resolve-path $(manifest_path $app $bucket) -ea silentlycontinue
+    if (!$manifest_path) {
+        return @()
+    }
+    $manifest_dir = split-path $manifest_path
+
+    pushd $manifest_dir
+    try {
+        # get commits containing $app
+        $commits = git rev-list --all -- ./$app.json
+
+        $manifests = $commits |% {
+            try {
+                git show $_`:./$app.json | convertfrom-json -ea silentlycontinue
+            }
+            catch {}
+        } | sort-object version -unique -descending
+    }
+    finally {
+        popd
+    }
+    $manifests
+}
+
+function install_app($app, $architecture, $global, $version) {
+    $app, $manifest, $bucket, $url = locate $app
+    if ($version) {
+        $history = app_history $app
+        $manifest = $history |? { $_.version -eq $version }
+
+        if (!$manifest) {
+            abort "couldn't find manifest for $app ($version)"
+        }
+    }
+
     $use_cache = $true
     $check_hash = $true
 
--- a/libexec/scoop-install.ps1
+++ b/libexec/scoop-install.ps1
@@ -14,6 +14,7 @@
 # Options:
 #   -a, --arch <32bit|64bit>  use the specified architecture, if the app supports it
 #   -g, --global              install the app globally
+#   -v, --version             install a specific version
 
 . "$psscriptroot\..\lib\core.ps1"
 . "$psscriptroot\..\lib\manifest.ps1"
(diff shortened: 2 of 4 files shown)


deevus:
@lukesampson Like to know what you think of this

lib/install.ps1
```diff
@@ -6,8 +6,43 @@ function nightly_version($date, $quiet = $false) {
...
+    try {
+        # get commits containing $app
+        $commits = git rev-list --all -- ./$app.json
+
+        $manifests = $commits |% {
+            try {
+                git show $_`:./$app.json | convertfrom-json -ea silentlycontinue
+            }
+            catch {}
+        } | sort-object version -unique -descending
```
deevus: Probably need to add smart ordering since version `1.1` will be ordered before `1.10`

lukesampson:
Wow, this looks great!

Would `sort_versions` from `lib/versions.ps` be useful for smart version ordering?

deevus:
@lukesampson That worked a treat!

deevus:
So with something like this at play, do we allow users to add older versions of an app? 

How do we determine which version to install? Let's say someone adds rsync 3.0.9 (#538). According to git, the HEAD will be that version, whereas a higher version exists in git history. 

Can we flag a version as broken somehow so that we don't even install it again?

deevus:
The alternative is having a separate bucket for older versions like homebrew called `scoop-versions` with manifests like `rsync309` and `python26` etc

lukesampson:
Yes, I don't think we need to retrospectively add manifests for older versions over the top of newer ones. I think it's good to limit this new feature to being installing old versions that you have *previously installed with scoop*.

The `scoop-versions` bucket sounds like a good workaround to allow the special cases.

Maybe flagging broken installs could be a separate feature. It seems like it would require a fair amount of maintenance without some sort of web service to support it.

lukesampson:
Hey @deevus, is this ready to merge or you're still working on it?

deevus:
It was working, but I'll have to rebase first

rivy:
If your interested, I've got a working version which handles multiple variants (eg, multiple version) of apps @ https://github.com/rivy/scoop/tree/tr-wip. The installation instructions are in the README.

It only sees current bucket manifests. I didn't want to search the history for fear that it would be impossible to parse out broken manifests.

It works by using application names of the form `[BUCKET/]NAME[@VERSION]`, eg:

    scoop bucket add rivy https://github.com/rivy/scoop-bucket
    scoop install rivy/gow@0.8.0 rivy/gow@0.8.0-min
    scoop list

I also added a 'disable' command which is an uninstall without application removal, allowing 'reset' to quickly restore functionality.

I had to refactor a lot of things, and I haven't split up the commits. So, it's still somewhat of a work-in-progress. But I think the code is fairly clean.

Let me know if you're at all interested.

deevus:
@rivy Do you think this solves any more than https://github.com/scoopinstaller/versions? 

I am interested in `scoop disable` and multiple arguments. Can you explain how they work?

rivy:
Unfortunately, I see no likelihood of any part being accepted. I'll just keep to my own fork.

DamianReeves:
Has this effort been abandoned?

felipecrp:
Also, some apps urls old versions changes when they release a new version. Take a look at R for an example.

https://cran.rstudio.com/bin/windows/base/R-3.3.2-win.exe
https://cran.rstudio.com/bin/windows/base/old/3.3.1/R-3.3.1-win.exe 

May be an 'old version sed' on the json would be able to convert those urls.

mr-moon:
Guys, please don't drop this.

deevus:
I think a better, more future-proof idea is to generate specific version manifests on the fly using the autoupdate feature. 

So lets say a user runs `scoop install nodejs@5.2.0` and that isn't the current version, we generate a manifest, then attempt to install that

PR #643, opened 2016-01-20
Feature/lessmsi install

deevus:
I have been having problems with `msiexec` on my machine where it just can't extract or install anything. Having discovered `lessmsi` I wanted to see if it would work in `msiexec`'s place, and it does

If you set `MSIEXTRACT_USE_LESSMSI` to `$true` using `scoop config`, it will use `lessmsi` over `msiexec`.

### DIFF
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -6,6 +6,15 @@ function requires_7zip($manifest, $architecture) {
     }
 }
 
+function requires_lessmsi ($manifest, $architecture) {
+    $useLessMsi = get_config MSIEXTRACT_USE_LESSMSI
+    if (!$useLessMsi) { return $false }
+
+    $(url $manifest $architecture |? {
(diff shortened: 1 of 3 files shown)


deevus:
@lukesampson Should I check that lessmsi is installed also?

lukesampson:
You could add something to `install_deps` in `depends` to check whether `lessmsi` will be required for an install.

deevus:
@lukesampson How's that?

lukesampson:
Something like

```powershell
function install_deps($manifest, $arch) {
    ...
    if(msi $manifest $arch -and get_config MSIEXTRACT_USE_LESSMSI) { $deps += "lessmsi" }
    ...
}
```

deevus:
@lukesampson That is essentially what I've done. Check the changes

lukesampson:
Sorry, misread your question. Looks good to me.

PR #653, opened 2016-01-21
Add aria2 utility

engelhro:
aria2 download utlity.

Proposal: add to main bucket, as this tool is comparable to the already contained wget utility (otherwise please move to the extra bucket instead)


lukesampson:
Sounds good, thanks :smile:

PR #657, opened 2016-01-24
Bumped Maven, removed dependency on OpenJDK 7

damnhandy:
In addition to bumping Maven to 3.3.9, I've removed the dependency on OpenJDK. Since Scoop's OpenJDK is stuck at 1.7.0, this gets wonky if you're also using Oracle's JDK 8 or 7. Rather than making Maven depend on a specific distribution of Java, I've added the note property to inform the user that they would likely need to install a JDK and how to do it.

### DIFF
--- a/bucket/maven.json
+++ b/bucket/maven.json
@@ -1,15 +1,15 @@
 {
     "homepage": "https://maven.apache.org/",
-    "version": "3.3.3",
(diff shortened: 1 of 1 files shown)


lukesampson:
Okay, that sounds reasonable. Thanks :+1:

PR #658, opened 2016-01-24
Add ability to search known buckets that haven't been added locally

deevus:
By using githubs api we can search known buckets that haven't been added
to the local scoop instance. When a local search doesn't return any
results, other known buckets will be searched via github. If the rate
limit has been reached (60 per hour as of impl) the search won't occur.


lukesampson:
Cool!

PR #659, opened 2016-01-24
Support for scoop install <bucket>/<app>

deevus:
- TODO: support install of same name apps from different buckets?

### DIFF
--- a/lib/buckets.ps1
+++ b/lib/buckets.ps1
@@ -29,7 +29,13 @@ function buckets {
     $buckets
 }
 
-function find_manifest($app) {
+function find_manifest($app, $bucket) {
+    if ($bucket) {
+        $manifest = manifest $app $bucket
+        if ($manifest) { return $manifest, $bucket }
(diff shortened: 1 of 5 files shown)


deevus:
This allows apps to have the same name across different buckets, but if you wanted to install both the stable and nightly versions of an app, I don't think it would work in the current state. 

So the next steps would be:

1) Allow multiple versions of the same app to be installed. First thought: store apps at `apps/<app>/<bucket>/<version>`?
2) Allow resetting of shims for same name apps: `scoop reset <bucket>/<app>` OR `scoop reset <app>` which toggles between versions

lukesampson:
This looks great! Yeah, I can see some people finding it useful to switch between apps of the same name from different buckets, that would be a good future enhancement.

kodybrown:
I love this feature! Thank you! :+1: 

I still would love to be able to install multiple versions of the same app, too. Which, could be done along the next steps you listed.

PR #664, opened 2016-01-28
Update path to scoop-extras

nikolaia:
Seems like you renamed the extras repo? Failed for me at least, had to add it manually with name and the complete adress.


deevus:
Yes, thank you!

PR #671, opened 2016-02-02
Update PHP 5.5 repository URL

sasindu555:
Current PHP 5.5 repository URL does not exist, It is now being updated to a new URL.
Old URL : http://windows.php.net/downloads/releases/php-5.5.30-Win32-VC11-x86.zip
New URL : http://windows.php.net/downloads/releases/php-5.5.31-Win32-VC11-x86.zip

### DIFF
--- a/bucket/php55.json
+++ b/bucket/php55.json
@@ -1,9 +1,9 @@
 {
     "homepage": "http://windows.php.net",
-    "version": "5.5.30",
+    "version": "5.5.31",
     "license": "http://www.php.net/license/",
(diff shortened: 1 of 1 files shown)


deevus:
@lukesampson What do you think about moving this (and others) to the versions bucket?

lukesampson:
I think that's a good idea, @deevus.

deevus:
@sasindu555 Can you reopen this PR here: https://github.com/scoopinstaller/versions ?

I'm going to remove these suffixed versions from the main bucket

sasindu555:
@deevus I created a PR in https://github.com/scoopinstaller/versions

Issue #678, opened 2016-02-07
Support for Startmenu Shortcuts


damnhandy:
For GUI applications that maybe found in the `scoop-extras` bucket, it would be helpful if a manifest could define a shortcut for the application in the Windows Start Menu. A manifest could define a shortcut in the manifest like so:

    "bin": [
         [ "app\\sourcetree.exe", "sourcetree" ],
         [ "app\\stree_gri.exe", "stree_gri" ]
     ],
     "shortcuts" : [
         [ "app\\sourcetree.exe" , "Atlassian SourceTree"]
     ] 

I tried using the shims to link to the start menu, but the user experience was awkward as it spawns a command prompt first, and then launches the app. Closing the command prompt window also kills the app. Linking to the exe directly makes everything behave as expected. I have a follow up PR as well. Again, this is an optional manifest item and shouldn't affect apps that do not want, or simply do not specify shortcuts.

PR #679, opened 2016-02-07
Startmenu Shortcut Support 

damnhandy:
PR for issue #678

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -554,16 +554,20 @@ function create_startmenu_shortcuts($manifest, $dir, $global) {
     $manifest.shortcuts | ?{ $_ -ne $null } | % {
         $target = $_.item(0)
         $name = $_.item(1)
-        startmenuShortcut "$dir\$target" $name
+        startmenu_shortcut "$dir\$target" $name
     }
 }
 
-function startmenuShortcut($target, $shortcutName) {
+function startmenu_shortcut($target, $shortcutName) {
     if(!(Test-Path $target)) {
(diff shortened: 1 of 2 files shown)


lukesampson:
This looks cool. What do you think about grouping the shortcuts under a start menu folder? Maybe called "Apps" or something generic. Just so we don't overwrite the user's other shortcuts accidentally.

Also can you check over the code style for `createShortcut` so it matches the other code? Lower-case, underscored names etc. Everything else looks good.

damnhandy:
Yeah, I was wondering about something like that myself. I was thinking something more specific to Scoop like `Scoop Apps` for `Scoop Applications`. Reason being I know a few places that group corporate apps under a Start menu folder called `Apps`. Grouping them under a single folder would make more sense.

lukesampson:
Cool, let's try it out!

Stanzilla:
@lukesampson Any reason why not put the shortcuts into the start menu directly instead of using the `Scoop apps` folder? Would make them feel more "native" if you know what I mean, taking scoop out of the shortcut making them feel like they were installed "normally".

rasa:
Or maybe set the folder name as a config option:
```
scoop config startmenu_folder “Scoop Apps”
```
and default to “” for new installs?

lukesampson:
Yes, that might be nicer installing directly to the start menu by default but making it configurable.

rasa:
This would address the Windows 10 issue in https://github.com/lukesampson/scoop/issues/2124

PR #699, opened 2016-02-18
Update go to 1.6

joe-chung:
Go 1.6 just [shipped](http://blog.golang.org/go1.6) so let's update the bucket.


lukesampson:
That was quick, thanks :)

PR #718, opened 2016-02-23
Update leiningen to 2.6.1

joe-chung:
I also changed the URL to point to 2.6.1 instead of stable, because stable changes as new versions are released.


lukesampson:
Okay cool

PR #722, opened 2016-02-28
Use https protocol

h404bi:
...to prevent connection reset. ~~(in a special country)~~


lukesampson:
Thanks. Wasn't aware that was a problem.

PR #743, opened 2016-03-13
Use standard (not ssl) mongodb version

william-ellis:
The current manifest points to the `2008plus-ssl` distribution, which doesn't work for me. I get this error when I try to execute `mongodb.exe`: *"The ordinal 341 could not be located in the dynamic link library C:\Users\William\AppData\Local\scoop\apps\mongodb\3.2.3\bin\mongod.exe"*. (The error can be solve by downloading some ssl libraries as described [here](http://stackoverflow.com/a/34020010). But then I might as well install mongodb manually).

Using the normal distribution fixes this problem.

Of course this is no good for people who need the ssl version. So maybe there should be a separate manifest for that?

### DIFF
--- a/bucket/mongodb.json
+++ b/bucket/mongodb.json
@@ -4,9 +4,9 @@
     "license": "http://www.mongodb.org/about/licensing/",
     "architecture": {
         "64bit": {
-            "url": "https://fastdl.mongodb.org/win32/mongodb-win32-x86_64-2008plus-ssl-3.2.4.zip",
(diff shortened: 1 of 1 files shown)


deevus:
@william-ellis Perhaps PR this into https://github.com/scoopinstaller/versions under the name `mongodb-nossl` 

@lukesampson

william-ellis:
The problem was solved in 32015c5

PR #746, opened 2016-03-18
Updates OpenSSL to 1.0.2g

berwyn:
1.0.2f actually leads to a 404, so OpenSSL can't be installed with the 1.0.2f manifest :/


deevus:
Thanks!

PR #748, opened 2016-03-23
Update Apache checksums/hashes

MPLew-is:
The checksums/hashes provided in the current manifest are wrong according to the Apache Lounge website ([64-bit](https://www.apachelounge.com/download/VC14/binaries/httpd-2.4.18-win64-VC14.zip.txt), [32-bit](https://www.apachelounge.com/download/VC14/binaries/httpd-2.4.18-win32-VC14.zip.txt)), and attempting to execute `scoop install apache` yields an error that the hash check failed.

There could have been a silent update or just old hashes kept when updating versions, but either way this change should fix the problem.


lukesampson:
Thank you :)

PR #754, opened 2016-03-27
Make installers/uninstallers more flexible

MPLew-is:
With some installers/uninstallers, there is no need to add arguments when running the executable, and scoop returns an error when there are no arguments present. 400284a fixes this by adding a check if there are arguments present in the manifest, only attempting to pass those arguments if they exist to avoid the error.

Additionally, the current installation method _always_ removes installers after installation; however, with programs such as the Visual C++ Redistributables, the installer is the same as the uninstaller, and needs to be kept for uninstalling in the future. Thus, ee57da6 adds a "keep" flag to the manifest, and does not remove the installer if that flag is present and set to the string "true". This change is fully backwards-compatible, and existing manifests without this flag will have their installers removed as normal.

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -360,12 +360,12 @@ function run($exe, $arg, $msg, $continue_exit_codes) {
         {
             $proc = start-process $exe -wait -ea stop -passthru -arg $arg
         }
-        
+
         else
         {
             $proc = start-process $exe -wait -ea stop -passthru
         }
-        
+
         if($proc.exitcode -ne 0) {
             if($continue_exit_codes -and ($continue_exit_codes.containskey($proc.exitcode))) {
                 warn $continue_exit_codes[$proc.exitcode]
@@ -476,7 +476,7 @@ function install_prog($fname, $dir, $installer) {
         if(!$installed) {
             abort "installation aborted. you might need to run 'scoop uninstall $app' before trying again."
         }
-        
+
         #Don't remove installer if "keep" flag is set to true
         if (!($keepInstaller.keep -eq "true"))
         {
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -361,7 +361,7 @@ function run($exe, $arg, $msg, $continue_exit_codes) {
         {
             $parameters.arg = $arg;
         }
-        
+
         $proc = start-process $exe -wait -ea stop -passthru @parameters
 
 
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -478,7 +478,7 @@ function install_prog($fname, $dir, $installer) {
         }
 
         #Don't remove installer if "keep" flag is set to true
-        if (!($keepInstaller.keep -eq "true"))
+        if (!($installer.keep -eq "true"))
         {
             rm $prog
         }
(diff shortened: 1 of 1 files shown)


lib/install.ps1
```diff
@@ -476,7 +476,12 @@ function install_prog($fname, $dir, $installer) {
         if(!$installed) {
             abort "installation aborted. you might need to run 'scoop uninstall $app' before trying again."
         }
-        rm $prog
+        
+        #Don't remove installer if "keep" flag is set to true
+        if (!($keepInstaller.keep -eq "true"))
```
lukesampson: I can't see where the `$keepinstaller` variable is set. Should this be `$installer.keep`?

MPLew-is:
You're completely correct; that was from a previous version of the changes while I was testing them. It should be fixed now; thanks for catching that!

lib/install.ps1
MPLew-is: You're completely correct; that was from a previous version of the changes while I was testing them. It should be fixed now; thanks for catching that!

lib/install.ps1
```diff
@@ -355,7 +355,17 @@ function args($config, $dir) {
 function run($exe, $arg, $msg, $continue_exit_codes) {
     if($msg) { write-host $msg -nonewline }
     try {
-        $proc = start-process $exe -wait -ea stop -passthru -arg $arg
+        #Check if arguments were provided in the manifest, and pass them to the installer if present
+        if ($arg)
```
lukesampson: I wonder if there's a way to make this more than concise in Powershell without `if/else`. If there's no way to make ps simply ignore a null argument, maybe we could prepare the arguments and use `@` splatting?

If you're not sure about the PS, do you have an example of an installer that doesn't take arguments and throws an error if you do `-arg $null`? I don't mind taking a look.

MPLew-is: This came up when I was adding the Visual C++ Redistributable manifests for #755. Luckily, I found some undocumented installation options and it's now a moot point, but I hoped this would eliminate issues in the future. As a demonstration, I've created a "loud" Visual C++ app manifest with no installer arguments at https://raw.githubusercontent.com/MPLew-is/scoop-experimental/master/visual-c-14-loud.json, or as `visual-c-14-loud` in the bucket at https://github.com/MPLew-is/scoop-experimental.

For me, this was failing every time until I added some argument (I added a dummy argument at one point just to see what would happen, and it did work), with PS saying that it couldn't bind arguments. I can get you the exact error text if you're not able to reproduce it.

MPLew-is: After resetting scoop to the current version (before this pull request, that is), this is the error that I get:

screen shot 2016-03-27 at 04 01 19

I'm definitely a relative PowerShell newbie, but the error message above seems to indicate there's no way to include any null values in the arguments without this failing. I added the check for arguments as it was the simplest way to accomplish the correct result, but if there's a more elegant approach, I'm all for it.

lukesampson: I think this will work....

```
    # allow for null $arg
    $params = @{ }
    if($arg) {
        $params.arg = $arg;
    }
```

and then instead of the if/else:
$proc = start-process $exe -wait -ea stop -passthru @params

Minor detail, but avoids repetition.

MPLew-is: Yep, that change worked perfectly, and I've added it to this pull request

lukesampson:
Thanks :+1:

PR #755, opened 2016-03-27
Add PHP and Apache Visual C++ dependency

MPLew-is:
Apache and PHP both depend on having Visual C++ 2014 installed, so this dependency is added to those manifests, along with a manifest for Visual C++ 2014 itself.

Also includes a minor change to the hashes for Apache, to explicitly label it as `sha256` for verbosity and future-proofing.

Note that the Visual C++ manifest makes use of functionality added in #754 (specifically, the `keep` flag), but the manifest will still work even if that pull request is not merged, since the added functionality will simply be ignored.


lukesampson:
As far as I know, people have been able to install PHP and Apache fine so far—maybe they have the VC redist already installed? I'm concerned that adding this as a dependency may prevent people from installing/updating PHP and Apache, because scoop will try to install the VC redist that's already installed.

So:
Can you confirm that this will work if VC redist is already installed?

MPLew-is:
I can confirm that both Apache and PHP work when the Visual C++ Redistributable is already installed. Of course, scoop also installs its own version; however, since the installer runs in quiet mode, it appears to either silently fail or silently overwrite the existing installation. Either way, both programs function as expected, and it means that the dependency will get installed when people who don't have it (like me, running on an essentially fresh Windows VM) attempt to install the apps that need it.

lukesampson:
Okay. It's a bit of a bummer that these 2 apps now pop up elevation prompts because of the installer where they didn't before. But if we get negative feedback from the PHP/Apache users we can address that.

To follow naming conventions and make it a bit clearer that this isn't [visualc](https://github.com/lukesampson/scoop-extras/blob/master/visualc.json), do you mind renaming `visual-c-14.json` to `vcredist14`?

MPLew-is:
Would you mind if it's `vc-redist-14`? I prefer some spacing so someone (or more likely, me after a long night) doesn't read it as "VCR edist" and wonder what the heck it is. I'm fine with it either way, but my vote would be for having the dashes.

lukesampson:
Yeah, we usually only use dashes if it's a variant on an app, and never for versions. So I'll have to be a pedant and insist that it's `vcredist14`, sorry!

lukesampson:
Wait, I would accept `vc-redist14`.

MPLew-is:
That works perfectly for me. I've added the changes to the pull request.

lukesampson:
Thanks :smile:

MPLew-is:
In relation to the comment about requiring administrator access, the main thing that bugs me is that the prompt tends to come up in the background. The ideal situation would be for it to come up in the foreground, but I'm not familiar enough with PowerShell to know if there's an easy way to do that.

If there's not a way of coercing `start-process` into prompting in the foreground, could a `sudo` flag be added to the manifest under the `installer` directive? The flag would simply run `sudo start-process ...` if true. It means every app that needs administrator rights to install would need `sudo` as a dependency, but, from the few quick experiments I've done, it would at least trigger the prompt in the foreground rather than have it sit in the task bar until I notice it.

lukesampson:
There's no way to control where the elevation prompt pops up, as far as I know. Even if we did `sudo start-process ...`, that would require an elevation prompt for `sudo`.

In testing these changes, the prompts did come up in the foreground for me. When I was working on sudo, I think the system would get into a state where all the prompts would start showing up without focus, but then it would kick back to foreground again (maybe after restarting powershell or the OS). So it might not be too much to worry about.

Even if it doesn't go away, I can't see much way around it besides trying as much as possible to do things that require elevation.

MPLew-is:
I may have found a slightly better solution that fits scoop's "self-contained" ideology a little better. I located the DLLs installed by the installers, and uploaded them to my `scoop-experimental` repo. From there, they can be directly downloaded and included in the app's folder as an additional `url` in the manifest, eliminating the administrator prompt altogether.

This solves the administrator-prompt issue, but means that the dependency is no longer shared. I'm assuming this is the preferred way of doing it, so should I make a pull request for this?

lukesampson:
That's great! Yes, please make a PR.

PR #758, opened 2016-03-29
Remove shortcuts from `Scoop Apps` on uninstall

neoeinstein:
Very minor change. I noticed that on some of the test install/uninstalls that I was doing from the `extras` bucket, the shortcuts weren't being removed from the correct start menu group.


lukesampson:
Nice catch, thank you.

PR #765, opened 2016-03-29
Update `git` to 2.7.4.windows.1

neoeinstein:
Yet another pull request to update versions in main.

_Side note_: in the future, would you rather me continue doing these as per-package PRs, or would you prefer fewer, larger PRs?


lukesampson:
Yes, fewer PRs would be good. Thanks.

neoeinstein:
Will do. Thanks for the quick merges in the meantime!

Issue #763, opened 2016-03-29
PHP 7.0.4 requires Visual C++ Redistributable for Visual Studio 2015


zQueal:
Using `scoop` with `scoop install php` on a newly formatted machine brings an error;

    The program can't start because VCRUNTIME140.dll is missing from your computer. Try reinstalling the program to fix this problem.

Simple research shows that PHP VC17 requires the `Visual C++ Redistributable for Visual Studio 2015` package to be installed. Perhaps `scoop` should automatically check and warn of this upon install? [Link to download](https://www.microsoft.com/en-us/download/details.aspx?id=48145).

lukesampson:
Oh right. That was supposed to be fixed in #756. I'm assuming you just installed scoop today or ran `scoop update` recently, is that right?

See also #755 for context—we tried installing this automatically, but the admin elevation prompt isn't great.

If you can work out some powershell to check if the dependency is installed, we could show a warning in the "post_install". Happy to accept a pull request if you can work it out. Or maybe @MPLew-is has thoughts?

zQueal:
> I'm assuming you just installed scoop today or ran scoop update recently, is that right?

Probably 4-5 minutes before installing PHP. Correct. 

> If you can work out some powershell to check if the dependency is installed, we could show a warning in the "post_install". Happy to accept a pull request if you can work it out.

I opened this issue because I couldn't find anything too similar and to simply track the issue--but apparently that's not the case. The installer is `MSI` so it should be relatively trivial to implement. I'll take a look into it soon.

MPLew-is:
Hmm, can you verify that there is a `vcruntime140.dll` file in the directory that contains the result of `scoop which php`? Scoop should be downloading the needed DLLs when you run `scoop install php`, and every version of PHP and Windows I've been able to test successfully found the DLLs and ran correctly, even when all Visual C++ Redistributables were uninstalled from the computer.

If that DLL is present and PHP still isn't finding it, then we might have to go back to the installer method.

MPLew-is:
I've gone ahead and made a pull request (#766) so that PHP will work. If we can't find a solution that uses the DLLs directly, or just want to restore functionality until the problem can be investigated further, the pull request will get everything back up and running, even if it does involve administrator prompts.

lukesampson:
Thanks @MPLew-is. Let us know how you go @zQueal.

zQueal:
@MPLew-is 

[...]

PR #767, opened 2016-03-29
Revert "Restore separate app for Apache/PHP Visual C++ dependency"

MPLew-is:
Reverts lukesampson/scoop#766, pursuant to the conversation in #763 that the direct DLL download does work.

Before merging this in, could you install the new version of PHP and confirm for sure that it works, even if your Visual C++ Redistributable is uninstalled? I'd like to make 100% sure this doesn't happen again.


lukesampson:
@MPLew-is Do I just run php.exe to make sure it's working?

MPLew-is:
Yep, if it comes up with a blank command window and no error messages, that means everything's running well.

lukesampson:
I uninstalled Visual C++ Redistributable, then ran `php`. All looks good to me. Hopefully it was a false alarm. Thanks for following up on this :+1:

PR #777, opened 2016-04-07
Adding conemu, ilspy & procexp to bucket

parthopdas:


lukesampson:
Thanks for this, but this main bucket is limited to open source command-line apps only.

You can you submit these to the [scoop-extras](https://github.com/lukesampson/scoop-extras) bucket—we have this version of conemu there already, and procexp is part of the sysinternals package, but we can have your standalone procexp too if you like.

parthopdas:
Hey Luke - thanks for looking into this.

I'll move these (and a few others) to the scoop extra bucket. As you can
see scoop was exactly what I was looking for and I got hooked as soon as I
discovered it :D

Did you consider enhancing scoop with the ability to consume installer from
http://portableapps.com/?

Its easy to create a manifest (with corresponding changes to silently
extract _online.paf.exe) of every app found in portableapps.
However I was thinking of directly consuming their Portable AppDirectory in
an automated fashion - right from searching.

Please let me know what you think. If you prefer I can cook up a prototype
for us to base the discussion on.

Thanks,
Partho

lukesampson:
Glad you're liking scoop. I'm not so interested in search the portable app directory directly from scoop as I think it introduces too much of a dependency, sorry. But if you want to make a scoop bucket and generate manifests based on their database that would be great.

parthopdas:
Separate bucket sounds great. It'll need some some support from core scoop
though - as they have a different but standard packaging mechanism.

Is there, if not already, any support for tracking download/install metrics
for packages?

Also - is there, of not already, a richer metadata based search? e.g. one
of the packages I was thinking was got credentials win store which is known
commonly as both gcws & its full hyphenated name. If I create a manifest
with either name, would searching with other name be successful?

You do have a fantastic idea with scoop. Being a homebrew user + a purist,
I would love to help you make the project a bigger success.

Partho

lukesampson:
Sorry, maybe I didn't explain well. I don't really want to extend scoop core functionality—but if you want to write some code that converts portable app into scoop manifests that would be good to see. One of the objectives of this project is to keep things simple, and I don't really have time to merge or maintain much change in the codebase at this point, sorry. I was thinking you might like to generate scoop manifests programmatically and keep them in a git repo, and just use the functionality that is already there.

`scoop search` matches the manifest name and the executables under `bin` in the manifest.

PR #781, opened 2016-04-08
Bringing back download progress

parthopdas:


lukesampson:
Thanks. I hate that powershell download progress, but it is better than nothing.

parthopdas:
Thanks for accepting the last 2 PRs.

I thought on the same lines considering it's really a marquee, not a
progress bar.

w.r.t. your previous email I haven't responded yet since the portable apps
installer doesn't have a silent mode (they say it does).

Along those lines - one of the themes of scoop is *nix command lines with
windows. How do you think scoop will be affected by bash-on-windows?

I guess the uber question I had was - would you be able to help me
understand your vision for scoop's future?

Partho

deevus:
bash-on-windows at this stage cannot interact with Windows applications. I
love that I can install apps through scoop and they will happily interact
with each other regardless.

lukesampson:
I haven't used Bash on Windows, but it seems like maybe a nicer alternative to Cygwin or Git Bash. Like @deevus said, if it doesn't run Windows apps, it's not so useful to me—but good on Microsoft for improving cross-platform support.

The main theme with Scoop was improving the command-line experience for Windows developers. I think it does a pretty good job.... maybe? For me at least. I don't have much time to work on it these days, beyond accepting PRs for manifest updates.

ntwb:
> I think it does a pretty good job.... maybe?

Yes, it does a great job IMHO 👍 

> I don't have much time to work on it these days

Nor I

PR #824, opened 2016-05-04
Update glide to 0.10.2

monotykamary:
SHA-256 taken from 7zip.


lukesampson:
Thanks for all these updates!

PR #825, opened 2016-05-05
Update openssl link

schliesser:
version changed to 1.0.2h


lukesampson:
Need to update the hashes on these.

lukesampson:
And version.

Issue #864, opened 2016-05-24
Unable to search on a new win10


wangzq:
```
PS C:\Users\wangzq> scoop search python27
invoke-webrequest : The response content cannot be parsed because the Internet Explorer engine is not available, or Internet Explorer's first-launch configuration is not complete. Specify the UseBasicParsing parameter and try again.
At C:\Users\wangzq\AppData\Local\scoop\apps\scoop\current\libexec\scoop-search.ps1:52 char:15
+     $result = invoke-webrequest $url | select -exp content | convertf ...
+               ~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotImplemented: (:) [Invoke-WebRequest], NotSupportedException
    + FullyQualifiedErrorId : WebCmdletIEDomNotSupportedException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand

```

Will send a PR shortly.

PR #865, opened 2016-05-24
#864 Fix following issue when searching on a new Win10

wangzq:
invoke-webrequest : The response content cannot be parsed because the Internet Explorer engine is not available, or Internet Explorer's first-launch configuration is not complete. Specify the UseBasicParsing parameter and try again.


lukesampson:
Thanks for this.

PR #867, opened 2016-05-28
Added NVM to the bucket

### LINKED ISSUE #783: Support for NVM Windows
I wasn't sure on what the process was for requesting package support but. 

Has there been any thought into adding nvm for windows.
https://github.com/coreybutler/nvm-windows

They have a noinstall version. I maybe able to try and add it if it's something that this project would like.

jahead:
This will install a working copy of nvm, it also installs all the nodes in the dir where scoop install nvm, which is nice because unless installed Globally should allow users to use different nodes versions. 

Sorry for taking so long. Fixes #783


lukesampson:
Sounds good. Thanks.

PR #890, opened 2016-06-03
Make friendly_path return input when $home is not set

### LINKED ISSUE #889: Running `scoop which` under LocalSystem account
When running `scoop which <installed package>` under LocalSystem account in Windows we get back a string like this:
```
C:~\programdata~\scoop~\apps~\
```

Running the same command under a regular account with Administrator privileges returns the following string:
```
C:\programdata\scoop\apps\
```

As you can see the `~` is present several times in the string causing it to not resolve properly when we are using it inside a script.

Any suggestions to why this is and how we can work around it?

vidarkongsli:
Trying to fix #889 
cc @lukesampson


lukesampson:
I already committed the change before I saw this, but thanks anyway.

PR #909, opened 2016-06-22
Adding ReSharper Command Line Tools 2016.1.2

gjmveloso:
Hi @lukesampson,

I just discovered your amazing package manager for Windows and I've immediately moved from Chocolatey. It's really well designed and maintaining apps are cheap, expressive and clear.

I would like to add ReSharper Command Line Tools which is free to use to the main bucket. I think many of .NET developers that relies on ```scoop``` will benefit by having it.

By the way, thanks for [pshazz](https://github.com/lukesampson/pshazz) :+1:


lukesampson:
Thanks for the new app. Glad you're liking Scoop!

PR #919, opened 2016-07-01
python 3.5.2

martinlindhe:
changed checksums from md5 to sha256.

also only tested the 64bit version, but checksums was calculated from the correct binary so it "should work"


lukesampson:
Yep, that works. Thanks 👍

PR #920, opened 2016-07-05
wget 1.18

martinlindhe:
only tested 64-bit version but calculated 32bit hash from https://eternallybored.org/misc/wget/releases/wget-1.18-win32.zip, so it "should" work too


martinlindhe:
related: https://blogs.securiteam.com/index.php/archives/2701

PR #921, opened 2016-07-05
mariadb: 10.1.14

martinlindhe:
only verified install of 64-bit build, but 32-bit binary checksum was calculated from https://downloads.mariadb.org/f/mariadb-10.1.14/win32-packages/mariadb-10.1.14-win32.zip so that too "should" work. verified install worked by manually starting mysqld.exe

### DIFF
--- a/bucket/mariadb.json
+++ b/bucket/mariadb.json
@@ -1,17 +1,17 @@
 {
     "homepage": "http://mariadb.org",
-    "version": "10.0.25",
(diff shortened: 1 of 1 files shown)


lukesampson:
Thanks for these. Did you know you can check 32 bit install with `scoop install [app] -a 32bit`?

martinlindhe:
Ah, no I didn't. Will do that from now on, thanks for the hint

PR #928, opened 2016-07-05
octave 4.0.3

martinlindhe:


martinlindhe:
Note: there dont seem to be a "portable" download anymore for octave 4. Tested this formula by launching octave-gui.exe after install succeeded.

PR #933, opened 2016-07-05
less: sha256 hashes

martinlindhe:
Hello. Are these wanted? (md5 -> sha256 hashes)

If so, I can submit patches for the packages using md5 / sha1.

### DIFF
--- a/bucket/less.json
+++ b/bucket/less.json
@@ -7,8 +7,8 @@
         "http://ufpr.dl.sourceforge.net/project/gnuwin32/less/394/less-394-dep.zip"
     ],
     "hash": [
-        "md5:20f94bcc8a3347bd9a9ca32587d82eae",
-        "md5:539dce63be2b89a22af7b526d1022f72"
+        "cd8233cb3caee8ccbbed6883443f44f2002218eceb4d697d38a0c40c2d181896",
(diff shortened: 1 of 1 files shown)


lukesampson:
Usually if there are MD5 hashes used, it's because that's what the software developers publish. E.g. http://gnuwin32.sourceforge.net/packages/less.htm publishes Md5Sum next to the downloads so they're authoritative, as well as easier to put into the manifest.

martinlindhe:
Okay, thanks for explaining.

The issue I would like to address by removing the md5 sums is hash collision. It is especially bad here because many of the buckets that come with md5 is served over http, thus allowing a man-in-the-middle attack on a untrusted network (the internet).
Md5 collision is quite cheap these days, and many of these older tools have had known md5 sums for years and years, so it would be safe to assume someone already produced collisions for them.

If you are unfamiliar with md5 collisions, see this page for some examples of win32 exes with different payload and same md5 hashes: http://www.mathstat.dal.ca/~selinger/md5collision/

Sha1 is also known to have collisions, but it requires much more processing power to produce collisions for these. Here is an example: https://eprint.iacr.org/2015/967

Sha256 is still a safe and good option to stick with.

lukesampson:
You make a good case for sha256. I don't think we need to force everyone to use it, but if you're happy to submit patches then I'm good with that. Thanks.

martinlindhe:
Okay, great! :smile:

PR #962, opened 2016-07-20
LaTeX: The whole zoo of executables is now added to PATH

moigagoo:
Previously, only pdflatex and latex were added to PATH. However, xelatex and mpm and all the other stuff is useful from time to time. I'm not 100% sure this is the right decision, but it definitely is for me as an extensive LaTeX user.

### DIFF
--- a/bucket/latex.json
+++ b/bucket/latex.json
@@ -4,10 +4,7 @@
     "license": "http://miktex.org/copying",
     "url": "http://mirrors.ctan.org/systems/win32/miktex/setup/miktex-portable-2.9.5987.exe#/dl.7z",
     "hash": "b6773ba988007a88c361f63d27833e63fefdb8011238f0ca843135a3c4ffded9",
-    "bin": [
-        "texmfs\\install\\miktex\\bin\\latex.exe",
-        "texmfs\\install\\miktex\\bin\\pdflatex.exe"
-    ],
+    "env_add_path": "texmfs\\install\\miktex\\bin",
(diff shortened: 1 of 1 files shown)


lukesampson:
Probably would be preferable to shim each .exe because the path environment variable is limited to 260 characters, which gets exhausted quickly if you install lots of apps.

moigagoo:
@lukesampson Thanks for the info, I was unaware of this restriction. Manually shimming each LaTeX executable is mundane since there're so many of them. It would be great if the `bin` param supported glob syntax, so instead of `bin: ["path/to/exec1", "path/to/exec2", ...]` we could do `bin: "path/to/all/execs/*". Do you think it's a good idea?

nightroman:
PATH is really limited and this is worth to keep in mind. It is not that short though - https://software.intel.com/en-us/articles/limitation-to-the-length-of-the-system-path-variable

lukesampson:
@moigagoo I just saw how many exes are in the latex bin directory—way more than I expected. I don't blame you for not shimming them all.

@nightroman oh you're right, that is much longer than 260 characters. I confused it with the maximum path length.

So it's probably not that big a deal then to keep the latest change, adding the whole directory to the path rather than shimming each .exe or the wildcard.

ForNeVeR:
There's also a [Microsoft patch](https://support.microsoft.com/en-us/kb/2685893) to reduce some of the symptoms on various versions of Windows 7 and 2008.

PR #995, opened 2016-08-17
Updates git, node and thrift

gjmveloso:
Hi @lukesampson,

I'm updating some packages to their new versions
I'm also fixing an additional http on thrift homepage that was preventing ```checkver.ps1``` to run as expected.

Thanks,
Gustavo


lukesampson:
Thanks 👍

PR #1016, opened 2016-08-30
Update checkvers

gjmveloso:
Hi @lukesampson,

I've fixed ```checkvers``` that were returning wrong information. Still don't be able to fix ```sliksvn.json```

Thanks,
Gustavo


lukesampson:
Thanks!

PR #1021, opened 2016-09-06
Change make source

gjmveloso:
Hi @lukesampson,

I've found a more updated ```make``` source for Windows at http://www.equation.com/servlet/equation.cmd?fa=make

I think it's a good idea to change the scoop location to it.

Thoughts?

Thanks,
Gustavo


lukesampson:
Sounds good to me. Thanks!

PR #1022, opened 2016-09-07
update/curl

gjmveloso:
Hi @lukesampson,

I've updated the ```curl``` package and the corresponding strategy to find the versions quickly.

Thanks,
Gustavo


lukesampson:
Thank you!

PR #1031, opened 2016-09-14
Update pngcrush manifest to version 1.8.6

mdiehr:
Version 1.8.4 was moved to an old_versions directory so this manifest wasn't functioning any more.

Updated urls to those from here: https://sourceforge.net/projects/pmt/files/pngcrush-executables/1.8.6/
Used sha1 hash as listed on sourceforge.


lukesampson:
Thanks

PR #1053, opened 2016-10-03
Update Nim to 0.15.0

moigagoo:


moigagoo:
@lukesampson I'm terribly sorry but it turns out they have changed the installer in 0.15.0 so it no longer works the same way as before. It installs all right, but it's practically unusable.

Could you please undo this PR and roll back to the previous version?

lukesampson:
No problem. Thanks for letting me know.

PR #1060, opened 2016-10-03
Revert "Update Nim to 0.15.0"

lukesampson:
Reverts lukesampson/scoop#1053


moigagoo:
@lukesampson Fixed the issues with Nim 0.15.0 in https://github.com/lukesampson/scoop/pull/1061

PR #1073, opened 2016-10-10
Update git-with-openssh to 2.10.1.windows.1

inosik:
Again, simply copied and pasted the changes from #1067.

Should we add some kind of a comment to both manifests, so that maintainers are aware that there are two of them?

### DIFF
--- a/bucket/git-with-openssh.json
+++ b/bucket/git-with-openssh.json
@@ -4,12 +4,12 @@
     "version": "2.10.0.windows.1",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


lukesampson:
Yeah we could add a comment. But I'm thinking we should maybe make this version THE `git.json`, since it seems OpenSSH is broken everywhere else, and this version seems to be the one most people want/need.

inosik:
I'm actually using `git` + `openssh` and haven't had any problems yet.

What if somebody installs another app that depends upon `openssh`?

lukesampson:
Oh okay, let's leave it as it is then. Are you on Windows 10?

inosik:
No, I'm on Windows 7.

lukesampson:
Okay thanks. I've added comments to both manifests.

PR #1075, opened 2016-10-10
Add Caddy Bucket

klaidliadon:
@lukesampson Is there a way to rename the bin after extraction?
I'd like to rename `caddy*.exe` to `caddy.exe`.


lukesampson:
Let me know if you want to merge as it is.

klaidliadon:
Now it looks ok! Can merge!

PR #1083, opened 2016-10-12
Add Gitter badge

tekacs:
This repo seems prime to have a bunch of conversation going in the Gitter channel - all sorts of other projects use it to great effect.

Whilst @lukesampson isn't on Gitter, the channel already has a few lurkers!

Might be worth adding a Gitter chat badge?


lukesampson:
Thanks.

PR #1109, opened 2016-11-07
Print '0 B' when cache is empty

asmgf:
If cache is empty `$total_length` is null, and the output is:
```
λ scoop cache show some-non-existing-app
total: 0 files,  B
```

Cast sum to double to get '0 B' instead


lukesampson:
Thanks 👍

Issue #1108, opened 2016-11-05
apache hash check failed


fghhfg:
```
λ scoop install apache
installing apache (2.4.23)
downloading https://www.apachelounge.com/download/VC14/binaries/httpd-2.4.23-win64-VC14.zip...(14.4 MB) done
checking hash...hash check failed for https://www.apachelounge.com/download/VC14/binaries/httpd-2.4.23-win64-VC14.zip. expected: 32A06E5E15CA3E344C2C548E69B8D0ADFE40F6AE955A7028E87B87FE677B3B70, actual: af90d95da25296f945953feb45ad755cf05acf064f3f51122ec4bc3b36905749!
```

rrelmy:
this can be closed

PR #1112, opened 2016-11-09
update apache hashs (#1108)

rrelmy:
It's the same version but compiled with newer OpenSSL versions

https://www.apachelounge.com/viewtopic.php?p=33747


lukesampson:
Thanks for all these PRs!

PR #1116, opened 2016-11-09
Update outdated apps

rrelmy:
I found the `bin/checkver` tool and was bored :-)

All apps expect `nano` (can't find a binary) and `rancher-compose` (not a stable release) are now up-to-date.


lukesampson:
Thank you!

PR #1119, opened 2016-11-10
Add checkver definition to many apps

rrelmy:
Many manifests were missing a `checkver` definition for easy version checking.

I added them to many of the apps.
Additionally I added a small script to `bin` listing all the manifests without an `checkver` definition


lukesampson:
Great! Thanks.

PR #1120, opened 2016-11-10
Command 'scoop update' delete scoop if download fail. 

### LINKED ISSUE #1118: Command 'scoop update' delete scoop if download fail.
When you run 'scoop update' it first uninstall scoop app, them it start a whole new install of scoop. If there is a problem in the download, like a proxy error, you get scoop scripts removed.

What do you think about download first, check hash, and them uninstall app?

felipecrp:
I´m thinking in something like this. But i don´t know how to test. Can you provide some instructions?

Also, what does '-ea stop' means?

fixes #1118

### DIFF
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -80,7 +80,12 @@ function update_scoop() {
     @(buckets) | % {
         "updating $_ bucket..."
         pushd (bucketdir $_)
-        git_pull -q
+
+        try {
+            git_pull -q -ea stop
+        } catch {
+            abort 'scoop bucket update failed'
+        } 
         popd
     }
     success 'scoop was updated successfully!'
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -58,7 +58,7 @@ function update_scoop() {
 
         try {
             # get git scoop
-            git_clone -q $repo --branch $branch --single-branch "`"$newdir`"" -ea stop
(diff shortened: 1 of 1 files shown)


lukesampson:
To test a failed update, it should work to create a `test.json` manifest in your working directory, install it with `scoop install .\test.json` , then update the version as well as adding a bad URL, hash etc.

`-ea stop` is short hand for the `-ErrorAction stop` parameter common to many (or all?) Powershell commandlets (or whatever they're called). If I remember correctly, `stop` causes the command to throw the exception allowing you to catch it, rather than print an error and continue. I'm not sure if it's necessary with `git_clone` because that's just a normal function rather than a commandlet.

This looks good as far as I can tell. Can you let me know when you've tested and you're happy and I'll try it out.

P.S. Sorry about the slow response—I've been moving house.

felipecrp:
Are there any way to switch my "scoop environment" for my local branch? I would like to test it isolated from my main instalation.

felipecrp:
Now it is OK! If you get proxy error on 'scoop update' it doen´t break your installation.

To test this patch, just run:

    scoop config proxy localhost:1  
    scoop update

Also, make sure the the current scoop instalation doesn´t contains a .git repo. If it has, you must delete this first. 

Before this patch, scoop would be missing. After, nothing will happen, so you can fix your proxy and try the update again, when everything gonna be ok.

lukesampson:
Great! I've tested and it works well to handle failure the first time `scoop update` is run. Thanks for this!

felipecrp:
👍

PR #1132, opened 2016-11-17
Bumped leiningen version

nrakochy:
Updated leiningen to 2.7.1


lukesampson:
Is there a version specific URL you can use rather than the "stable" one? It's better to use a version specific URL if there is one, otherwise installs will break (the hash will fail) whenever they release a new version

nrakochy:
Sounds good. Thanks again for the great tool.

PR #1138, opened 2016-11-26
First implementation of manifest autoupdate

rrelmy:
Hi

Some time ago I added `checkver` informations to a lot of manifests, now we know that a lot of manifests are outdated. It will take a lot of work to update all those manifests manually.

I started implementing an easy way to update manifests. The idea is that the `checkver` script not only finds the newest version string but also the download urls and hashes for the files.

Most time the download url can be build by only knowing the version string, alternatively it can be extracted from the website. The hashes can be obtained from the download page (php etc.) or from a separate file (nodejs etc.) the last resort would be downloading the file and hash it locally.

I implemented a the base for it, the nodejs manifest contains the new `autoupdate` block. Running `bin/checkver.ps1 nodejs -u` updates the manifest file with the newest version. Of course the current nodejs manifest is up to date.

Writing the JSON file from powershell results in a _not that pretty_ file as you can see in the attached commit, this is one of a few things that can be improved

I would like to have some feedback about the idea
 - What do you think about it?
 - Would a complete implementation be appreciated?
 - I am new to powershell scripting, any comments/tips to my coding style?
 - `checkhash` could be implemented to spot broken manifests

Todo (some todos are documented in the sourcecode)
 - Implement more ways of getting the URL and Hash (Extract from page, Github Release, etc,)
 - Better JSON output (any ideas how to do that)
 - Testing
 - Writing documentation

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -1,10 +1,22 @@
 # checks websites for newer versions using an (optional) regular expression defined in the manifest
 # use $dir to specify a manifest directory to check from, otherwise ./bucket is used
-param($app, $dir)
+param(
+    [String]$app,
+    [String]$dir,
+    [Switch]$update = $false
+)
+
+if (!$app -and $update) {
+    # While developing the feature we only allow specific updates
+    Write-Host "[ERROR] AUTOUPDATE CAN ONLY BE USED WITH A APP SPECIFIED" -f DarkRed
+    exit
+}
 
 . "$psscriptroot\..\lib\core.ps1"
 . "$psscriptroot\..\lib\manifest.ps1"
 . "$psscriptroot\..\lib\config.ps1"
+. "$psscriptroot\..\lib\buckets.ps1"
+. "$psscriptroot\..\lib\autoupdate.ps1"
 
 if(!$dir) { $dir = "$psscriptroot\..\bucket" }
 $dir = resolve-path $dir
@@ -79,6 +91,10 @@ while($in_progress -gt 0) {
             } else {
                 write-host "$ver" -f darkred -nonewline
                 write-host " (scoop version is $expected_ver)"
+
+                if($update -and $json.autoupdate) {
+                    autoupdate $app $json $ver
+                }
             }
 
         } else {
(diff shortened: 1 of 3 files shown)


lukesampson:
This is great! Nice work. Should keep programs much more up to date and save lots of tedious work at the same time. It'd be great if you can keep going with this.

How hard would it be to split out the checkver and autoupdate parts into another `autoupdate` directory, i.e. `autoupdate\node.json`? I think it's worthwhile keeping app manifests simple so it's easy to see what install steps will be taken, and to make it less intimidating to get started writing manifests. And a separate directory will allow us to see at a glance which apps have autoinstall and which don't. Would that work?

Your Powershell coding style is nice. No complaints there!

JSON output – I think `scoop create` suffers from similar problems. The automated tests show an error about the UTF8-BOM, so might need to do [this sort of thing](http://stackoverflow.com/questions/5596982/using-powershell-to-write-a-file-in-utf-8-without-the-bom/5596984#5596984). I know it's lame, but maybe you could do a regexp replace to clean up the indentation and any other formatting problems.

Thanks for your work, and I'm looking forward to seeing what you come up with next!

rrelmy:
Thanks, I will continue implementing it 

I am not sure if splitting the manifest into multiple files will be a good idea. I would like to encourage people to also add the optional `checkver` and `autoupdate` blocks if possible. But I understand your point.
It would be easy to do it.

The `ConvertTo-Json` will be improved soon [PowerShell#2736](https://github.com/PowerShell/PowerShell/issues/2736) I will add a function to reformat the json output.

Is there a better place to discuss this?

lukesampson:
Great! If you think it works better to keep autoupdate in the manifest then that's fine by me. I think package maintainers will jump on this functionality either way if it's available.

You can email me directly at me@lukesampson.com if you like, otherwise here on GitHub is probably the best place for discussion.

rrelmy:
Better JSON Output will land with PowerShell 6 https://github.com/PowerShell/PowerShell/releases/tag/v6.0.0-alpha.17

PR #1155, opened 2016-12-14
service management

deepakSP:
we have two alias - srvman and service

some useful command for development

sudo service start apache

sudo service stop nginx

sudo service restart php-cgi

sudo service add <file.exe/file.sys> [service name] [display name] [/type:<service type>] [/start:<start mode>] [/interactive:no] [/overwrite:yes]

sudo service delete mysql

if you want to use add, edit or delete command please refer official documents. - http://tools.sysprogs.org/srvman/

### DIFF
--- a/bucket/service.json
+++ b/bucket/service.json
@@ -1,0 +1,15 @@
+{
+  "homepage": "http://tools.sysprogs.org/srvman/",
+  "version": "1.0",
+  "architecture": {
+    "64bit": {
+      "url": "https://raw.githubusercontent.com/deepakSP/srvman/master/x64/srvman.exe"
+    },
+    "32bit": {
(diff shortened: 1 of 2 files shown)


lukesampson:
Cool! This is a great start. A couple of small things:

* Can you rename the file to srvman.json? Scoop packages go by the name of the program rather than the executables. It looks like it would still be discoverable with `scoop search service` since you're adding that alias.
* The app veyor build is failing because we have a test to ensure all files end with newlines. If you can fix that up it helps me merge quicker.

deepakSP:
sure, i am checking veyor

deepakSP:
done 👍

lukesampson:
Great! By the way, you can also run the tests locally with the `bin/test.ps1` command in the scoop source directory.

PR #1160, opened 2016-12-16
nssm hash update 

deepakSP:
nssm hash update

### DIFF
--- a/bucket/nssm.json
+++ b/bucket/nssm.json
@@ -2,7 +2,7 @@
   "homepage": "https://nssm.cc",
   "version": "2.24",
   "url": "https://nssm.cc/release/nssm-2.24.zip",
-  "hash": "7f0075c5b544af189416845077a45edeedd1d48e37e702abb22b41b120e80bcd",
+  "hash": "727d1e42275c605e0f04aba98095c38a8e1e46def453cdffce42869428aa6743!",
   "architecture": {
     "64bit": {
       "extract_dir": "win64"
--- a/bucket/nssm.json
+++ b/bucket/nssm.json
@@ -2,17 +2,17 @@
   "homepage": "https://nssm.cc",
   "version": "2.24",
   "url": "https://nssm.cc/release/nssm-2.24.zip",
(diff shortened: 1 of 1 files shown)


lukesampson:
The ! at the end of the hash won't work. You can test these before committing by running `scoop install /path/to/manifest.json`

deepakSP:
@luiscoms how can i add valid hash ?

lukesampson:
One way to get a hash is to create your manifest without hashes, install it with scoop, and copy the hashes from the warnings that scoop gives you. Hashes are numbers and letters—don't include the ! at the end of the message.

You can also `scoop install shasum` and the run `shasum path\to\download\file.zip` to get a SHA256 hash manually.

Busybox and gow (available as scoop packages) also contain a sha1sum tool.

Whichever way you choose, you should always test the install locally using `scoop install path\to\your\manifest.json` to make sure it works before submitting a pull request.

deepakSP:
actually i tried this method but when i tried installing i got error hash miss match, so i removed it. Will try with shasum.

Thank you.

what you think about adding phar to windows executable 
example - 
if user wants to install php-cs-fixer.phar they can run
scoop install php-cs-fixer
and internally scoop will save phar file in shims directory and run
echo @php "%~dp0php-cs-fixer.phar" %* > php-cs-fixer.cmd
to create executable file.

this will save lots of time to search - how to run phar file and use inside app like phpstrom

lukesampson:
I'm not sure about this Phar thing. It might be best to create a private [bucket](https://github.com/lukesampson/scoop/wiki/Buckets) for yourself if you want to do something like this. If it's a well-known or well-defined program that would be different, but it sounds more like a helper script.

luiscoms:
`Phar` program is a `php` executable file

lukesampson:
I think I roughly understand how it works.

It sounds like you're mainly interested in using the shimming capabilities of Scoop to create an .exe that calls php and passes it the .phar file, is that right?

I'd be happy to accept a pull request for something like this to [scoop-extras](https://github.com/lukesampson/scoop-extras), so the process for installing it would be something like:

```
scoop bucket add extras
scoop install php-cs-fixer
```

PR #1165, opened 2016-12-20
Update apache 2.4.23-->2.4.25.

vidarkongsli:
Installation was broken because the old distro was removed from the web page.


lukesampson:
Thanks

PR #1168, opened 2016-12-20
Add emacs

asmundg:


lukesampson:
Actually there's already a [manifest for Emacs](https://github.com/lukesampson/scoop-extras/blob/master/emacs.json) in scoop-extras. You need to run

```
scoop bucket add extras
scoop install emacs
```

asmundg:
Aha, awesome. Thanks!

PR #1174, opened 2016-12-25
set execution policy on scoop install

mikhail-tsennykh:
I've tested this by running updated script as below:
```iex (new-object net.webclient).downloadstring('https://gist.githubusercontent.com/mikhail-tsennykh/f65a3addf22b36681b1eb7c6d708bed2/raw/4040fe0c156512a63b1dcfea2fd5595b26e1a1f1/scoop%2520auto%2520set%2520execution%2520policy%2520test')```

nice to skip extra step (unless there is some good reason not to do it on scoop install).

### DIFF
--- a/bin/install.ps1
+++ b/bin/install.ps1
@@ -4,6 +4,15 @@
 #   iex (new-object net.webclient).downloadstring('https://get.scoop.sh')
 $erroractionpreference='stop' # quit if anything goes wrong
 
+# set correct execution policy automatically if not set already:
+$ep = get-executionpolicy cu; 
+if ($ep -ne "unrestricted") {
+    # Set execution policy on the process to hide error message
+    # that will otherwise show first time this script is run
+    set-executionpolicy unrestricted p
+    set-executionPolicy unrestricted cu 
+}
+
 # get core functions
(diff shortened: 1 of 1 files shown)


lukesampson:
Good idea, but I think it's a bit impolite to change people's powershell execution policy for them without notice. If this could show a warning about what needs to be done, tell people what command to run and then exit, that would be good.

mikhail-tsennykh:
Good point, pushed a change to handle this, will now confirm with user if they want to change current user execution policy to unrestricted, also stop the process if not confirmed, showing relevant notification message.

mikhail-tsennykh:
Just realized it shouldn't actually change the policy, just show the message it needs to be done with powershell command and stop the process.

lukesampson:
Thanks!

lukesampson:
Does the `break` on line 12 stop execution of the script? Should it be `exit 1` instead to return a non-zero exit code?

mikhail-tsennykh:
yep, `break` just stops script execution but keeps powershell window open, while `exit 1` closes the window, which might be inconvenient if user needs to set execution policy then run install again

lukesampson:
Ah okay, cool.

PR #1177, opened 2016-12-26
upgrade dotnet core to 1.1

yunspace:
The code release version is actually `1.0.0-preview2-1-003177`, however it is packaged as `1.1` in the official download page https://www.microsoft.com/net/download/core#/sdk/current. So I opt to `1.1` that instead of the never ending and non-distinguishable `1.0.0-preview2-*` versions

### DIFF
--- a/bucket/dotnet.json
+++ b/bucket/dotnet.json
@@ -1,14 +1,14 @@
 {
-    "version": "1.0.0-preview2",
+    "version": "1.1",
     "homepage": "https://www.microsoft.com/net/core#windows",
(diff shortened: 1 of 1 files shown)


lukesampson:
Okay I'll accept this. I do think it would be safer to use the full version—that way if there's a `1.0.0-preview3-*` or `1.0.0-rc1-*` packaged as `1.1` then scoop will be able to let people know there's an update available.

yunspace:
Hmm I think that is a good point. I'll resubmit

PR #1181, opened 2016-12-30
Add Windows NT to useragent (fixes #1180)

### LINKED ISSUE #1180: Sourceforge needs "Windows NT" in user agent to display proper version
Hitting https://sourceforge.net/projects/smartmontools/files/smartmontools/ with a user agent of "[Scoop/1.0 (+http://scoop.sh/)](https://github.com/lukesampson/scoop/blob/master/bin/checkver.ps1#L46) " causes SourceForge to output `Download smartmontools-6.5.tar.gz`. Setting the user agent to "Scoop/1.0 (+http\://scoop.sh/) (Windows NT 6.1; WOW64)" causes SourceForge to output `Download smartmontools-6.5-1.win32-setup.exe`, which is the current version.

rasa:


lukesampson:
Thanks for finding and fixing this problem!

PR #1186, opened 2017-01-01
Autoupdate improvments / App update / SHA512 support

rrelmy:
_Sorry for the big pull request, I wanted to make multiple ones, but I ended up having dependencies between them_

## Autoupdate
 - [Breaking] I changed the `url` reference inside `autoupdate` to match the core logic ([commit](https://github.com/rrelmy/scoop/commit/68781a890ed84e3ee7646ded03ed1df98a29de7d))
 - Autoupdate can now extract hashes from RDF XML files. (The imagemagick already use it)
 - Option to update `extract_dir` with new version
 - Hash mode `download` is now the default if nothing is specified

_I also started to document the autoupdate pieces in the wiki_

@rasa 
I noticed you created your own [scoops](https://github.com/rasa/scoops) repo, nice work there.
You need to update your manifests if you have separate 32bit and 64bit URLs. (_Take a look at the breaking change_)
And you can remove `"hash": {"mode": "download"},` because it's the default behaviour now.

## Checkver
 - Added a [github helper](https://github.com/rrelmy/scoop/commit/98fba649fb6a87761c879006ae78ce482416c152)
 - New option `-forceUpdate` to force an autoupdate

## General
While updating `ant` I only found official sha1 and sha512 hashes, instead of using sha1 I added support for sha512 to scoop ([change](https://github.com/rrelmy/scoop/commit/5170cb9185dbdcada3855de41cbb5a0a9d03bd87#diff-c2a7e0e3e8869b7de54b5f172fc77009)).
@lukesampson Trying to install the updated `ant` manifest with an old version of scoop will fail, but this should never happen as `scoop update` will update scoop and the manifests. Am I right?

## Updated apps
I also added autoupdate if not already there

 - imagemagick
 - yarn
 - nodejs
 - ant
 - curl
 - caddy
 - pt
 - mariadb
 - rancher-compose

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -3,7 +3,8 @@
 param(
     [String]$app,
     [String]$dir,
-    [Switch]$update = $false
+    [Switch]$update = $false,
+    [Switch]$forceUpdate = $false
 )
 
 if (!$app -and $update) {
@@ -90,6 +91,11 @@ while($in_progress -gt 0) {
             $ver = $matches[1]
             if($ver -eq $expected_ver) {
                 write-host "$ver" -f darkgreen
+
+                if ($forceUpdate -and $json.autoupdate) {
+                    Write-Host "Forcing autoupdate!" -f DarkMagenta
+                    autoupdate $app $json $ver
+                }
             } else {
                 write-host "$ver" -f darkred -nonewline
                 write-host " (scoop version is $expected_ver)" -NoNewline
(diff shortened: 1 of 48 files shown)


lukesampson:
Yes, `scoop update` will update Scoop first, so the `ant` update shouldn't be a problem.

rasa:
@rrelmy: Great work. Thanks for the heads up. I've updated my repo with your suggested changes, and checkver works fine.

PR #1187, opened 2017-01-02
Use manager credential helper for git

asmundg:
Manager has apparently replaced the wincred helper as of 2.x.


lukesampson:
Thanks!

PR #1192, opened 2017-01-04
Bump imagemagick and make

rasa:


bucket/imagemagick.json
```diff
@@ -19,7 +22,7 @@
         "conjure.exe",
         "convert.exe",
         "dcraw.exe",
-        "ffmpeg.exe",
+        ["ffmpeg.exe", "iffmpeg"],
```
rasa: The ffmpeg executive shipped with Imagemagick is rather dated. Aliasing the shim allows us to install the [ffmpeg](https://github.com/lukesampson/scoop/blob/master/bucket/ffmpeg.json) package alongside Imagemagick. Is there a better way?

lukesampson: @rrelmy how does that sound to you? I will merge this since the current version is 404'ing, but please let us know or submit a pull request if there's a better way.

rasa: I'm happy to undo it, if it's an issue. We could solve this in one of three ways:

1. Add a "priority" flag to each app (or each exe), and assign a higher priority to ffmpeg. install/uninstall then needs to search *all* installed apps to see if *any* app supplies an exe. Ugh.
2. When a new app installs/updates over a shim,  ask the user which one s/he wants. Breaks the ffmpeg command, if the user selects the imagick version, but later uninstalls it. Yuk.
3. Add ffmpeg's path to PATH before the shims path. Yuk, but doesn't break on uninstall.

rrelmy: That's fine for me, I just was curious.

A fourth option would be removing this alias completely

rasa: Another option would be to remove ffmpeg from imagick's bin list and add a `"depends": ["ffmpeg"]`? That probably the best solution. Agreed?

lukesampson: Yes, agreed 👍

rrelmy: I did that change in #1197 ;-)

PR #1194, opened 2017-01-05
Add support for lzh format. (with 7zip)

zeero:
Thank you for really great software!
I like scoop so much.

Please add support for extracting .lzh format.
Because in Japan, some softwares is provided this legacy format...
ex)  http://ftp.vector.co.jp/43/51/2176/SoftTilt111.lzh

7zip can extract .lzh format.


lukesampson:
Thanks for this 👍

PR #1204, opened 2017-01-10
Add Android SDK Platform Tools adb and fastboot

r15ch13:
Google now provides a small package only containing `adb` and `fastboot`.
https://developer.android.com/studio/releases/platform-tools.html

Could be handy 👍


r15ch13:
That was fast! ❤️

lukesampson:
Haha I checked my email at just the right time!

PR #1257, opened 2017-01-24
Update aria2 to 1.31.0

rasa:


rasa:
I removed `extract_dir` accidentally.

PR #1277, opened 2017-01-29
Don't remove zip if size changes (fixes #1276)

### LINKED ISSUE #1276: Fails if zip file contains a file of the same name as the zip file
For some reason `ftp://ftp.info-zip.org/pub/infozip/win32/zip300xn-x64.zip` contains a `zip300xn-x64.zip` file inside of it, which https://github.com/lukesampson/scoop/blob/master/lib/install.ps1#L297 erroneously deletes. Patch to come to fix it.

rasa:
See #1276 for details


lukesampson:
Thanks for this. Fixed with 211249238e20e21d5e46b9ae69f9636587db7d1b.

PR #1279, opened 2017-01-30
Fix scoop-reset's undefined `get_config` error

mrkishi:
`scoop reset` is currently erroring at an undefined `get_config` call:

```
$ scoop reset $app
resetting $app ($version)
get_config : The term 'get_config' is not recognized as the name of a cmdlet, function, script
file, or operable program. Check the spelling of the name, or if a path was included, verify that
the path is correct and try again.
At $path\scoop\apps\scoop\current\lib\install.ps1:638 char:8
+     if(get_config NO_JUNCTIONS) { return $versiondir }
+        ~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (get_config:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

In this PR I just followed the other commands' structure and sourced `lib/config.ps1` from `scoop-reset.ps1`. Feel free to close this if it's not the correct approach -- I just thought it was a small enough change to risk it.

As you can see, however, the problem actually comes from `lib/install.ps1`. Perhaps we should consider moving dependencies to the actual usage file? I'm not familiar enough with Powershell to know if that'd be better or how to "dot source guard" double inclusions.

### DIFF
--- a/libexec/scoop-reset.ps1
+++ b/libexec/scoop-reset.ps1
@@ -10,6 +10,7 @@ param($app)
 . "$psscriptroot\..\lib\help.ps1"
 . "$psscriptroot\..\lib\install.ps1"
 . "$psscriptroot\..\lib\versions.ps1"
+. "$psscriptroot\..\lib\config.ps1"
 
 reset_aliases
 


lukesampson:
Thanks for that, you nailed it.

I'm not sure about a better way to do dependencies either. I think if we dot-sourced directly from the lib files there would double inclusions and circular dependencies.

mrkishi:
So, I did some tests and could replicate include guards in Powershell like so:

```powershell
if (${scoop\lib\reset}) { exit }
${scoop\lib\reset} = $true
```

This (arguably ugly preamble) would prevent double inclusions at the cost of extra early-return dot-sources. The benefits would be explicit per-file dependencies. Would it be worth it? I have no idea.

lukesampson:
That's a cool way to do it.

As to whether it's worth it, I'm thinking probably not. The boilerplate isn't so bad, but there are many interdependencies between various lib files. So we'd probably end up with most lib files being included even when they're not needed for a particular command.

Even if we add a `require` function to `core.ps1` to do a similar thing, we'd still need to work out all the dependencies, which would require a lot of care.

To get the same result and avoid this issue coming up again we could just have an `includes.ps1` which includes every lib file. It's not as cool, but it would save having to work out what all the interdependencies are.

But I don't even think the `includes.ps1` is worth it, since this problem has only come up a couple of times in a few years, and it's fairly easily fixed.

PR #1286, opened 2017-01-30
Remove -SkipLast to restore Powershell 3 compatibility (fixes #1278)

### LINKED ISSUE #1278: Autoupdate throws error on Windows 7
[The `Select -SkipLast 1` filter](https://github.com/lukesampson/scoop/blob/master/lib/autoupdate.ps1#L144) in `autoupdate.ps1` throws an error on Windows 7. The `-SkipLast` parameter isn't available in PowerShell 3.0.

However, I was able to perform an automatic update for one of my custom apps. But my `autoupdate` definition doesn't use any of the `$match*` variables. They probably wouldn't work because of this error.

r15ch13:
Fix for #1278


r15ch13:
I messed that up :neutral_face:

Issue #1289, opened 2017-01-30
Possible search improvements?


mrkishi:
Hello, folks.

I just installed `ripgrep` with scoop and in doing so came across a couple of usability problems (imo). These problems arose because it is actually named `rg` in scoop:

```
> scoop search ripgrep

>
```

1. Searches without results just return to the shell -- it took me a second to realize there wasn't a problem, just no matches.
2. There are no descriptions or alternative names for scoop to search through. Even though it's officially named `ripgrep`, one has to search for `rg`. The same thing happens (coincidentally) to `The Silver Searcher` aka `ag`, for instance.

I can open a PR for the first point if there's no objection.

The second one is a bit trickier, though. I could argue `rg` should be renamed to `ripgrep` providing the `rg` binary (I still think that should probably happen), but I feel like in the long run it's probably a good idea to recommend manifest creators to add a short description field and start also providing search results from that.

rasa:
There is an undocumented [`description`](https://github.com/lukesampson/scoop/blob/master/schema.json#L354) field. We may want to look at (bower)[https://github.com/bower/spec/blob/master/json.md] or similar json based schemas and move towards a similar naming conventions (`description`, `keywords`, `name`, etc.)

lukesampson:
I think renaming `rg` to `ripgrep` would be a good move.

I'm not so sure about the `description` field, because I think you usually either know what you're installing or the `homepage` will do a better job of describing it.

`keywords` could be a good idea, but then manifest creators might feel the need to add descriptive tags like "java text search engine" etc, and I think that sort of broad search-ability is unnecessary because people usually know more precisely what they want to install and they just want to find the manifest name.

I think an `aka` property would be good for the cases you outlined where the program is also known as different things that wouldn't otherwise be matched in a search. And it's more clear that package maintainers don't always need to use it. And as you pointed out, in the case of `rg` being renamed to `ripgrep` but retaining the `rg` shim then it wouldn't even be necessary.

talha131:
> I'm not so sure about the description field, because I think you usually either know what you're installing or the homepage will do a better job of describing it.

> keywords could be a good idea, but then manifest creators might feel the need to add descriptive tags like "java text search engine" etc, and I think that sort of broad search-ability is unnecessary because people usually know more precisely what they want to install and they just want to find the manifest name.

In my case, I use tools like brew and scoop to search for new better tools. These tools are not just installation helper but also serve as curated list of apps.

[...]

PR #1292, opened 2017-01-31
Notify empty results for scoop-search

mrkishi:
As the first point in #1289 is a simple change, I went ahead with it.

Let me know if it should be changed or if it's unwanted. 👍

### DIFF
--- a/libexec/scoop-search.ps1
+++ b/libexec/scoop-search.ps1
@@ -114,7 +114,9 @@ function search_remotes($query) {
 }
 
 if (!$local_results -and !(github_ratelimit_reached)) {
-    search_remotes $query
+    $remote_results = search_remotes $query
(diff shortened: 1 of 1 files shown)


lukesampson:
I think it might be better to `[console]:error.writeline($message)` and return 1 as the exit code, so that the standard output remains empty if there are no matches, and it doesn't print in red as it's not really an error from the user's perspective.

Also can you make the message just "no matches found"? The search is not just matching apps, but also executables shim'ed by the app.

mrkishi:
Would you believe me if I told you I started with "writeline + exit 1" but saw abort somewhere and changed my mind? 😄 

Updated.

lukesampson:
Haha, yes I would believe you. Thanks for this!

PR #1319, opened 2017-02-04
Ensure shortcut set explicit working directory

wangzq:
Ideally we should let the manifest specify a working directory, but for now let's at least make sure it is by default set to the directory of the shortcut's target.

### DIFF
--- a/lib/shortcuts.ps1
+++ b/lib/shortcuts.ps1
@@ -22,6 +22,7 @@ function startmenu_shortcut($target, $shortcutName) {
     $wsShell = New-Object -ComObject WScript.Shell
     $wsShell = $wsShell.CreateShortcut("$scoop_startmenu_folder\$shortcutName.lnk")
     $wsShell.TargetPath = "$target"
+    $wsShell.WorkingDirectory = Split-Path $target
     $wsShell.Save()
 }
 


lukesampson:
What are the side effects of not having a working directory set?

wangzq:
The working directory will be "random" and some programs which need to read some files from the base directory will not work. After a second thought, while setting this always to the directory of the target should be fine and helps for those programs, it may still not work for all programs and it is not clean. I am thinking about adding a third item to the shortcut array like this: `["path", "Name", "workdir"]`, of course the third one will be optional. One small concern about this is that since it is an array instead of an object, it is not easy to tell which one is for which, may we could also optionally support using an object notation, I am thinking just use all the property names of the shortcut object as-is:

```
Arguments        Property   string Arguments () {get} {set}
Description      Property   string Description () {get} {set}
FullName         Property   string FullName () {get}
Hotkey           Property   string Hotkey () {get} {set}
IconLocation     Property   string IconLocation () {get} {set}
RelativePath     Property   string RelativePath () {set}
TargetPath       Property   string TargetPath () {get} {set}
WindowStyle      Property   int WindowStyle () {get} {set}
WorkingDirectory Property   string WorkingDirectory () {get} {set}
```

Then we can just apply the properties to the created shortcut object by enumerating all the properties of the shortcut object in manifest.

Let me know which approach you prefers.

Thanks

wangzq:
@lukesampson  P.S. Can you share info on how do you test scoop while developing it? Thanks

wangzq:
I used `set-alias scoop c:\github\scoop\bin\scoop.ps1` to test the changes; I will abandon this and create a new PR with the other approach I mentioned above.

r15ch13:
[OBS Studio](https://obsproject.com/) need a correct WorkingDirectory to start properly.

Rufus doesn't start in portable mode it can't find `rufus.ini` in it's working directory (which is set to `C:\Windows\System32\C:\Windows\System32\` when started from start menu)
See https://github.com/lukesampson/scoop-extras/pull/579

Every program installed on windows sets the working directory the same as the main executable. I think we should default to this behavior.

r15ch13:

PR #1320, opened 2017-02-04
Support optional properties for shortcut

wangzq:


lukesampson:
I much prefer being selective about which properties are used. It's important that adding a shortcut to a manifest is approachable—it should be easy to get right and shouldn't require looking up the Windows API. This is why I asked for an example of where the working directory is required, because if it ain't broke don't fix it. Is there a specific Scoop install that doesn't work because the working directory isn't set for the shortcut?

wangzq:
I am currently trying to create a manifest for [FreeMind] (http://sf.net/projects/freemind) which is a Java application and uses batch file to open, its batch file will only work when the working directory is the script's base directory. I understand this is really a bug in itself, but on the other hand I also think having the ability to configure all the shortcuts' properties in Scoop is something worth doing. Please feel free to close this if you disagree and I will find another workaround. Thanks.

lukesampson:
Ah okay. The idea was that shortcuts would point to a GUI .exe (like Atom, Notepad2 etc), although that's not explicit in the documentation. I don't think these sort of programs need all the extra shortcut properties.

Simplicity is a big priority here, rather than supporting every feature of the Win API. So I'd prefer not to add all these extra shortcut properties.

If you do want to use the .bat as the Start Menu shortcut, perhaps you could use a `pre_install` which does something like this to write a new batch file to the install directory which sets the correct directory...

```json
{
...
"pre_install": ["cd %~dp0 >> start.bat", "call freemind.bat >> start.bat" ],
"shortcuts": [ "start.bat", "Freemind" ]
...
}
```

I'm not sure if that's correct code, but maybe something like that would work.

wangzq:
No problem. Yes pre_install should be able to fix this.

r15ch13:
[OBS Studio](https://obsproject.com/) need a correct WorkingDirectory to start properly.

r15ch13:

PR #1331, opened 2017-02-12
[WIP] Validate manifests against JSON Schema in CI Tests

r15ch13:
I am implementing manifest validation against the JSON Schema using dlls from [Newtonsoft.Json.Schema](http://www.newtonsoft.com/jsonschema).
Currently it works on my machine, but it fails on appveyor. (See [my build log](https://ci.appveyor.com/project/r15ch13/scoop/build/8-schema-validation) or [the log of this PR](https://ci.appveyor.com/project/lukesampson/scoop/build/1906-master))

Could this have anything to do with different powershell or .NET versions?
PS on appveyor:
```
Major  Minor  Build  Revision
-----  -----  -----  --------
5      0      10586  117
```
PS on my machine:
```
Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      14393  693
```

I have no idea why this error `Cannot find an overload for "IsValid" and the argument count: "3".` pops up.

Any ideas? 😄


r15ch13:
Nope doesn't work either

rrelmy:
Sorry, I can't help here.
I have no experience with AppVeyor or. NET ;-)

lukesampson:
Is it possible that the AppVeyor build environment already has an older version of `Newtonsoft.Json.Schema` installed and it's taking precedence even after you `Add-Type` the specific version?

Is it worth adding something like this after you run the `Add-Type` to check the file path of the loaded assembly?
```
[appdomain]::currentdomain.getassemblies() |? { $_.getname().name -eq 'Newtonsoft.Json.Schema' } |% { $_.codebase }
```

r15ch13:
Nice good idea!
https://ci.appveyor.com/project/r15ch13/scoop/build/15-test#L181
Newtonsoft.Json 8.0 is already there 👎

deevus:
Try calling it using the static class?

`[NewtonSoft.Json.Schema.SchemaExtensions]::IsValid($manifest)`

test/Scoop-Manifest.Tests.ps1
```diff
@@ -3,41 +3,83 @@
...
+        it "fails with broken manifest" {
+            $json = gc "$working_dir\broken_wget.json" -raw -Encoding UTF8
+            { [Newtonsoft.Json.Linq.JToken]::Parse($json) } | should throw
+        }
+        it "fails with invalid manifest" {
+            {
+                $json = gc "$working_dir\invalid_wget.json" -raw -Encoding UTF8
+                $manifest = [Newtonsoft.Json.Linq.JToken]::Parse($json)
+                $schema = [Newtonsoft.Json.Schema.JSchema]::Parse($schema_json)
+                [Newtonsoft.Json.Schema.SchemaExtensions]::IsValid($manifest, $schema, [ref]$validationErrors)
```
r15ch13: Already doing that

r15ch13:
Okay i have written a small c# tool for validating the manifests.
Calling a process for each manifest increased the duration from around 10-20ms to 250-260ms (on an SSD) per test ...

r15ch13:
Curse you Appveyor!

lukesampson:
Wow, that's frustrating. Could it be an x86/x64 difference between what AppVeyor is using and what the Scoop.Validator.dll is targeting? Would setting a platform in appveyor.yml make a difference, or making Scoop.Validator.dll match whatever platform the Newtonsoft dlls target?

rasa:
Would you guys consider using python for this test? https://github.com/rasa/scoops/blob/master/validate.py is MIT licensed.

r15ch13:
Ah nice 👍 
Results from trying it out:
* changed validate import to Draft4Validator (because we use the "required" field)
* writing the exception stacktrace to .failed raises an exception 😄 (using Python 3.6)
```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "validate.py", line 51, in <module>
    f.write(trace)
TypeError: a bytes-like object is required, not 'str'
```

rasa:
Sorry, I've only tested it with Python 2.7.13. Remove the `b` in line 50:
````
with open(failed, 'a+') as f:
````
and it should work in Python 3

**Comment by r15ch13** [comment]
Yup that worked. I'm going to redo this in python tomorrow or so. I hope that works on appveyor. 😀

**Comment by r15ch13** [comment]
Apparently git has desided to destroys the dll files:
```
warning: LF will be replaced by CRLF in supporting/validator/Newtonsoft.Json.dll.
The file will have its original line endings in your working directory.
```

r15ch13:
Eureka! It works!

rasa:
Two thoughts:
1. Was removing `UniqueItems` intentional? 
2. Requiring `version` means we can't have pure meta-packages, with just a `depends` list.

r15ch13:
1. I have no idea why i removed `uniqueItems` 😕 Will add it back.
2. Without `version` the installation won't run at all (see [lib/install.ps1#L20](https://github.com/lukesampson/scoop/blob/master/lib/install.ps1#L20))

PR #1333, opened 2017-02-13
Exclude bitbucket from URL check (like github)

phitsc:


rrelmy:
I think we can remove the check completely, it seems not many of the big sites support HEAD requests.

Or has anyone a better idea to check if the url is correct without downloading the complete file?

lukesampson:
Do we always need to download the whole file anyway to generate the hash, @rrelmy? If that's the case, then I agree it would be good to remove this `check_url` function.

I'll merge this in since it's consistent with what's already there.

rrelmy:
@lukesampson most of the time the file will be downloaded for the hash generation.
But not always, there is support to extract the hash from the webpage or a separate hash file (example: [ant](https://github.com/lukesampson/scoop/blob/master/bucket/ant.json))

I don't have stats on how much of the autoupdates download the file anyway but I think it's the majority.

PR #1342, opened 2017-02-18
Improvements for schema.json

r15ch13:
* Use architecture definition
* Arrays of string shouldn't be empty
* Add back uniqueItems directive
* match version against pattern from [lib/install.ps1#L21](https://github.com/lukesampson/scoop/blob/master/lib/install.ps1#L21))
* `extract_dir` in autoupdate only works with string (at the time)
* reindent


rrelmy:
I can't test it right now but it looks good to me

PR #1347, opened 2017-02-24
cmake 3.8 update

jfmherokiller:


r15ch13:
Was already added to the extras bucket with this pr https://github.com/lukesampson/scoop-extras/pull/361

lukesampson:
Thanks for this, but we try to keep the main bucket for stable releases only. As @r15ch13 mentioned, @berwyn's 3.8-rc1 version is already in extras. You can install it with `scoop bucket add extras; scoop install cmake-rc`.

I'll close this PR for now, but please submit 3.8 again when it goes stable.

jfmherokiller:
thanks for letting me know but how do i install the cmake version in the extras bucket because i need it for vcpkg

lukesampson:
```
scoop bucket add extras
scoop install cmake-rc
```

jfmherokiller:
thanks

PR #1351, opened 2017-02-27
Add fossil 1.37

bpollack:
Fossil is the version control system used by sqlite, among others. I didn't quite get autoupdate working, because I can't find a single page that has only one version number on it, which is causing checkver to behave a bit unpredictably. That said, they do like four releases a year, tops; I'll just do this by hand.


lukesampson:
Okay sure. Thanks for adding this 👍

PR #1352, opened 2017-02-27
Add cacert.pem as its own package

r15ch13:
Maybe it could now be added as a dependency to wget and curl


lukesampson:
Sounds good!

PR #1357, opened 2017-02-28
Update curl with cacert's pem file

rasa:
Per https://github.com/lukesampson/scoop/pull/1352#issue-210426165


r15ch13:
Nice one! But the suggest field should look like this:
```
"suggest": {
    "cacert": [ "cacert" ]
}
```

rasa:
Dang. Sorry, will fix now.

PR #1385, opened 2017-03-19
Add referer-header to all webrequests

r15ch13:
A solution for #368 but needs testing

This should not effect any sites, because browsers will send in anyway.

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -94,6 +94,7 @@ $queue | % {
         jsonpath = $jsonpath;
     }
 
+    $wc.headers.add('Referer', $url)
     $wc.downloadstringasync($url, $state)
 }
 
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -94,7 +94,7 @@ $queue | % {
         jsonpath = $jsonpath;
     }
 
-    $wc.headers.add('Referer', $url)
+    $wc.headers.add('Referer', (strip_filename $url))
     $wc.downloadstringasync($url, $state)
 }
 
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -25,7 +25,7 @@ function check_url([String] $url) {
     }
 
     try {
-        $response = Invoke-WebRequest -Uri $url -Method HEAD
+        $response = Invoke-WebRequest -Uri $url -Method HEAD -Headers @{'Referer' = $url}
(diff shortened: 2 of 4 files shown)


lukesampson:
It seems strange to me to set the referrer to the same URL that is being downloaded—would a browser do that? Maybe we could set it to the base URL with the path stripped?

lukesampson:
What do you think about setting a "referer" property in the manifest (or "referrer" to use the normal spelling), similar to the "cookie" property used [here](https://github.com/lukesampson/scoop-extras/blob/master/oraclejdk.json#L22) to set a browser cookie?

r15ch13:
An extra property would add more clutter to the `dl_with_cache`, `do_dl` and `dl` functions in `install.ps1`.
Just stripping the filename of the url should be enough.

lurdan:
It seems that some site blocks to fetch because of this change (adding referer, assuming same domain?).

I want to have something in the manifest to skip this.

lukesampson:
Thanks for reporting the problem.

* What are the download URLs that fail?
* Can you identify or share the manifests that exhibit this problem?
* Were you able to confirm that removing the referrer makes the download succeed?

lurdan:
> What are the download URLs that fail?
> Can you identify or share the manifests that exhibit this problem?

https://github.com/lurdan/scoop-bucket/blob/master/keyswap.json
I've found out this manifest being broken when setup new computer.
(ftp.vector.co.jp is a software collection site for japanese users)

> Were you able to confirm that removing the referrer makes the download succeed?

I've tested with wget.
- Success: wget http://ftp.vector.co.jp/53/91/2331/keyswap.zip
- Fail: wget http://ftp.vector.co.jp/53/91/2331/keyswap.zip --referer "http://ftp.vector.co.jp"

comp500:
This still appears to be a problem, for a site I'm using: [https://files.sshnuke.net/cemuhook_1114_0563.zip](https://files.sshnuke.net/cemuhook_1114_0563.zip)

It seems to break with `$wreq.referer = strip_filename $url`, but not with `$wreq.referer = $url` or if the referer is commented out in install.ps1.

[Here is my manifest](https://github.com/comp500/scoop-comp500/blob/master/cemuhook.json), which works completely fine if the referer isn't `strip_filename $url`. Wireshark analysis just indicates that the website does a 302 if the referer is incorrect.

Making the referer configurable (even just turning it off would work) would be a workaround to this problem.

rasa:
See also https://github.com/lukesampson/scoop-extras/issues/683

PR #1386, opened 2017-03-19
remove unncessary setting of $HOME env var which causes other issues

yunspace:
I've been experiencing quite a bit of issues with the setting of $HOME env var causing failures in building Unreal Engine HTML5 WebGL: https://answers.unrealengine.com/questions/457972/could-not-find-file-epic-games412engineintermediat.html

Follow their support team's suggestions, I look around in my `scoop` installations for where $HOME was being set. And I noticed that `env_ensure_home` is commented out and the general direction in scoop is not to set $HOME:

https://github.com/lukesampson/scoop/blob/a6f6e5425f873098a565b4e85b7639053d0c894e/lib/install.ps1#L840

However I did find $HOME still being set in `openssh` which I have installed in my system. So that's what this PR is about.

### DIFF
--- a/bucket/openssh.json
+++ b/bucket/openssh.json
@@ -26,6 +26,5 @@
         "bin\\ssh-keygen.exe",
         "bin\\ssh-keyscan.exe",
         "bin\\ssh.exe"
-    ],
-    "env_set": { "home": "$home" }
+    ]
 }


lukesampson:
Thanks for this. That does look like a mistake in the manifest, as the "$home" variable would be undefined, I think. Can you confirm—was it setting $env:home to null/empty?

yunspace:
Thanks, I will try to confirm. 

Also a bit more context on my scenario:

- I have a `boostrap.ps1` that installs a whole heap of scoop apps and I don't really know which app really is setting the `$HOME`. But I do know for sure that after I run my `bootstrap.ps1` a `$HOME` does get set.
- by using trial and error, installing apps one by one, plus github search I think I've narrowed the code to `openssh`
- once the `$HOME` is set, even if I unset it, my Unreal WebGL build still wouldn't work. So I need to rollback to a system restore point. Which is very strange indeed.

Now that this is merged, I will also verify if I run my whole `bootstrap.ps1`, the $HOME should not be set.

lukesampson:
Okay, thanks.

You probably already did this, but make sure to remove the `HOME` environment variable through system preferences (machine and user) to be sure, because if you set `$env:home` or `rm env:\home` it only affects the current PowerShell session.

PR #1393, opened 2017-03-22
update pshazz – adds ssh hostname completion

rrelmy:
I assume the version of pshazz is just the date?


lukesampson:
Yes, that's right :)

PR #1396, opened 2017-03-23
Update dotnet to 1.1.1

inosik:
This updates the `dotnet` manifest to v1.1.1. Or, at least that's the version the ZIP-file has in it's name.

@yunspace @jggrant do these changes look right to you?

### DIFF
--- a/bucket/dotnet.json
+++ b/bucket/dotnet.json
@@ -1,14 +1,14 @@
 {
-    "version": "1.0.0-preview2-1-003177",
+    "version": "1.1.1",
     "homepage": "https://www.microsoft.com/net/core#windows",
     "architecture": {
         "64bit": {
-            "url": "https://go.microsoft.com/fwlink/?LinkID=834991#/dl.zip",
-            "hash": "D4056A91F83699894C9F9FBAC1B8965597DED079B0D8848DACC7505F372F85A0"
+            "url": "https://go.microsoft.com/fwlink/?linkid=843454#/dl.zip",
+            "hash": "e729afcf3cc69f17ec7968468b399c843b8b8327523e62c03450e4653115cf76"
         },
         "32bit": {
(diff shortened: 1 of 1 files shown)


lukesampson:
Have you noticed how `dotnet --version` reports version 1.0.1? It's confusing that the .zip has 1.1.1, but I think we should probably go with what the exe reports instead. Thoughts?

rrelmy:
OMG, that's a mess. [Microsoft](https://www.microsoft.com/net/download/core#/sdk) calls it **.NET Core SDK (contains .NET Core 1.0 and 1.1)**.
Inside `shared/Microsoft.NETCore.App` are two directories for 1.0.4 and 1.1.1 also with a `dotnet.exe` too. 🤣 
If one of those binaries would be interesting the `bin` reference should be updated inside the manifest.

_The dotnet.exe binaries inside `/shared/` do not output a version if executed with the `--version` flag._ 🙈

lukesampson:
Errrrr.... okay! Thanks for looking into it @rrelmy. Maybe `1.1.1` isn't such a bad idea after all??

rrelmy:
Without `--version` the dotnet.exe in the root reports version 1.1.0. The binary inside the shared/1.1.1 directory should the same version and build, but not with `--version` and has also a different hash.
Can someone explain?

dotnet-whatthefuck

lukesampson:
It looks like [we're not the only ones confused](https://github.com/dotnet/cli/issues/3773)! 

It seems like `--version` is the CLI version, and the version printed without args is the version of the shared framework.

I'm still not sure what version we should use for Scoop...

rrelmy:
Unbelievable, what a mess.  Ok for me to merge, even I am not really sure if the version should be 1.1.0 or 1.1.1.

As inosik said, 
>at least that's the version the ZIP-file has in it's name

inosik:
Thanks for merging!

> Unbelievable, what a mess.

The department for version confusion in Redmond did a really good on this one :smile:

PR #1397, opened 2017-03-23
idea scoop manifest

gro-mar:


lukesampson:
Is this a trial version? It might be better in the [scoop-extras](/lukesampson/scoop-extras) bucket if you'd like to submit it there. 

Also, you're missing a newline at the end of the file which is why the checks failed.

It'd be good if you could add a couple more things to the manifest, like a "homepage" and "shortcuts". Check out the manifest reference [here](https://github.com/lukesampson/scoop/wiki/App-Manifests) for more detail.

rrelmy:
Yes, that seems to be the Ultimate version. Probably better of in the extra bucket and called `idea-ultimate`.

I think we would run into several problems if the integrated updater is used.

I would recommend using the official [JetBrains Toolbox App](https://www.jetbrains.com/toolbox/app/) to install those IDEs instead of using scoop.

rrelmy:
I just found [Webstorm](https://github.com/lukesampson/scoop-extras/blob/master/webstorm.json) (another Jetbrains IDE) inside the extras bucket 😉 

I close this pull request, if you still want to add IntelliJ please do it inside the extras bucket and base it on the webstorm manifest with autoupdate.

PR #1402, opened 2017-03-26
Update URLs for Yarn

Daniel15:
The Yarn package should use the official download URLs rather than the GitHub URLs 😃

### DIFF
--- a/bucket/yarn.json
+++ b/bucket/yarn.json
@@ -11,14 +11,14 @@
             "nvm"
         ]
     },
-    "url": "https://github.com/yarnpkg/yarn/releases/download/v0.21.3/yarn-0.21.3.msi",
+    "url": "https://yarnpkg.com/downloads/0.21.3/yarn-0.21.3.msi",
     "hash": "3b4e34f8a3b3c4cb63f2da5411fe7a7977f35e65679b45b68773de4ed24a0d1c",
     "bin": "Yarn\\bin\\yarn.cmd",
     "checkver": {
-        "url": "https://github.com/yarnpkg/yarn/releases/latest",
-        "re": "/releases/tag/v([\\d.]+)"
+        "url": "https://yarnpkg.com/latest-version",
+        "re": "([\\d.]+)"
     },
     "autoupdate": {
-        "url": "https://github.com/yarnpkg/yarn/releases/download/v$version/yarn-$version.msi"
+        "url": "https://yarnpkg.com/downloads/$version/yarn-$version.msi"
     }
 }


deevus:
Thanks @Daniel15 :)

deevus:
Just saw this:

image

:+1:

Daniel15:
Yeah, @tech4him1 added that section to the Yarn site (https://github.com/yarnpkg/website/pull/425) which prompted me to take a look at how Scoop downloads Yarn. I built the packaging (including the Windows installer) for Yarn 😃

deevus:
Awesome! 

How are you going with the autoupdate process? I just tried to force an autoupdate locally and I'm getting a 403?

```
$ bin/checkver yarn -u -f
yarn: 0.21.3
Forcing autoupdate!
Autoupdating yarn
The remote server returned an error: (403) Forbidden.
URL https://yarnpkg.com/downloads/0.21.3/yarn-0.21.3.msi is not valid
Could not update yarn
```

Daniel15:
Huh, that's strange... That URL is valid and should work. It works when I hit it. I didn't try forcing an auto-update, I just checked that the version number was correct.

Daniel15:
Can you hit https://yarnpkg.com/downloads/0.21.3/yarn-0.21.3.msi in your browser or do you get an error?

deevus:
Yeah I can hit it in the browser. Just fails on the cmd. Tried it a couple of times. We might be using a specific user-agent.

deevus:
https://github.com/lukesampson/scoop/blob/master/bin/checkver.ps1#L50

Daniel15:
Huh, that's interesting... `scoop install` seems to be fine though, at least when I was testing.

deevus:
Yeah, I noticed that too. It seems that we're using a similar UA there 

```
 $ pt user-agent /i
lib\core.ps1
97:    $wc.headers.add('User-Agent', 'Scoop/1.0')

bin\checkver.ps1
50:    $wc.Headers.Add("user-agent", "Scoop/1.0 (+http://scoop.sh/) (Windows NT 6.1; WOW64)")
```

Daniel15:
I tried a request using Postman and it seems fine:


I wonder if there's other headers that the server doesn't like. yarnpkg.com is hosted by Netlify with CloudFlare in front of it, perhaps CloudFlare doesn't like something about the request 😕 

We can revert it back to the GitHub URL if we can't figure it out. Really strange.

deevus:
If what I was reading is correct and you're looking at automating the manifest updates yourselves: as long as it works for you it's possibly not an issue.

I'll try again later to see if there's just something weird going on with my connection.

deevus:
I've found the problem. The autoupdate code tries a `HEAD` request to validate the url before downloading anything. The `HEAD` method is obviously blocked.

Daniel15:
Ahh, good catch! Looks like that's an existing issue - The 403 on `HEAD` is coming from GitHub. I see the same thing when I do a HEAD request to the GitHub URL (`https://github.com/yarnpkg/yarn/releases/download/v0.21.3/yarn-0.21.3.msi`) too.

Daniel15:
Ahh, good catch! Looks like that's an existing issue - The 403 on `HEAD` is coming from GitHub. I see the same thing when I do a HEAD request to the GitHub URL (`https://github.com/yarnpkg/yarn/releases/download/v0.21.3/yarn-0.21.3.msi`) too. The yarnpkg.com URL just redirects to the GitHub release, we just use it so we're not tightly-coupled to GitHub's release mechanism.

deevus:
Fair enough. How annoying! 😅

PR #1403, opened 2017-03-26
Fix yarn autoupdate

deevus:
- Added yarnpkg.com to HEAD request blacklist
- General refactor
- See #1402


rrelmy:
Maybe we should get rid of this HEAD request. It makes more problems than it solves.

Could do a GET request instead and abort it as soon we have the HTTP Status?

deevus:
@rrelmy You're probably right.

Issue #1372, opened 2017-03-13
Discussion: Persistent data, update improvements


rrelmy:
I want to start discussion about enhancing the update functionality solving multiple issues

## Persistent data
 - Folders per Version #126
 - PHP configuration directory along versions leads to problems #1185
 - I think npm global packages should be installed in it's own directory #1220

Multiple solutions were suggested for this. [Travelling directories](/lukesampson/scoop/wiki/Travelling-directories) moves data between versions, already partially implemented (not working).
A `backpack`/`data`/`conf` directory alongside the `apps` directory.

Travelling directories should be easy to implement but has very big problem. When an update fails you are required to uninstall the app and install it again, if that happens you will lose your data. (See last section)

Using a custom data directory may not be possible for all apps, the app must support changing the default directories. Could that work with conjunctions/symlinks?

We may have to implement both.
Any comments or ideas?

**Affected apps**:
I would like to create a list of affected apps in this issue
 - php, configuration
 - mysql, database content
 - mariadb, database content
 - nodejs, globally installed packages
 - python, pip packages
 - ruby, gems
 - *this list is incomplete*

## Development/testing problematic

Testing those update paths is currently not easy.
Installing updated manifests with `scoop install bucket/app.json` does not take the same code path as `scoop update app`.
Using `scoop update bucket/app.json` does not work.

Prevent such updates through `scoop install` and make them work with `scoop update` should be pretty straightforward.
What do you think about that?

## Update failuers

I never liked the fact that failures during updates force you to uninstall the app and reinstall it completely.
While generally looking at the update procedure I would like to check if there are some possible improvements.

For example try to download the update before uninstalling the current one would catch some basic errors.
Or if possible remove the shims just before the new ones are set.

r15ch13:
Something for the List
* nginx, configuration

Testing updates is quite easy, if they are in the main bucket.
Just make your changes and run `.\bin\scoop.ps1 install <updatedappname>` from the repo directory.
If the changes are made in the extras bucket, you can add your local repository as a bucket with `scoop bucket add X:\scoop-extras extras-testing`

[...]

PR #1410, opened 2017-03-29
Persistent data

rrelmy:
Based on the discussion in #1372 

Any objections? 😅


lib/install.ps1
```diff
@@ -44,6 +46,10 @@ function install_app($app, $architecture, $global, $suggested) {
     if($global) { ensure_scoop_in_path $global } # can assume local scoop is in path
     env_add_path $manifest $dir $global
     env_set $manifest $dir $global
+
+    # persist data
+    persist_data $manifest
```
lukesampson: I think it would be safer and clearer to explicitly pass `$original_dir`, `$persist_dir` variables here, instead of having `persist_data` inherit them from the parent scope.

lib/install.ps1
```diff
@@ -786,10 +796,21 @@ function find_dir_or_subdir($path, $dir) {
     return [string]::join(';', $fixed), $removed
 }
 
+function prepare_env_path ($path) {
```
lukesampson: Would really love to not have to add this function unless it's absolutely necessary.

lib/install.ps1
```diff
@@ -786,10 +796,21 @@ function find_dir_or_subdir($path, $dir) {
...
+    return $path_dir
+}
+
 function env_add_path($manifest, $dir, $global) {
     $manifest.env_add_path | ? { $_ } | % {
-        $path_dir = "$dir\$($_)"
-        if(!(is_in_dir $dir $path_dir)) {
+        $path_dir = prepare_env_path $_
+
+        if(!(is_in_dir $dir $path_dir) -and !(is_in_dir $persist_dir $path_dir)) {
```
lukesampson: If `prepare_env_path` can be removed, it would be good to not have to change this either.

lib/install.ps1
```diff
@@ -820,7 +842,7 @@ function env_set($manifest, $dir, $global) {
     if($manifest.env_set) {
         $manifest.env_set | gm -member noteproperty | % {
             $name = $_.name;
-            $val = format $manifest.env_set.$($_.name) @{ "dir" = $dir }
+            $val = format $manifest.env_set.$($_.name) @{ "dir" = $dir; "persist_dir" = $persist_dir }
```
lukesampson: If prepare_env_path can be removed, it would be good to not have to change this either.

lib/install.ps1
```diff
@@ -941,12 +963,61 @@ function show_suggestions($suggested) {
...
+        $target = fname($source)
+    }
+
+    return $source, $target
+}
+
+function persist_data($manifest) {
+    $persist = $manifest.persist
+    if($persist) {
+        $persist_dir = ensure $persist_dir
```
lukesampson: As above for passing this variable explicitly rather than inheriting it from parent scope.

libexec/scoop-uninstall.ps1
```diff
@@ -17,9 +18,10 @@
 reset_aliases
 
 # options
-$opt, $apps, $err = getopt $args 'g' 'global'
+$opt, $apps, $err = getopt $args 'gp' 'global', 'purge'
```
lukesampson: 👍 `--purge` option

r15ch13:
For nginx i suggest adding: `"notes": "To use the persistent configuration run 'nginx -p \"$env:SCOOP/persist/nginx\"' or 'nginx -p \"%SCOOP%/persist/nginx\"'"` and also include the `temp` folder.

`php.ini-production` should only be copied over if `php.ini` doesn't exist.

lib/install.ps1
rrelmy: Removed 👍

rrelmy:
I merged all the commit's inside your branch.
e7814ea295744fbf6a140c83ac1490cfaa9ebccd removes all the persist related stuff from the env variables.

I am not happy with the php version.
The persisted php.ini breaks my setup and will also break others. [Documentation](/lukesampson/scoop/wiki/Custom-PHP-configuration)
Normally the php.ini is not within this conf directory. The conf directory is there to overwrite configs from the php.ini.
My current setup has multiple files (debug.ini, default.ini, extensions.ini etc.) inside the conf directory. Copy the php.ini to there would overwrite my configuration

Ideally we would copy the `php.ini-production` to `php.ini` and persist it as `php.ini` inside `persist/php/` (without `conf`!).
The mysql manifest is similar.

rrelmy:
What do you think of an optional object syntax for `persist` described in https://github.com/lukesampson/scoop/issues/1372#issuecomment-289600813

We could do somethink like

``` json
{
    "persist": [
        {
            "source": "php.ini",
            "target": "php.ini", // can be omitted because it's the same as source
            "boilerplate_if_not_exists": "php.ini-production"
        }
    ]
}
```

r15ch13:
Another idea would be to use the standard Debian/Ubuntu behavior, which would result in:

`persist\php\cli\php.ini`
`persist\php\cli\conf.d\custom.ini`

Manifest:
```
"persist": [
        "cli",
        [
            "php.ini-production",
            "cli\\php.ini"
        ]
    ],
    "env_set": {
        "PHP_INI_SCAN_DIR": "$dir\\cli;$dir\\cli\\conf.d;"
    },
    "post_install": "
# Create directory for custom PHP configuration
if(!(Test-Path \"$dir\\cli\\conf.d\")) {
    (New-Item -Type directory \"$dir\\cli\\conf.d\") | Out-Null
}

# Copy ini scan dir from old location before persistent update
if ((Test-Path \"$dir\\..\\conf\") -and !(Test-Path \"$dir\\cli\\conf.d\")) {
    Write-Host -ForegroundColor Cyan 'We are moving your php configuration to $dir\\cli\\conf.d'
    Move-Item \"$dir\\..\\conf\" \"$dir\\cli\\conf.d\"
}

# Enable extensions to be found in installation-relative folder (the default is to search C:/php)
(gc \"$dir\\cli\\php.ini\") | % { $_ -replace '; extension_dir = \"ext\"', 'extension_dir = \"ext\"' } | sc \"$dir\\cli\\php.ini\"
",
```

rrelmy:
Approved 👍 
Can you push these change to the PR?

Didn't came to my mind that we are able to add multiple directories to `PHP_INI_SCAN_DIR`

r15ch13:
Done 👍

rrelmy:
@r15ch13 Aren't you able to push it directly to my `persist` branch on `rrelmy/scoop`?
Then it shows up inside this PR.

r15ch13:
Didn't know i could push to your fork Oo

rrelmy:
@r15ch13 You can because you are a maintainer of this repo and I allowed maintainers to pull into my forked branch inside this PR.

That's the little checkbox «Allow edits from maintainers.» when you create a PR ;-)

I think we are ready.
I am doing some final checks and will merge it tomorrow morning if that is ok for you.

rrelmy:
@r15ch13 I never really understood how one would run nginx, when I tested I always got errors.
Thanks to your note I now understand it.

One problem with this note, the `SCOOP` env variable is only available if the user manually set it. There is no such variable set by default.
What about adding a custom one for nginx, and use that in the note?
`"env_set": { "NGINX_HOME": "$dir" }`

And is it possible to use the `%...%` notation? I can't get that to work. 😕

PR #1415, opened 2017-04-01
env_add_path added for maven

onacit:
I'm sorry I don't know how to test.
Please add some wiki entry.

### DIFF
--- a/bucket/maven.json
+++ b/bucket/maven.json
@@ -5,9 +5,7 @@
     "url": "https://archive.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.zip",
     "hash": "bb37146a67d067069754c775b9c9f03b52da953261981738de963057023cda3a",
     "extract_dir": "apache-maven-3.3.9",
-    "bin": [
-        "bin\\mvn.cmd"
-    ],
+    "env_add_path": "bin",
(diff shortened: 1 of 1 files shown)


lukesampson:
You can test a local manifest by passing a file path to `scoop install`, e.g. `scoop install bucket\mvn.json`. Or, if you've branched your fork of Scoop in place of the master branch so that it's at `~/scoop/apps/scoop/current` you can just run `scoop install mvn`.

I'm not familiar with Maven, but could you explain why you're switching from an explicit shim of `mvn.exe` to adding the `bin` directory to the path?

We prefer explicit shims because you can see from the manifest which commands will be added, and these commands become searchable. Whereas with `env_add_path` it's not clear. We would usually only use `env_add_path` when the exact executables are not known at install time, e.g. when a package manager like `npm` or `gem` will place user-installed executables at that location. Is that the case with Maven?

r15ch13:
Possible duplicate https://github.com/lukesampson/scoop/issues/1237

onacit:
I borrowed it from the `ant.json` which put the whole `bin` directory to the path.
Denoting single `mvn.cmd` works find in windows command.
But I couldn't find any way to work with the `mvn` command with my Cygwin.
Adding an alias doesn't work.

Issue #1417, opened 2017-04-03
env_add_path seems to be broken


sakai135:
`scoop install python` ended up adding `C:\Users\user\scoop\apps\python\current\C:\Users\user\scoop\apps\scoop\current\bin\scoop.ps1` as one of the entries to the system `PATH` environment variable.

From the manifest and glancing at the code, I'm guessing that it was supposed to add `C:\Users\user\scoop\apps\python\current\scripts` instead.

rrelmy:
Thanks for the report.

The issue has been fixed, please run `scoop update`, uninstall python and install it again.
Unfortunately you have to remove the broken path from `$PATH` manually as it does not get picked up while uninstalling the app.

Sorry for the inconvenience

PR #1418, opened 2017-04-03
[WIP] Add more unit tests

rrelmy:
In response to #1417 I started to write some basic unit tests for the install functions.

I think we have some functions that are worth to test, what do you think?


deevus:
Oops I just saw the WIP tag. You'll have to create another PR if there are more tests to add :( Sorry

PR #1420, opened 2017-04-04
Add xsv

Congee:


lukesampson:
xsv.json looks good, but do you mind redoing this pull request so it doesn't have the extra commit history and merges?

PR #1435, opened 2017-04-15
FTP SIZE request

asmgf:
Came across this when tried to install zip.
ContentLength is -1 when downloading from their server.


lukesampson:
Thanks!

PR #1437, opened 2017-04-16
Repaired the axel app manifest 

boazj:
Repaired the axel app manifest to use maintainer release (https://github.com/ghuntley/cygwin-axel) instead of dropbox


r15ch13:
Thanks

PR #1441, opened 2017-04-20
Fixed bug where $home doesn't support .endswith

MartyGentillon:
In Windows 10, at least after the Creators update, $h is of type
PathInfo, which does not support String operations.  By toStringing it,
we turn it into a String and enable the expected String operations.

see https://github.com/lukesampson/scoop/issues/1440


lukesampson:
Thanks. I just noticed we should probably be using `$env:userprofile` rather than `$home`, as we do in other places (e.g. locating the Scoop base directory).

PR #1443, opened 2017-04-21
Fix sbt package installation problem

altrive:
When trying to install latest sbt package (` 0.13.15`) .
It fails to copy extracted zip folders `sbt-launcher-packaging-0.13.15`.
It seems archive folder name is changed to "sbt" from v0.13.15.

#### Note:
I've confirmed next pre-release package `1.0.0-M5` use same folder naming.
<https://github.com/sbt/sbt/releases>


lukesampson:
Thanks 👍

PR #1447, opened 2017-04-23
sonarqube-x86-64 added

onacit:
This is a simple manifest for running [SonarQube](https://www.sonarqube.org/) in local machine.

### DIFF
--- a/bucket/sonarqube-x86-64.json
+++ b/bucket/sonarqube-x86-64.json
@@ -1,0 +1,16 @@
+{
+    "homepage": "https://www.sonarqube.org",
+    "version": "6.3.1",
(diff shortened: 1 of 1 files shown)


onacit:
* The zip file contains single directory entry whose name is same with the enclosing output zip file.
* Both architecture(32/64) contains in a single `bin` directory.
* I'm not sure I'm understood how [autoupdate](https://github.com/lukesampson/scoop/wiki/App-Manifest-Autoupdate) works.

lukesampson:
Thanks for this.  I have renamed it to `sonarqube.json` and added the autoupdate for you, as well as an `extract_dir` to shorten the path.

PR #1452, opened 2017-04-27
Nircmd

klaidliadon:
See http://www.nirsoft.net/utils/nircmd.html


lukesampson:
Thanks for this. I've removed nircmd 2.75 from the extras bucket.

PR #1463, opened 2017-05-05
VcXsrv: Open Source Windows X-server based on the xorg git sources

se35710:
I know there is Xming in the bucket, but that is [donation ware](http://www.straightrunning.com/XmingNotes/#head-16).


r15ch13:
I think it might be better to add it to the extras bucket.

se35710:
Okay, I'll create a new pull request in extras.

PR #1465, opened 2017-05-07
Stack 1.4.0

asmgf:


r15ch13:
Thanks, but you forgot to add two `$version`-variables to the update url :smile:

asmgf:
Nice catch. Thank you.

PR #1470, opened 2017-05-11
Added OpenResty, a fork of NGINX for writing webapps with Lua

Guard13007:
Not sure if this is more appropriate here or in extras, but considering Apache and NGINX are here, I thought it fit in. *shrugs*


lukesampson:
It's probably better off in extras because it looks fairly niche. Correct me if I'm wrong. Do you mind submitting this over at extras instead?

Guard13007:
Sorry I have been a bit bad about noticing things lately..for months now. I will do so.

PR #1490, opened 2017-05-15
Adding additional mirror to maven.json

NaanProphet:
In some locations, the main archive mirror is not available. Listing the next most popular one at the time of this update

### DIFF
--- a/bucket/maven.json
+++ b/bucket/maven.json
@@ -2,7 +2,10 @@
     "homepage": "https://maven.apache.org/",
     "version": "3.5.0",
     "license": "Apache 2.0",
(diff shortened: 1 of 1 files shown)


NaanProphet:
Could anyone clarify the usage of multiple URLs correct? It appears it still fails sometimes locally if the first URL host cannot be reached...

r15ch13:
All URLs in this array will be downloaded, they are not for mirroring purposes.
With `location` do you mean country? Would like to check this via VPN then :smile:

NaanProphet:
Ohh I see...thanks! No, by "location" I was referring to particular networks. Maybe it is a VPN thing? Hmm.

Would be any interest in a new feature request for failover URLs?

rasa:
This PR can be closed.

PR #1493, opened 2017-05-16
Add Allure Framework commandline tool 

ehborisov:
Hello, we would really appreciate if you could add our commandline tool for the [Allure reporting framework](https://github.com/allure-framework) to Scoop. Thanks!

### DIFF
--- a/bucket/allure.json
+++ b/bucket/allure.json
@@ -7,15 +7,15 @@
     "extract_dir": "allure-2.0.1",
     "bin": "bin\\allure.bat",
     "env_set": {
-	    "ALLURE_HOME": "$dir"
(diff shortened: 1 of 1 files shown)


r15ch13:
What about adding [persistence](https://github.com/lukesampson/scoop/wiki/Persistent-data) for `config` and `plugins`? Is `ALLURE_HOME` necessary? Because `allure.bat` sets it to the current directory.

ehborisov:
@r15ch13 ok, I added persistence, but I didn't get your point about `ALLURE_HOME`, on my machine it points to the `current` folder where active Allure installation lies. This variable can be used from other tools to discover and configure Allure installation, so everything seems right to me.

PR #1494, opened 2017-05-16
Version substitution for license property

se35710:
To allow something like this:
```
"license": "https://<some-site>/1.0.0/license/en.html",
"autoupdate": {
     "url": "https://<some-site>/$version/<software>-$version.zip",
     "license": "https://<some-site>/$version/license/en.html"
}
```


lukesampson:
Thank you.

PR #1501, opened 2017-05-19
PowerShell script to locate Java on Windows

se35710:
Optionally sets JAVA_HOME and JRE_HOME

### DIFF
--- a/bucket/find-java.json
+++ b/bucket/find-java.json
@@ -1,9 +1,9 @@
 {
-    "version": "7",
+    "version": "9",
     "homepage": "https://gist.github.com/se35710/43693e679701387d722206eff1e85f5f",
     "description": "PowerShell script to locate Java on Windows, optionally sets JAVA_HOME and JRE_HOME",
-    "url": "https://gist.githubusercontent.com/se35710/43693e679701387d722206eff1e85f5f/raw/57182a52ab44d78bd4f8a865b3aa23ff51732e25/find-java.ps1",
-    "hash": "99134c4e8bf5b40acb3d1d1c272f62dfadce0e2791764ba89c68d4b1ef34bff1",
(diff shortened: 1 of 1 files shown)


lukesampson:
Thanks. I think this would be better in the [extras](https://github.com/lukesampson/scoop-extras) bucket — do you mind resubmitting there please?

tresf:
If not already, this script should also detect the 32-bit registry version, such as when running 32-bit Java on 64-bit Windows (Windows-on-Windows64, aka WOW64)

se35710:
@lukesampson Sure, but I do think it is a good idea to include it in the main bucket. Some installers (Flexeras Install Anywhere for example) depends on `JAVA_HOME` being set, even if java.exe is available in `PATH`. Instead of using `”depends”: “OpenJDK”` and download 100 MB, this 11 kB script will find a suitable Java with `"depends": "find-java"`. Java (when installed outside of scoop) does not set `JAVA_HOME`, and its install directories is also version dependant. For that reason I think this helper script could be helpful.

For manifests, like the Tomcat one, the catalina shim could be updated to find a suitable Server JRE, for all the Eclipse manifests a fullblown JDK, etc.

Another more complicated way would to extend the same functionality within scoop itself, and if that is something that is preferred, I offer my help.

@tresf I’ve updated the script now so it searches through the WoW registry node too. `-Achitecture $arch` could be used in scoop manifests to force `$arch` for 32 bit Java programs running on 64 bit Windows.

lukesampson:
I thought we had removed all the instances of `"depends": "openjdk"` — maybe we missed some.

The new approach is to use `"suggest"` instead of `"depends"` for Java SDKs. It leaves the choice up to the developer, leaves open the possibility of a non-Scoop installed JDK, and it's not hard to run an extra `scoop install oraclejdk`.

PR #1505, opened 2017-05-23
use diffutils from msys2

wsw0108:
because diff utils from msys crash under win10


lukesampson:
Nice one — thank you 👍

PR #1510, opened 2017-05-23
Update composer.json

nueko:
Set Composer home and register composer global bin

### DIFF
--- a/bucket/composer.json
+++ b/bucket/composer.json
@@ -13,6 +13,14 @@
         echo 'if($args -eq \"selfupdate\" -or $args -eq \"self-update\") { & scoop update composer }' | out-file $dir\\composer.ps1
         echo 'else { & php (join-path \"$psscriptroot\" \"composer.phar\") @args }' | out-file $dir\\composer.ps1 -append
     ",
+    "post_install": "
+        $och = \"$Env:APPDATA\\Composer\"
+        if(Test-Path $och) {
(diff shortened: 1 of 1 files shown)


r15ch13:
Composer uses `%APPDATA%\Composer` to store it's data. What's the use case for setting `COMPOSER_HOME`?

nueko:
multi user purpose when scoop directory not inside the `%USERPORFILE%`

r15ch13:
Also requires less space, because all use the same `vendor` directory. I see.
But `persist` can't be empty. So let's add a `home` directory, make it persistent and point the `COMPOSER_HOME` to it. The `bin` path points to the linked `~\scoop\composer\current\home` directory.
Should look like this:
```
    "persist": [
        "home"
    ],
    "env_set": {
        "COMPOSER_HOME": "$persist_dir\\home"
    },
    "env_add_path": "home\\vendor\\bin",
```

nueko:
also, should we move %APPDATA%\Composer to the new one?

r15ch13:
Yeah, I agree. At least until the next composer update.

nueko:
I think, we should just throw warning about `COMPOSER_HOME` variable creation and recommend user to move it to the new location.
Or it is better to write an `postinstall` script?

r15ch13:
A `postinstall` script and a warning only when something was moved should be okay.

PR #1513, opened 2017-05-25
Add unbound

nueko:


lukesampson:
`post_install` is not an appropriate place for starting services.

r15ch13:
https is available for all URLs

nueko:
@lukesampson the `post_install` just replace a string on an extracted file.
@r15ch13 will update it.

lukesampson:
Sorry, I meant `pre_install` — what's that doing?

nueko:
it is just proxing `unbound` command without parameter to provide right default configuration, 
user should run `unbound -c ~/scoop/apps/unbound/current/service.conf` to run unbound
but with the `pre_install` i created, just  `unbound` is enough. also it runs in bg as it should.
is that too much?

lukesampson:
Thanks for explaining. I'm a bit slow — I thought it was running that command rather than writing it to a file. It is a more code than is ideal, but it's okay in the absence of a `scoop service` command.

Issue #1514, opened 2017-05-25
Installing apps globally creates shortcuts in AppData instead of ProgramData


se35710:
`scoop install -g <some app with shortcut in manifest>` puts the shortcut in user's AppData instead of ProgramData for all users.

se35710:
Fixed in PR #1516

PR #1516, opened 2017-05-26
Fix global shortcuts (issue #1514)

se35710:


lukesampson:
Looks great! Thanks for that.

PR #1531, opened 2017-05-31
persist npm cache

nueko:
As of npm@5 supports offline install it worth to persist the cache


r15ch13:
Nice, remind me to add this to the versions bucket too.

PR #1537, opened 2017-06-01
Persist cache

nueko:


r15ch13:
This is more clear: `"post_install": "yarn config set cache \"$persist_dir\\cache\"",`

The note will not work if `npm` isn't installed. Also `yarn` adds global binaries to `%APPDATA%\Local\Yarn\bin`.

nueko:
> %LOCALAPPDATA%\Yarn\bin to the PATH might be a better idea. 

Agree!

> The note will not work if npm isn't installed. Also yarn adds global binaries to %LOCALAPPDATA%\Yarn\bin.

That's why it just a note, yarn global bin will set to the npm global bin if we set the prefix to yarn config

r15ch13:
If we add the global bin to PATH, then we don't need the note.

nueko:
Can You tweak it, mate? I have to _sahur_ right now.

r15ch13:
Yep

r15ch13:
In which way would this be better? I think it is more confusing.

nueko:
Ah, nevermind.
That's user choice.
I think that's enough

PR #1539, opened 2017-06-01
add hashcat and just hashcat

7anner:


r15ch13:
Add `"extract_dir": "hashcat-3.5.0"` and `"extract_dir": "hashcat-$version"` to hashcat. Then remove `hashcat-3.5.0\` from every `bin` option. Because `bin` can't use `$version` substitions. Easy way to test if everything works. Change the `version` to `3.4.0` and run `.\bin\checkver.ps1 hashcat -u` in your repo.

r15ch13:
protobuf is already available in the [extras bucket](https://github.com/lukesampson/scoop-extras/blob/master/protobuf.json)

7anner:
okay, shoot.  What do I need to do to appease appveyor?

r15ch13:
You can remove the `architecture` completely from `autoupdate` :smile:

PR #1541, opened 2017-06-03
Enable commands in the bucket

nueko:
this PR enable bucket commands in the /commands directory
usage:
`scoop php/info $args`

### DIFF
--- a/lib/commands.ps1
+++ b/lib/commands.ps1
@@ -1,7 +1,20 @@
+. "$psscriptroot\buckets.ps1"
+
+if ($cmd.Contains('/')) {
+    $bucket, $cmd = $cmd -split ('/');
+    $bucketcmd = "$bucketsdir\$bucket\commands"
+    $bucketcmd_exists = Test-Path $bucketcmd
+}
+
 function command_files {
-    (gci (relpath '..\libexec')) `
-        + (gci "$scoopdir\shims") `
-        | where { $_.name -match 'scoop-.*?\.ps1$' }
+    if ($bucket -and $bucketcmd_exists) {
+        $files = gci ($bucketcmd)
+    }
+    else {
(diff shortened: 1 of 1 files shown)


nueko:
`scoop help php/info` also works like what expected.
>what this would be used for?

let's say it is about php
bucket creator can add commands to start the app, de/activate functionalities, etc...
eg: 
`scoop php/start [7.1]` to run `php-cgi.exe -b 127.0.0.1:9000 -c php.ini
`scoop php/start nginx`
`scoop php/dismod redis [7.1]`
`scoop php/enmod redis [7.1]`

lukesampson:
Okay thanks. I think we can use this, but I'm not sure how I feel about using `scoop` as the entry point for these bucket commands—it still feels like there should be more separation between what's Scoop and what could be completely separate functionality. Maybe it's just a matter of doing it as `scoop run [bucket]/[cmd]`?

What do you think, @r15ch13? How does it fit in with what you were planning? Is there overlap with a `scoop services` command?

nueko:
it is not about starting services.
see the last snippet, it enables/disables php module. another example, maybe to create adservers or create root.key

lukesampson:
Okay, but `php-cgi.exe -b 127.0.0.1:9000 -c php.ini` is very similar to starting a service.

nueko:
it's just example :smile

lukesampson:
Understood. I think it's a good idea and I'm sure we can get something working but we might need to tweak it a bit. We need to move slowly when making changes to core functionality like this. And it'd be good to coordinate with @r15ch13 because he seems to be working on similar areas.

rasenplanscher:
> Maybe it's just a matter of doing it as `scoop run [bucket]/[cmd]`?

I think, you're spot-on here.
That would be a lot cleaner, and going without `run` would not help all that much in reducing typing...

rashil2000:
Closing as stale

PR #1542, opened 2017-06-04
Improve feedback when Scoop is outdated and updated

TeaDrivenDev:
When there is an update to Scoop and `scoop status` is run, the output is
```
Scoop is out of date. Run 'scoop update' to get the latest changes.
Everything is ok!
```
This is not ideal, because "Everything is ok!" is what it says when it has checked for package updates and there are none - but in this case it apparently doesn't even check.

The next step if I want to do anything is `scoop update`:
```
Updating Scoop...
```
Now to find out if there are updates to the installed packages, as was the original intention, `scoop status` needs to be run a second time, but I might not think of that, because it said "Everything is ok" earlier.

This PR improves the feedback in two places:
- Don't say "Everything is ok!" when Scoop is out of date; instead warn with "Cannot check for app updates while Scoop is out of date."
- Additionally say "Run 'scoop status' to check for updates to installed apps." after updating Scoop as an additional reminder to the user.

### DIFF
--- a/libexec/scoop-status.ps1
+++ b/libexec/scoop-status.ps1
@@ -102,8 +102,4 @@ if(!$old -and !$removed -and !$failed -and !$missing_deps -and !$needs_update) {
     success "Everything is ok!"
 }
 
-if ($needs_update) {
-    warn "Cannot check for app updates while Scoop is out of date."
-}
-
 exit 0
--- a/libexec/scoop-status.ps1
+++ b/libexec/scoop-status.ps1
@@ -98,8 +98,12 @@ if($missing_deps) {
     }
 }
 
-if(!$old -and !$removed -and !$failed -and !$missing_deps) {
+if(!$old -and !$removed -and !$failed -and !$missing_deps -and !$needs_update) {
     success "Everything is ok!"
+}
+
+if ($needs_update) {
+    warn "Cannot check for app updates while Scoop is out of date."
 }
 
 exit 0
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -89,6 +89,7 @@ function update_scoop() {
         popd
     }
     success 'Scoop was updated successfully!'
+    "Run 'scoop status' to check for updates to installed apps."
 }
 
 function update($app, $global, $quiet = $false, $independent, $suggested) {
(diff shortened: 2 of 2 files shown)


libexec/scoop-status.ps1
```diff
@@ -98,8 +98,12 @@ if($missing_deps) {
     }
 }
 
-if(!$old -and !$removed -and !$failed -and !$missing_deps) {
+if(!$old -and !$removed -and !$failed -and !$missing_deps -and !$needs_update) {
     success "Everything is ok!"
 }
 
+if ($needs_update) {
+    warn "Cannot check for app updates while Scoop is out of date."
```
lukesampson: This is not true.

libexec/scoop-update.ps1
```diff
@@ -89,6 +89,7 @@ function update_scoop() {
         popd
     }
     success 'Scoop was updated successfully!'
+    "Run 'scoop status' to check for updates to installed apps."
```
lukesampson: Don't need this level of instruction.

libexec/scoop-status.ps1
```diff
@@ -98,8 +98,12 @@ if($missing_deps) {
     }
 }
 
-if(!$old -and !$removed -and !$failed -and !$missing_deps) {
+if(!$old -and !$removed -and !$failed -and !$missing_deps -and !$needs_update) {
```
lukesampson: This is the only change required.

libexec/scoop-status.ps1
TeaDrivenDev: It's what I'm seeing, though. Whenever there is a Scoop update *and* an update to an installed package, it will only show me the latter when I run `scoop status` again after first updating Scoop itself. That is the confusing thing I'm trying to get fixed.

lukesampson: If Scoop is out of date and there are app updates too, these changes will make Scoop print that it can't check for app updates right after it has printed out what the app updates are.

If you were meaning to prevent Scoop checking for app updates if it itself is out of date — that's a judgement call that I don't agree with. Scoop is updated many times a day and is almost guaranteed to be out of date. It's still useful to print the apps it knows to be out of date.

You're right about the 'Everything is ok!' message being confusing.

r15ch13:
@lukesampson if you use `Squash and merge` for old or long (many commits) pull requests, then the git log will look a bit cleaner. 😁  

pr-merge

lukesampson:
Thanks @r15ch13, I'll try to do that.

PR #1436, opened 2017-04-16
maven:3.5.0

onacit:
Apache Maven 3.5.0

Issue #894, opened 2016-06-07
Git-extras


jahead:
Hey luke, 

How about git-extras?
https://github.com/tj/git-extras

it has a install.cmd?

lukesampson:
Yes, please submit a pull request if you can get it working. Thanks.

JanPokorny:
I was thinking about ways to implement this. Scoop does not seem well-suited for an app that basically installs over another app.

I can think of two possibilities how to implement this:

1. Create `git-with-extras` manifest, which is the `git` manifest with `git-extras` installation step added. Would install alongside normal `git` and override its shim. Seems the most straightforward, but comes at the cost of having two full copies of gits installed, one clean and one with git-extras.
1. Create a `git-extras` manifest that:
  - Depends on `git`
  - On installation, patches the current installed version of `git`
  - Overrides the `git` shim with one that first checks if the current version of `git` is patched, and if not, patches it. It then forwards everything to real `git`.
  This has the obvious disadvantage of patching `git` lazily in case `git` is newer than `git-extras`. Also seems rather complicated. But only one version of git is installed.

What do you think is better?

pxgamer:
@JanPokorny, I ended up going with option 2 [in my bucket](https://github.com/pxgamer/scoop-bucket/blob/master/bucket/git-extras.json) while testing. Not sure what the best way to do the Git shim patching though. 🤔

Overall option 1 seems easier to implement, but I think option 2 is better.

JanPokorny:
@pxgamer Your solution also has a problem with updating -- updating `git` removes the extras. (That's why I mentioned the need for lazy patching.)

Looking at it now, keeping two copies of git does not really seem like a big deal, compared to patching it on demand.  Scoop actually has direct support for two apps providing the same shims (that's what `scoop reset` is for). Thus, I believe that creating `git-with-extras` is the way to go.

pxgamer:
Yeah, I think that makes the most sense. Would be good to see a PR if you want to do it. 👌

JanPokorny:
@pxgamer I plan on creating the package in the near future (it's the exam period now...).

As Scoop itself depends on `git`, you would need to have it installed alongside `git-with-extras`. As `git-with-extras` is a superset of `git`, it should not break anything even if some tool uses `git-with-extras` while expecting vanilla git.

JanPokorny:
[...]

PR #1544, opened 2017-06-05
Create shims for bash scripts

r15ch13:
Potential fix to help Cygwin/MinGW users.
Related issues: #1436, #894, https://github.com/lukesampson/scoop-extras/issues/434

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -215,6 +215,9 @@ set args=%args:)=``)%
 set invalid=`"='
 if !args! == !invalid! ( set args= )
 powershell -noprofile -ex unrestricted `"& '$(resolve-path $path)' %args%;exit `$lastexitcode`"" | out-file $shim_cmd -encoding ascii
+    } else {
+        $shim_cmd = $(strip_ext($shim))
+        "#!/bin/sh`n`"$(resolve-path $path)`" `"$@`"" | out-file $shim_cmd -encoding ascii
     }
 }
 


lukesampson:
I'm not sure if I'm following the code logic — does this only create a bash shim if the target to be shimmed doesn't have an extension of `exe`, `bat`, `cmd`, or `ps1`?

So if a manifest has `"bin": "code.cmd"`, a bash shim wouldn't be created, but if it has `"bin": "code.xyz"` it would?

r15ch13:
Basically yes. All bash script have no extension, but I can add a check for the [Shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) so it will not create useless shims.

lukesampson:
Okay sounds good. I was thinking it would try to shim every `bin` for bash, but I guess that's a lot harder to do.

lukesampson:
I think you also need to avoid writing the .ps1 shim ([lines 181 - 190](https://github.com/lukesampson/scoop/blob/02730deed3ccf986de6ca51dbd100ecff88fd88f/lib/core.ps1#L181-L190)) if you're shimming a bash script, otherwise it will create a .ps1 shim that won't execute, and it would overwrite the correct .ps1 shim in the case of `"bin": [ "bin/code.cmd", "bin/code" ].

nrakochy:
If I understand this correctly, this pr would allow the executables to be picked up by `Cygwin`? Would be a great help for me.

Ash258:
@r15ch13 #2741?

rasa:
This issue appears obsolete. If not, please reopen.

Issue #1546, opened 2017-06-08
Update nvm to 1.1.5 breaks nvm


u9520107:
It seems the commit that added the persistence broke the installation. I used the 1.1.5 manifest prior to the persistence commit, and the installation works fine.

I'll try to test some more later.

pepetopo:
Same here. Upgraded to NMV 1.1.5 and tried upgrading to Node 8.0.0. The installation goes smoothly, but upon trying to actually enable the new version, I get this error message:
```
C:\Git\vagrant-ansible-lemp\sites>nvm use 8.0.0
exec: "C:\\Users\\elias\\scoop\\persist\\nvm\\nodejs\\elevate.cmd": file does not exist:
```
Downgrading to older version of Node won't work after that either, and reinstallation does'nt solve the issue

r15ch13:
Should be solved with #1547

PR #1547, opened 2017-06-08
modify the bucket manifest to allow persist

u9520107:
#1546 

I've tested with this update manifest. 
1. The downloaded node versions will be persisted properly. 
2. The nvm created symlink lives in apps folder and is not persisted. 
3. This symlink seems to prevent scoop uninstall to delete the nvm folder in apps. It'll always say that the folder is in use. But after manually remove the symlink, uninstall will work fine.
4. I tried to put the symlink in the persist folder, but I am not sure how to properly add that symlink to path. ```"env_add_path": "$persist_dir\\nodejs"``` doesn't seem to work.

Please advise if more change is needed.

### DIFF
--- a/bucket/nvm.json
+++ b/bucket/nvm.json
@@ -8,24 +8,25 @@
         "elevate.vbs"
     ],
     "persist": [
-        "nodejs",
-        "settings.txt"
+        "settings.txt",
+        "elevate.cmd",
+        "elevate.vbs"
     ],
     "env_add_path": "nodejs",
     "env_set": {
-        "NVM_HOME": "$dir",
-        "NVM_SYMLINK": "$persist_dir\\nodejs"
+        "NVM_HOME": "$persist_dir",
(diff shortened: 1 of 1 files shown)


u9520107:
@r15ch13 Yes, that would allow the nvm use to run ok. But trying to run node would fail.
the nodejs symlink in the apps folder points to the nodejs folder, and the node installations are subdirectories in the nodejs folder. So, node.exe from v8.0.0 would be in $persist_dir\\nodejs\\v8.0.0, and the system still fails to find node.exe.

u9520107:
Ok, taking @r15ch13 's idea, and modifies it to make nvm create symlinks in the nodejs folder instead. Also changed the path to nodejs\\nodejs. Now everything seems to work as intended with properly nvm symlink persistence.

r15ch13:
That works! 👍

PR #1548, opened 2017-06-08
Create python2.json

prasannavl:
Python2 is still unfortunately, a dependency for a number of projects including some mainstream ones like node-gyp, etc. So, adding the manifest for it.

This will however, create conflicting python.exe executable, depending on which one is installed first. Not sure how to solve this. But I presume there is a clean way to do this, since jdk seems to do it. However, I'm creating a PR for this, since a working resolvable (since they are just the shims) conflict is better than nothing.

### DIFF
--- a/bucket/python2.json
+++ b/bucket/python2.json
@@ -1,0 +1,37 @@
+{
+    "homepage": "https://www.python.org/",
+    "license": "https://docs.python.org/2/license.html",
+    "version": "2.7.13",
+    "architecture": {
+        "64bit": {
+            "url": "https://www.python.org/ftp/python/2.7.13/python-2.7.13.amd64.msi#/py2.msi",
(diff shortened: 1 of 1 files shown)


r15ch13:
Python2 is already available in the versions bucket. https://github.com/scoopinstaller/versions/blob/master/python27.json

Run `scoop bucket add versions` to add the bucket.

prasannavl:
Ah. Great. Thanks for the info. 

PS: Consider adding that to the documentation? Would be helpful.

r15ch13:
It's in wiki under buckets. Where would you suggest adding it? Main readme.md maybe?

prasannavl:
I saw the `extras` bucket. But missed the versions one. My bad. But I guess
a little bit of highlight would be helpful. Wouldn't you think so? Perhaps,
a small section for all the included buckets, may be?

Issue #1526, opened 2017-05-30
scoop services install <app>


r15ch13:
Discussion about implementing service management similar to [`brew services`](https://github.com/Homebrew/homebrew-services).

Previous discussions about this topic: #1512, #513

I stumbled over [Laravel Valet for Windows](https://github.com/cretueusebiu/valet-windows) which uses [WinSW](https://github.com/cretueusebiu/valet-windows/blob/master/cli/Valet/WinSW.php) to setup a PHP developing environment. I know this is not the purpose of `scoop`, but I would like to create a documentation for setting up an updateable PHP dev environment. These zero-configuration setups (like xampp or valet) are cool, but most of them don't allow changing components or updates at all. With `scoop` it would be easy to swap out nginx for caddy/apache, MySQL for MariaDB and so on. Loosely coupled!

We could utilize https://github.com/kohsuke/winsw for installing/starting/stopping services.
In each manifest, we could provide a basic configuration that is copied to `~/scoop/persist/<app>/service.xml` so it can be easily modified.

E.g. `scoop services install nginx` would copy `winsw.exe` to `~/scoop/services/nginx.exe` (winsw has to be renamed) and register it, using a configuration located in `~/scoop/persist/nginx/service.xml`.
All other commands ` run|start|stop|restart` can be implemented like `brew services` commands.

Example installation for a full PHP environment:
```
λ scoop install php
λ scoop install composer
λ scoop install nginx
λ scoop install mariadb
λ scoop services install php-fpm
λ scoop services start php-fpm
λ scoop services install nginx
λ scoop services start nginx
```

Programs which could be run as a service:
- [apache](https://github.com/lukesampson/scoop/blob/master/bucket/apache.json)
- [caddy](https://github.com/lukesampson/scoop/blob/master/bucket/caddy.json)
- [minio](https://github.com/lukesampson/scoop/blob/master/bucket/minio.json)
- [php](https://github.com/lukesampson/scoop/blob/master/bucket/php.json)
- [rethinkdb](https://github.com/lukesampson/scoop/blob/master/bucket/rethinkdb.json)
- [unbound](https://github.com/lukesampson/scoop/blob/master/bucket/unbound.json)
- [mongodb](https://github.com/lukesampson/scoop/blob/master/bucket/mongodb.json)

What to you guys think?

lukesampson:
WinSW sounds like it would be useful for programs that don't know how to run as a Windows service. Is that necessary though?

[...]

PR #1549, opened 2017-06-10
[WIP] scoop services (first draft)

r15ch13:
First draft PR for #1526

Services XML files will be located in `./bucket/services` for the main bucket and `~/scoop/buckets/<bucketname>/services` for other buckets.

## Steps for testing
Checkout this PR and install `sudo`, `winsw` and `php`.

> scoop install sudo winsw
> scoop install php

> scoop services install php

Installing a service will look for a suitable XML file in the `services` directory.
In this case, the `service` property of `php` is set to `php-fpm`.
It will copy `<bucket>/services/php-fpm.xml` to `~/scoop/persist/php/service/php-service.xml` and create a copy of `WinSW.exe` in at the same location (`php-service.exe`)

Start/stop/restart the php services with:
> scoop services start php
> scoop services stop php
> scoop services restart php

Uninstall the php services with (`php-service.xml` and `php-service.exe will remain untouched)
> scoop services uninstall php

Lists all services that are available and their status (only show installed apps)
> scoop services list

## Todo list:
- [ ] show information after installing an app with a service
- [ ] uninstall service before uninstalling an app
- [ ] implement/test `--all` option
- [ ] implement/test `--global` option
- [ ] things I overlooked or didn't thought about jet ...

## Services todo list:
- [ ] apache
- [x] caddy
- [ ] minio
- [ ] mongodb
- [ ] nginx
- [x] php
- [x] php-nts
- [ ] rethinkdb

### DIFF
--- a/bucket/caddy.json
+++ b/bucket/caddy.json
@@ -12,7 +12,9 @@
             "hash": "51d2de0ebebfa3519062772f144632de4b6d0348fcb3a394ed5a98e04ab24031"
         }
     },
+    "service": "caddy",
     "bin": "caddy.exe",
+    "post_install": "if(!(test-path \"$persist_dir\\Caddyfile\")) { \"localhost:2015\" | Out-File -encoding \"ASCII\" \"$persist_dir\\Caddyfile\" }",
     "checkver": {
         "github": "https://github.com/mholt/caddy"
     },
--- a/bucket/php-nts.json
+++ b/bucket/php-nts.json
@@ -12,6 +12,7 @@
             "hash": "sha1:79ddfd52e23eaafa772b9c1b6266eff82aac4d55"
         }
     },
+    "service": "php-fpm",
     "bin": [
         "php.exe",
         "php-cgi.exe"
--- a/bucket/php.json
+++ b/bucket/php.json
@@ -12,6 +12,7 @@
             "hash": "sha1:5408393a7527463701aa2576faab68396c4aab92"
         }
     },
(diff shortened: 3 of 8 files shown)


r15ch13:
@deevus thought about doing that at first. If you take a look at the [sample-allOptions.xml](https://github.com/kohsuke/winsw/releases/download/winsw-v2.1.0/sample-allOptions.xml) you'll see that WinSW has many options and I believe it's unnecessary to reimplement them just for a service running with the least amount of configuration. If need the user can change all options afterward.

lukesampson:
Sorry, I just got back from holiday. This is looking promising.

I share @deevus's sentiment that it would be very nice to do it in JSON and generate XML.

Nice because Scoop could act as a simpler wrapper around WinSW, making it easier for manifest authors to do what's needed most of the time, like executable, arguments, and environment. It would also be nice to generate defaults for the service id, name, description, logging and restart without having to put them in the manifest.

I get your points about WinSW supporting lots of options, and allowing users to change options later. Wouldn't it be possible to keep both these benefits if we generate the XML?

jvican:
This is a nice improvement I find myself needing in scoop to ship many of the developer tools I work on. Is there someone working actively on it? :smile:

pedzed:
I just `scoop install mysql`'ed, but I'm not sure how to do service management. Apparently it's not yet possible with Scoop?

gpetrov:
Hope this gets implemented soon - such a useful feature

PR #1561, opened 2017-06-18
HTTPS everywhere

rasa:
There are all the http and ftp URLs that have working https equivalents.


rasa:
I used https://github.com/rasa/scoops/blob/master/check_https.py to update the .jsons.

r15ch13:
Cool script! 👍

PR #1563, opened 2017-06-18
Fix various hashes + HTTPS everywhere

rasa:


rasa:
Used new version of https://github.com/rasa/scoops/blob/master/check_https.py

rasa:
Sorry, but this should be reverted as many https urls have invalid certs, such as https://strawberryperl.com.

PR #1564, opened 2017-06-19
Fix bad https URLs, etc.

rasa:


r15ch13:
These three sites have valid certificates, why did you change them back? :smile:
https://www.erlang.org
https://geteventstore.com/
https://www.radare.org/r/

rasa:
Gosh dang it, sorry. My script reported them as invalid. I'm using Python's urllib3, and it's been a real PITA to get this thing right. I will fix and resubmit.

r15ch13:
Check if Tls, Tls1.1, Tls1.2 are supported :)

PR #1566, opened 2017-06-20
add msys2

AStupidBear:


lukesampson:
Thanks for that. I fixed the couple of failing tests and added 32-bit and autoupdate support. I had to remove the dots from the version (`20161025`) to support autoupdate.

PR #1572, opened 2017-06-29
Add Helm

neoeinstein:
Adds the Helm tool.


r15ch13:
Thanks! 👍
There is a typo in the checkver URL and you can ommit the `releases/latest` part.

PR #1578, opened 2017-07-03
Remove extract_dir from git-lfs

rasa:
extract_dir was added 4 months ago in https://github.com/lukesampson/scoop/commit/48441508dd81412a35a88a32119b90ea29809fb0


lukesampson:
Thanks

PR #1580, opened 2017-07-07
Add LilyPond

mintchkin:


r15ch13:
It might be better suited for [scoop-extras](https://github.com/lukesampson/scoop-extras)

PR #1584, opened 2017-07-08
Add cygwin at 2.8.1

rasa:


rasa:
I tried persisting $dir via:
````
    "persist": [
        "."
    ],
````
but that gave me a folder named `persist/cygwin/bucket` and nothing worked.

PR #1586, opened 2017-07-09
Revert cygwin to version 2.8.1

rasa:
And removed the autoupdate section which is not useful.

### DIFF
--- a/bucket/cygwin.json
+++ b/bucket/cygwin.json
@@ -1,5 +1,5 @@
 {
-    "version": ".",
+    "version": "2.8.1",
     "homepage": "https://cygwin.com/",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


r15ch13:
How about using `nightly` as the version? The setup changes from time to time.

rasa:
You're right. 2.8.1 is the version of Cygwin.dll, not setup*.exe.

rasa:
The version number for setup*.exe is 2.881 per https://www.cygwin.com/ml/cygwin-announce/2017-07/msg00017.html

rrelmy:
The setup.exe does check if there is an update available, so there is some sort of check for the latest version.

r15ch13:
Can you find the URL via wireshark or the sourcecode? :)

rrelmy:
I did not check the source but it most likely uses the `setup-version` property from the `setup.ini` which is present on each mirror
```
# This file was automatically generated at 2017-07-11 02:37:28 UTC.
#
# If you edit it, your edits will be discarded next time the file is
# generated.  See http://cygwin.com/setup.html for details.
release: cygwin
arch: x86_64
setup-timestamp: 1499740648
setup-version: 2.881
```

I took this from https://linux.rz.ruhr-uni-bochum.de/download/cygwin/x86_64/setup.ini

I do not know how fast the mirrors are updated and if there could be a difference between this file and the setup from the original site

r15ch13:
Or we use the Taglist from the repository: https://cygwin.com/git/gitweb.cgi?p=cygwin-apps/setup.git;a=tags (regex: `release_([\d.]+)`)
The problem stays the same. When the setup.exe is not updated in time, the update thingy will use the wrong hash.

rasa:
The canonical location for setup.ini is ftp://ftp.cygwin.com/pub/cygwin/x86/setup.ini

r15ch13:
This works too https://cygwin.com/pub/cygwin/x86/setup.ini

PR #1594, opened 2017-07-12
Re-add extract_dir to git-lfs (getting old)

rasa:


r15ch13:
Wow, that's annoying, why do they keep changing every now and then? ...

rasa:
No idea, it may have been a one-time glitch in their build script (which I couldn't find).

PR #1595, opened 2017-07-12
Fix autoupdate URLs in git-lfs

rasa:
Sorry, I didn't see they moved it from https://github.com/github/git-lfs to https://github.com/git-lfs/git-lfs.


r15ch13:
No problem. Github redirects to the new repo name anyway 👍

PR #1600, opened 2017-07-15
Add vcredist2015 to rg (fixes #1599)

### LINKED ISSUE #1599: ripgrep (rg) requires vcruntime140 (Visual C++ Redistributable)
`rg` doesn't work after installing it, because of a missing dependency "vcruntime140.dll".

rasa:


rasa:
Shouldn't use github's edit function. Too much trouble.

PR #1604, opened 2017-07-18
Update gcc version from version 6.3 to 7.1, gradle from 4.0 to 4.0.1

liaoya:
The gcc in main bucket is quite old, I update it. I always use latest gradle and that in main bucket is a little old.


lukesampson:
Thanks for these 👍

PR #1610, opened 2017-07-19
Add .NET SDK

inosik:
The version number in the file name I mentioned in [#1396](https://github.com/lukesampson/scoop/pull/1396#issue-216351620) isn't correct anymore, they changed it from 1.1.1 to 1.0.1 at some point. So we can't update the old manifest to 1.0.4, because that would be lower than the old version.

I'm not sure about the deprecation note, I think we should rather remove the `dotnet` manifest for now. What do you think?


inosik:
I've just updated the .NET SDK to 2.0.0.

PR #1619, opened 2017-07-21
Update sed to 4.4-2

rasa:


rasa:
Move to msys2's build
Add autoupdate & checkver
Add 64bit build

PR #1620, opened 2017-07-21
Update which to 2.21-2

rasa:


rasa:
Move to msys2's build
Add autoupdate & checkver
Add 64bit build

rasa:
This method is too complicated.

r15ch13:
Too many deps?

rasa:
Yeah, here's what I've found:

file:
````
msys-magic-1.dll (in file)
msys-gcc_s-1.dll (provided by gcc-libs-6.3.0-1)
msys-2.0.dll (provided by msys2-runtime-2.8.2-1)
````
grep:
````
msys-gcc_s-1.dll
msys-iconv-2.dll (provided by libiconv-1.14-2)
msys-intl-8.dll (provided by libintl-0.19.8.1-1)
msys-2.0.dll
msys-pcre-1.dll (provided by libpcre-8.40-2)
````
sed:
````
msys-gcc_s-1.dll
msys-intl-8.dll
msys-2.0.dll
````
which:
````
msys-2.0.dll
msys-gcc_s-1.dll
````

r15ch13:
Using msys as a whole package, would be a better solution then extracting every dependency. What do you think about we remove these smaller versions and hint to the use of [msys2.json](https://github.com/lukesampson/scoop/blob/master/bucket/msys2.json)?

rasa:
Agreed. Or host statically linked versions, per https://sourceforge.net/p/msys2/mailman/msys2-users/thread/CACac1F8WLEtNCkuCipFU%3DjNW6ZQn%2Bj-VX0y0pyGZr%3D1gUopB%3DA%40mail.gmail.com/#msg33756517

PR #1630, opened 2017-07-26
Add reverse property to checkver

r15ch13:
Add a reverse property to checkver.
If set to true, the regex matches get reversed.
This helps to find new versions in directory listings. (see x264)

Like: https://download.videolan.org/pub/x264/binaries/win64/

I know there are other programs with the same problem, but I can't remember them right now :grin:

**Needs testing on Unix!**

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -87,12 +87,15 @@ $queue | % {
         $regex = $json.checkver
     }
 
+    $reverse = $json.checkver.reverse -and $json.checkver.reverse -eq "true"
+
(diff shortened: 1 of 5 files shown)


matthewjberger:
What is the status of this? I have some manifests I'd like to do where this would be useful and I'm looking forward to this pull request being merged. Excellent idea!

r15ch13:
Tested on unix

r15ch13:
@matthewjberger

matthewjberger:
Thank you!

PR #1632, opened 2017-07-29
Fix checkver.re for sbt

asmgf:


r15ch13:
They change the URL and don't host it on bintray as it seems.

PR #1637, opened 2017-08-04
Added putty

camden7306:
putty is useful


r15ch13:
putty is already available in the extras bucket.
https://github.com/lukesampson/scoop-extras/blob/master/putty.json

You can add it like so: `scoop bucket add extras`

PR #1643, opened 2017-08-08
Add artifact v0.9.0.

matthewjberger:
[Artifact](https://github.com/vitiral/artifact) is a design document tool for developers. I figured it would fit the main bucket well because it's a standalone tool with a single executable. I can submit the manifest on the extras repo instead if it doesn't belong here though!


r15ch13:
I don't understand what it does, but sure why not :grin:

PR #1648, opened 2017-08-14
add cleanup command

jfmherokiller:
this pull request adds a cleanup command to scoop which will remove older versions of apps while preserving persistent data.

### DIFF
--- a/libexec/scoop-cleanup.ps1
+++ b/libexec/scoop-cleanup.ps1
@@ -1,0 +1,80 @@
+# Usage: scoop cleanup <app> [options]
+# Summary: Update apps, or Scoop itself
+# Help: 'scoop cleanup' cleans Scoop apps by removing old versions.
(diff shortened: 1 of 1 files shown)


libexec/scoop-cleanup.ps1
```diff
@@ -0,0 +1,80 @@
+# Usage: scoop cleanup <app> [options]
+# Summary: Update apps, or Scoop itself
```
h404bi: Should be `"Clean all or specific Scoop apps"`, instead of `"Update apps, or Scoop itself"`?

r15ch13:
@jfmherokiller  A bit late, but what does `fsutil.exe reparsepoint delete $file.FullName` do?
If I test this on file/directories with the `LinkType: HardLink` I get an Error: `The file or directory is not a reparse point.`

PR #1652, opened 2017-08-16
Allure: remove persist section

baev:
This pull request fixes https://github.com/allure-framework/allure2/issues/561


r15ch13:
In which way does it fix this issue? Do the configuration or plugins get outdated?

baev:
this folders are contains default configs and plugins, and should be updated for each release

r15ch13:
Okay

PR #1654, opened 2017-08-18
Suggest Git instead of depending on it

inosik:
That way, people using `git-with-openssh` won't end up with two versions of Git installed.


r15ch13:
Good idea 👍

PR #1658, opened 2017-08-22
Update rustup manifest to use new un/installer script block syntax.

matthewjberger:
This pull request updates the rustup manifest to use script blocks in the install and uninstall sections instead of creating them in the 'pre_install' section. Thanks to @r15ch13's idea [from this pull request](https://github.com/lukesampson/scoop/pull/1657).

### DIFF
--- a/bucket/rustup.json
+++ b/bucket/rustup.json
@@ -13,14 +13,6 @@
         ".cargo",
         ".rustup"
     ],
-    "pre_install": "
-        # Create the installation script
-        echo '& \"$dir\\rustup-init.exe\" -y --no-modify-path' | out-file $dir\\install-rustup.ps1
-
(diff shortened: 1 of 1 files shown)


r15ch13:
The script property isn't even necessary for the installer 😀
```json
    "installer": {
        "file": "rustup-init.exe",
        "keep": true,
        "args": [
            "-y",
            "--no-modify-path"
        ]
    },
```

matthewjberger:
When I tried that initially it opened up a new console window and failed with error code 121. Have you seen anything like that happen with previous apps?

r15ch13:
Hm, you're right, while using cmder it opens a new tab/window and fails.
Okay let's use your way 👍 (also because it shows the output in the same console 😀)

PR #1671, opened 2017-08-28
create apex/up

klauern:
I saw an apex.json, so the `up` command shouldn't be too far behind.  This is from [`apex/up`](https://github.com/apex/up), which is a service for deploying apps.  Compared to `apex`, `up` can deploy more single applications, and `apex` can install Functions-as-a-Service (Faas).


r15ch13:
Thanks 👍

r15ch13:
apex got downgraded by the update services because v0.15.0 is only a pre-release 😞

PR #1675, opened 2017-08-29
update haskell 8.2.1 / fix autoupdate / drop 32bit

rrelmy:
Haskell currently does not provide 32bit, the autoupdate part is
"commented out".

Original note from the haskell page
> The 8.2.1 platform release is for 64 bit Windows only, due to issues
> with GHC 8.2.1 on 32 bit Windows. Future releases are anticipated to
> resume 32 bit support.

### DIFF
--- a/bucket/haskell.json
+++ b/bucket/haskell.json
@@ -34,9 +34,6 @@
         "architecture": {
             "64bit": {
                 "url": "https://haskell.org/platform/download/$version/HaskellPlatform-$version-core-x86_64-setup.exe"
-            },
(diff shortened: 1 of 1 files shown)


r15ch13:
Except that `COMMENTED_OUT_UNTIL_32BIT_SUPPORT_IS_ADDED_AGAIN_32bit` is not valid according to the JSON Schema 😁

rrelmy:
Why not? 😁
I remove it and create a PR for it to be there when we need it

PR #1678, opened 2017-08-30
notepad-plus-plus 7.5.1

danijeljw:
Notepad++ was missing from the list of available applications to install


inosik:
Notepad++ already exists in the ["extras" bucket][extras]. You can add it with `scoop bucket add extras`.

  [extras]: https://github.com/lukesampson/scoop-extras

r15ch13:
@inosik is right :smile:
Closing this.

PR #1688, opened 2017-09-06
Ruby 2.4.1-2

alvr:
Update Ruby to v2.4.1-2

### DIFF
--- a/bucket/msys2.json
+++ b/bucket/msys2.json
@@ -20,6 +20,7 @@
             "msys2"
         ]
     ],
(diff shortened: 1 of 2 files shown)


r15ch13:
Nice! 👍 
I added some improvements to the setup.
It's now possible to use MSYS2 installed with scoop, instead of installing it with `ridk`.
Added some notes for clarification.

alvr:
Sorry for the late reply. Yes using the updated version of ruby and msys2 works perfectly.

PR #1689, opened 2017-09-06
Simplify rustup installer.

matthewjberger:
This installs rustup directly to the persistent data directory by setting the CARGO_HOME and RUSTUP_HOME environment variables for the process before running the rustup installer.


r15ch13:
Neat!

PR #1706, opened 2017-09-13
Update haxe.json

lguzzon:
For this version must be used "extract_dir" otherwise ... nothing work


r15ch13:
They have really inconsistent updates ... this change basically makes it impossible to auto update ...

PR #1709, opened 2017-09-15
Add replace for regexp, allow regexp and replace for jsonpaths

se35710:
Allows for more dynamic version extractions, eg when jsonpath doesn't exactly match the version, or when you want to rearrange the regexp matches for a better version string.

I've changed so now both `jp` and `re` are allowed at the same time. `jp` is executed first, and then `re` is applied on the jsonpath. If `replace` is defined, it is applied on the regexp match.

Examples:
```
    "checkver": {
        "url": "https://superduper.io/versions.json",
        "jp": "$..name",
        "re": "(?<version>[\\d.]+)",
        "replace": "2.${version}"
    },
```

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -61,6 +61,7 @@ $queue | % {
     }
     $regex = ""
     $jsonpath = ""
+    $replace = ""
 
     if ($json.checkver -eq "github") {
         if (!$json.homepage.StartsWith("https://github.com/")) {
@@ -83,6 +84,10 @@ $queue | % {
         $jsonpath = $json.checkver.jp
     }
 
+    if ($json.checkver.replace) {
+        $replace = $json.checkver.replace
+    }
+
     if(!$jsonpath -and !$regex) {
         $regex = $json.checkver
     }
@@ -96,6 +101,7 @@ $queue | % {
         json = $json;
         jsonpath = $jsonpath;
         reverse = $reverse;
+        replace = $replace;
     }
 
     $wc.headers.add('Referer', (strip_filename $url))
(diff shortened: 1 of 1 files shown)


r15ch13:
Cool 👍 
Do you have an example of the replace thing? Does it result in `2.1.0` if the regex matched `1.0`?

superduper.io 😁

se35710:
Yes, and it works with mutiple matching groups too, so `"${1}-${date}"` would create a version string of `"0.0.1-19620102"`.

The example above was with jsonpath, but this would work too.

```
    "checkver": {
        "url": "https://superduper.io/versions.html",
        "re": "(?<date>[\\d]+)\\sdupes-([\\d.]+)",
        "replace": "${1}-${date}"
    },
```

r15ch13:
When `checkver` is only a string (see 7zip, wget) then `$replace` would be the replace() function from System.String. 😁

r15ch13:
@nightroman does this work on your end?

nightroman:
> @nightroman does this work fine on your end?

It sounds good and I have use cases for this feature. I am on holiday, will try only in a week or two.

nightroman:
It seems to be working for me with a manifest like

```jscript
{
    "url":  "https://download.sysinternals.com/files/SysinternalsSuite.zip",
    "version":  "June.14.2017",
    "env_set":  {
                    "SysinternalsSuite":  "$dir"
                },
    "homepage":  "https://technet.microsoft.com/en-us/sysinternals/bb842062.aspx",
    "checkver":  {
                     "re":  "\\bUpdated:\\s*(\\w+)\\s+(\\d+),\\s+(\\d+)\\b",
                     "replace":  "${1}.${2}.${3}"
                 }
}
```

PR #1715, opened 2017-09-19
Make Vim installation more standard and less intrusive

niboan:
I've made a few adjustments to the Vim installation, to make it adhere more to Vim standards, and to be less intrusive.

The current installation has a few problems:

- Even if the user has an existing Vim configuration, eg. in `$HOME\vimfiles\vimrc,` a new vimrc file will be created at `$HOME\.vimrc` during installation. The new default configuration will take precedence over the existing configuration, and if the user deletes default vimrc file, a new one will be created next time Vim is updated.

- The installation will break any existing installation of Vim, as it sets the environment variable `VIM`.

To address these issues, I have made the following changes:

- Check for an existing vimrc file in (almost) all standard locations for Vim in Windows. (`:help vimrc`)
  The installation checks the following locations:
    `$HOME\_vimrc`
    `$HOME\.vimrc`
    `$HOME\vimfiles\vimrc`
    `$VIM\_vimrc`

- Use `_vimrc` instead of `.vimrc` as default vimrc file name. This is the standard for Vim in Windows. (`:help vimrc`)

- Don't set `VIM` environment variable.
  This should not be necessary in a normal installation, as Vim does a good job of guessing the right value. (`:help $VIM`)

### DIFF
--- a/bucket/vim.json
+++ b/bucket/vim.json
@@ -17,11 +17,11 @@
         "vim.exe",
         "gvim.exe"
     ],
-    "post_install": "if(!(test-path ~\\.vimrc)) {
-        cp \"$dir\\vimrc_example.vim\" ~\\.vimrc
-        \"set shell=$((gcm powershell).path)\\ -executionpolicy\\ bypass\" | out-file  ~\\.vimrc -append -encoding ascii
-        echo '~/.vimrc was created with your shell set to Powershell.'
-    } else { echo '~/.vimrc exists, skipping' }",
+    "post_install": "if( !(test-path ~\\.vimrc) -and !(test-path ~\\_vimrc) -and !(test-path ~\\vimfiles\\vimrc) -and !(test-path $env:VIM\\_vimrc) ) {
+        cp \"$dir\\vimrc_example.vim\" ~\\_vimrc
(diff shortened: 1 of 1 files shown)


r15ch13:
Sounds good 👍 
But does it effect https://github.com/lukesampson/scoop/commit/983cf1e0a2ae68e3b136a468f58a1bed4658773a?

niboan:
I don't think it does. I've tried a fresh install without setting $VIM, and everything seems fine, including syntax highlighting.
Doing an `:echo $VIM` in Vim shows the same directory, regardless of whether VIM is set in environment or not.

niboan:
Maybe the problem that https://github.com/lukesampson/scoop/commit/983cf1e0a2ae68e3b136a468f58a1bed4658773a fixed, was that if a user already had a his $VIM pointing to an existing (non-Scoop) Vim installation, the Scoop Vim would try to read syntax files from the old Vim.

The problem I had, was actually that when i started my old Vim, with $VIM pointing to Scoop Vim, it would complain about syntax files.

With no $VIM set, both installations will work.

r15ch13:
Okay, the first installation works on my machine (incl. highlighting) 👍

PR #1720, opened 2017-09-21
Add command for checking if an app is installed

Thor-Kristian-Valderhaug:
In my dayjob we use scoop to install apps on multiple servers. 
To be able to create install script that can reused and potensialy run on a server that already have some of the apps installed we need to be able to check is a app is alreay installed.

This PR creates a new command `installed`


lukesampson:
Your code looks great. But I'm wondering if we can get by without this given that Scoop already skips any apps that are already installed.

E.g. running `scoop install git cowsay wget` with `git` and `wget` already installed:

```
$ scoop install git cowsay wget
git is already installed. Skipping.
wget is already installed. Skipping.
Installing 'cowsay' (0.2013.07.19).
[...]
'cowsay' (0.2013.07.19) was installed successfully!
```

I can see the `installed` command could give you that info up-front with some extra work, but the end result would be the same as far as I can see. Or am I missing something?

There is a small extra cost that comes with every additional command — more documentation, more maintenance, and more opportunities for people to get confused about which command to use — e.g. I can see people wondering "do I need to run `installed` before every `install` to make sure it will work"?

Thor-Kristian-Valderhaug:
I agree that every additional command comes with a small extra cost. 
It looks like there is a difrence in the return code if you try to install one or mulitiple apps that are installed.
When you try to install a single app that is allready install the command exits with exitcode 1. Causing the rest of our scripts to fail

Single app:
```
scoop $ scoop install sudo
'sudo' (0.2017.03.24) is already installed.
Use 'scoop update sudo' to install a new version.
scoop $ $lastExitCode
1
```

Mulitiple apps:
```
scoop $ scoop install sudo sysinternals
sudo is already installed. Skipping.
sysinternals is already installed. Skipping.
scoop $ $lastExitCode
0
```

Is this difference intended ?

lukesampson:
Ah okay. No, that inconsistency is unintentional. Thanks for pointing it out.

I think the better behaviour would be to warn and return 0 for the single app case, to match the multi app case. Would that work better for your situation installing apps on multiple servers?

To fix this it's probably just a matter of calling Scoop's `warn` function instead of `abort`. If you're happy to submit a pull request that would be great.

rasa:
Based on the above dialog, it appears this PR can be closed. Please reopen, if I’m misunderstanding. Thanks!

PR #1727, opened 2017-09-26
Update openjdk to version 9+181 (blocked)

kuhsaft:
Java 9 has changed the version string to a new format and has a `+` before the build number.
[JEP 223: New Version-String Scheme](http://openjdk.java.net/jeps/223#Version-strings)

### DIFF
--- a/bucket/openjdk.json
+++ b/bucket/openjdk.json
@@ -1,11 +1,11 @@
 {
     "homepage": "https://adoptopenjdk.net/",
-    "version": "9+181",
+    "version": "9.181",
     "license": "GPL2",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/AdoptOpenJDK/openjdk9-releases/releases/download/jdk-9+181/OpenJDK9_x64_Win_jdk-9.181.zip",
-            "hash": "68fa2183288caa5b2be239b597a42917d9fce008ef44504a9eb85ddbc8a6d028",
+            "url": "https://github.com/AdoptOpenJDK/openjdk9-releases/releases/download/jdk-9%2B181/OpenJDK9_x64_Win_jdk-9.181.zip",
+            "hash": "223f646726a9bc0dfae5655a40ed11bf1db16b510b45d73c03e34e1102455ff7",
             "extract_dir": "jdk-9+181"
         }
     },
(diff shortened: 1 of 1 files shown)


kuhsaft:
It seems that scoop only accepts a valid semver as the manifest version which is a bit of a predicament. The new version-string scheme for Java explicitly states:

>  When $MAJOR is incremented, all subsequent elements are removed.

r15ch13:
Wow ... so complicated... 😞
* zip file with a dot: `OpenJDK9_x64_Win_jdk-9.181.zip`
* sha256 file with a dot: `OpenJDK9_x64_Win_jdk-9.181.sha256.txt`
* content of sha256 file with a plus: `68fa...d028 OpenJDK9_x64_Win_jdk-9+181.zip`

Made some adjustments to skip the plus sign entirely. In the meantime, their bot published a new release without changing the build number...

kuhsaft:
Ikr... 😔. I didn't notice AdoptOpenJDK's API endpoints which helps out a lot. Their Jenkins build bot seems to be running on a timer, so, the actual build number is the same. I wouldn't merge this yet though.

r15ch13:
Found those API endpoints by looking through their repos 😄

The homepage also shows "The AdoptOpenJDK Build Farm is still a Work In Progress. Please DO NOT use downloaded binaries for your production purposes.". So lets wait until this is stable. 👍

kuhsaft:
Agreed. I submitted an issue on their tracker about the inconsistent `.` `+` fiasco.

bucket/openjdk.json
```diff
@@ -1,11 +1,11 @@
 {
     "homepage": "https://adoptopenjdk.net/",
-    "version": "9+181",
+    "version": "9.181",
```
se35710: Should this not be a dash instead of a '.', eg `9-181`? I think it will be easier to read when 9.1.3+22.

kuhsaft: It should be a ‘+’ as specified by the JEP and my original commit. Build number is separated by a plus and on major version bumps all minor numbers are removed. So 9.0.0+181 => 9+181. It’s unnecessarily complicated imo.

kuhsaft:
Blocked by AdoptOpenJDK/openjdk-build#173

bucket/openjdk.json
kuhsaft: Even the oracle java download link labels it as `9+181`: `http://download.oracle.com/otn-pub/java/jdk/9+181/jdk-9_windows-x64_bin.exe`

se35710: Maybe I misunderstood, but I thought scoop had some issues with + as a version number divider, and therefor @r15ch13 suggested `.` instead of `+`. I just wanted to sync the scoop-extras PR with this one, so that both uses the same version numbers. If that is `+` or `-` I leave up to you guys to decide. I’ll update my PR in extras based on the decision.

kuhsaft: That's correct. Scoop doesn't accept `+` in the version string. I prefer the hyphen because it's the build number. It makes more sense because otherwise 9.181 > 9.0.1.xxx

osmancis:
Seems quite a while now, but any idea when this conflict resolved and PR merged? Else is there any way (say parameterized) we can pass java version from command line to install latest java?

r15ch13:
Solved by https://github.com/lukesampson/scoop/pull/2209

PR #1730, opened 2017-09-26
Update gpg to version 3.0.0

kripteria:
Gpg4win version 3.0.0 don't contain gpg2.exe

### DIFF
--- a/bucket/gpg.json
+++ b/bucket/gpg.json
@@ -1,15 +1,33 @@
 {
     "version": "3.0.0",
     "license": "GNU",
-    "url": "https://files.gpg4win.org/gpg4win-3.0.0.exe#/dl.7z",
+    "url": "https://files.gpg4win.org/gpg4win-3.0.0.exe",
     "hash": "2565bf6faf8defb8fa61b0b1a30f0e68e2ca5ceb3177d08516e00ca1620252bf",
     "homepage": "https://www.gpg4win.org",
-    "bin": "gpg.exe",
+    "env_add_path": [
+        "GnuPG\\bin",
+        "Gpg4win\\bin"
+    ],
     "checkver": {
         "url": "https://www.gpg4win.org/get-gpg4win.html",
         "re": "Download Gpg4win ([\\d.]+)"
     },
     "autoupdate": {
-        "url": "https://files.gpg4win.org/gpg4win-$version.exe#/dl.7z"
+        "url": "https://files.gpg4win.org/gpg4win-$version.exe",
+        "hash": {
+            "url": "https://gpg4win.org/package-integrity.html"
+        }
+    },
+    "installer": {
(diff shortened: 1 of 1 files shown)


r15ch13:
There isn't even a `gpg.exe`

kripteria:
```
gpg4win-3.0.0.exe
├── $TEMP
│   ├── gnupg-w32-2.2.1-20170919-bin.exe
│       ├── bin
│           ├── gpg.exe
```

r15ch13:
Now it uses the gpg4win installer instead of extracting it. Is this working for you?

kripteria:
Yes, it's works. But now I have UAC promt. Maybe replace Gpg4win with clean gpg from [gnupg.org](https://www.gnupg.org/download/index.html)

r15ch13:
Or we could just add gnupg as a separate manifest (extracting this will work fine). 😄
And we also extract Gpg4win. Don't know if these two will work together (but it's probably just a PATH thing)

kripteria:
**Gpg4win components:**
- gnupg - backend, cli
- Kleopatra - gui for gnupg
- GpgOL - outlook plugin
- GpgEX - shell extension

Anyway we can't use shell integrations without installer because it requires adding registry keys.
Old version manifest for gpg4win used extraction executable file and bin section contains only gpg and gpg2(now is gpg.exe).
If we use only cli app, why we download [gpg4win](https://www.gpg4win.org/index.html) instead [gnupg](https://www.gnupg.org/download/index.html)?

> Focuses on developer tools. While it would be easy to install Skype with Scoop, this will probably never be in Scoop's main bucket (app repository). Scoop focuses on open-source, command-line developer tools.
[Chocolatey Comparison](https://github.com/lukesampson/scoop/wiki/Chocolatey-Comparison)

r15ch13:
We are working in the Gpg issue in #1730

se35710:
Just a friendly reminder that GnuPG is in extras: [gnupg.json](https://github.com/lukesampson/scoop-extras/blob/master/gnupg.json).

Meeeeow:
Any progress?

r15ch13:
* Added [gpg4win.json](https://github.com/lukesampson/scoop-extras/blob/master/bucket/gpg4win.json) to the extras bucket
* Added [gnupg.json](https://github.com/lukesampson/scoop/blob/master/bucket/gnupg.json) as an [gpg.json](https://github.com/lukesampson/scoop/blob/master/bucket/gpg.json) alias to main bucket
* Switched gpg4win to gpg https://github.com/lukesampson/scoop/commit/9e191a107e80103033eb566ce28bfa1320d0a0da

rasa:
Based on the above dialog, it appears this PR can be closed. Please reopen, if I’m misunderstanding. Thanks!

PR #1748, opened 2017-10-06
add persist to maven conf

winteryoung:


winteryoung:
@r15ch13 Would you take some time to review this PR?

PR #1755, opened 2017-10-09
git: add new tig.exe to bin

toxeus:
This useful git browser was added to the 2.14.2 release.

See
https://github.com/git-for-windows/git/releases/tag/v2.14.2.windows.1


toxeus:
Not sure if scoop recognizes new versions by git commit hash or by the version number in the json file. For the latter case we'd need to append a package level version digit.

r15ch13:
Scoop uses git to update itself. Run `scoop update` and `scoop update git-with-openssh -f` should work.

toxeus:
So `-f` does the trick. Thx!

PR #1757, opened 2017-10-10
set working directory for shortcuts

winteryoung:
@r15ch13 I haven't test this. Do you have a low cost way to test core? I don't have a VM.

### DIFF
--- a/lib/shortcuts.ps1
+++ b/lib/shortcuts.ps1
@@ -28,6 +28,7 @@ function startmenu_shortcut($target, $shortcutName, $global) {
     $wsShell = New-Object -ComObject WScript.Shell
     $wsShell = $wsShell.CreateShortcut("$scoop_startmenu_folder\$shortcutName.lnk")
     $wsShell.TargetPath = "$target"
+    $wsShell.WorkingDirectory = "$dir"
     $wsShell.Save()
     write-host "Creating shortcut for $shortcutName ($(fname $target))"
 }


matthewjberger:
If you want a cheap and quick windows VM setup you can do this:

1.) Install VirtualBox
2.) run `scoop install vagrant`
3.) Make a directory named win7 and run the following two commands in it:

```
vagrant init matthewjberger/win7-64-en
vagrant up
```

4.) Then open virtualbox and you'll see your VM running. Select it and it will pop up the gui for the VM

Good luck!

Note:
You can spin up a new vm in this same manner and you'll already have the win7 base ready to go because it only needs to download the first time. Each VM you spin up is a copy of the base so you can safely delete it without destroying the original win7 base box that you downloaded.

r15ch13:
@winteryoung I test everything with my current production system. One year old Win10 installation.
I really should switch to using VMs, but I think they are annoying. (I often fix scoop issues between two rounds of PUBG :D)

matthewjberger:
I agree. They were annoying to me too until I started using vagrant! :)

winteryoung:
I think I will give docker for windows a try. The hyper-v solution seems promising.

r15ch13:
@winteryoung Keep in mind that VirtualBox and VMware Workstation Player will not work anymore when you install Hyper-V.

winteryoung:
@r15ch13 thx. I don't have them installed.

winteryoung:
There's a problem to test core. The [install script](https://raw.githubusercontent.com/lukesampson/scoop/master/bin/install.ps1) is coupled with lukesampson urls. Is there any guidelines to how to test core?

winteryoung:
If there's no guidelines, I will create a new script to install locally.

r15ch13:

PR #1760, opened 2017-10-10
New packages: `fd` and `qaac`

Congee:


tehbilly:
+1 Very much enjoy fd, would be great utility to add

r15ch13:
Made some adjustments! 👍

PR #1761, opened 2017-10-11
Add flyway

dsbert:
This pull request adds the flyway database migration tool

https://flywaydb.org/


r15ch13:
What about the `sql`, `drivers` and `conf` directories? Is it necessary to make them persistent? Are they required or will they be created for each project?

dsbert:
Is there a way to persist the files within the `drivers` directory without overwriting the entire folder upon update? Since that contains the `.jar` files that flyway is expecting to use, it would need to be merged with any updated versions.

r15ch13:
Sadly this is not possible. 😞

dsbert:
Then for now I think the best compromise is to persist the `conf`, `jars`, and `sql` folders.

PR #1768, opened 2017-10-16
Update Mono to version 5.4.0

Calinou:
This updates Mono to the latest version.

### DIFF
--- a/bucket/mono.json
+++ b/bucket/mono.json
@@ -1,9 +1,9 @@
 {
-    "version": "5.2.0.224",
+    "version": "5.4.0.201",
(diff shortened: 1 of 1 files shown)


r15ch13:
Thanks. But the update service got it earlier 😄

Calinou:
Thanks for merging :)

Would you be interested in adding 64-bit installers? I can do it, but I have some trouble with the automatic updater regex, as the file names of the 32-bit and 64-bit installers differ, especially since the 64-bit installer does not include Gtk#.

r15ch13:
The best would be to add gtk-sharp as a separate manifest and make it a dependency.

PR #1773, opened 2017-10-20
Add Gutenberg to bucket

Calinou:
Gutenberg is a static site generator written in Rust.

Note that currently, only 64-bit binaries are provided by upstream. Also, the binaries use the MSVC ABI, which require `vcredist` to be installed on the system in order to run, if I'm not mistaken.


r15ch13:
Nice 👍
The `extract_dir`property can be ommited.
Could you add auto update information? (See: https://github.com/lukesampson/scoop/wiki/App-Manifest-Autoupdate)

PR #1774, opened 2017-10-20
Add persistent GOPATH to go manifest.

matthewjberger:
This pull request sets the GOPATH for the go manifest to a persistent directory and adds the bin folder of that GOPATH to the PATH so that downloaded libraries (obtained with `go get`) are accessible after downloading.

I found myself having to add the default GOPATH (`%USERPROFILE%\go` after go 1.8) to the path manually so that libraries for linting and code completion were visible to Spacemacs after I obtained them with `go get`.

What do you think of this?


r15ch13:
Looks good!

PR #1775, opened 2017-10-20
Update leiningen to version 2.8.0

brandon93s:
Update to `2.8.0`. 

_Could_ use `autoupdate` and point to [latest stable](https://raw.githubusercontent.com/technomancy/leiningen/stable/bin/lein.bat) but run the risk of the checkver and this diverging resulting in hash failures. Thoughts?

### DIFF
--- a/bucket/leiningen.json
+++ b/bucket/leiningen.json
@@ -1,10 +1,15 @@
 {
     "version": "2.8.0",
     "license": "Eclipse Public License 1.0",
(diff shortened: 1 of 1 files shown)


r15ch13:
Made some changes 😄

brandon93s:
Awesome, learned something new! Wasn't aware you could target tags in the URL like that. 🏅

PR #1785, opened 2017-10-26
Update forge.json

Krzysztof-Cieslak:


r15ch13:
The automated update service would have grabbed it in 20min 👍

PR #1794, opened 2017-10-29
add JSON manifest for qpdf-7.00

oltolm:
Hello,

I have added a JSON manifest for [QPDF](http://qpdf.sourceforge.net/) a command-line program that does structural, content-preserving transformations on PDF files.


r15ch13:
Added `extract_dir` because `bin` property can't be updated.
Switched to sha512.

PR #1795, opened 2017-10-29
add manifest for cloc

oltolm:
I have added a manifest for [cloc](https://github.com/AlDanial/cloc).


r15ch13:
Adding `#/cloc.exe` to the end of the URL renames the file to `cloc.exe`.
The `bin` property can't be updated via auto update.

PR #1806, opened 2017-11-02
Do not use proxy if set to none

gitolicious:


r15ch13:
Related to https://github.com/lukesampson/scoop/issues/1728?

gitolicious:
Indeed... I missed that one!
After some further investigation, git also doesn't work with the currentuser@default trick. Need to dig deeper and will get back to you.

r15ch13:
`lib\config.ps1` is already loaded when `lib\git.ps1` is used, so `[net.webrequest]::defaultwebproxy` should return the correct address + credentials I guess.
Does something like this work? `$proxy = [net.webrequest]::defaultwebproxy`

I don't know and use any proxy, so I can't test it.

h404bi:
@r15ch13 @rasa close this, it's already implemented.

PR #1815, opened 2017-11-04
Add gogs at 0.11.29

rasa:


r15ch13:
This might be better suited for the extras bucket. The fork gitea is already there.

PR #1816, opened 2017-11-04
fix bfg powershell script

oltolm:
This commit fixes the powershell start script for bfg. Bfg did not start at all.


oltolm:
One of the goals of Scoop is not to pollute the environment. The existing BFG manifest adds an unnecessary environment variable `BFG_HOME` to the environment. This commit removes that variable.

PR #1822, opened 2017-11-06
Added retry when unzip fail because of AV

Hollow1838:
Trend AV analyses and lock zip files just after download. I have added 2 retries in case of this event.

At work I can't change antivirus settings so I have no other choice than to modify the unzip function.

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -155,51 +155,65 @@ function env($name,$global,$val='__get') {
     else { [environment]::setEnvironmentVariable($name,$val,$target) }
 }
 
-function unzip($path,$to) {
-    if(!(test-path $path)) { abort "can't find $path to unzip"}
+function isFileLocked([string]$path) {
+    $file = New-Object System.IO.FileInfo $path
+
+    if ((Test-Path -Path $path) -eq $false) {
+        return $false
+    }
+
+    try {
+        $stream = $file.Open([System.IO.FileMode]::Open, [System.IO.FileAccess]::ReadWrite, [System.IO.FileShare]::None)
+        if ($stream) {
+            $stream.Close()
+        }
+        return $false
+    }
+    catch {
+        # file is locked by a process.
+        return $true
+    }
+}
+
+function unzip($path, $to) {
+    if (!(test-path $path)) { abort "can't find $path to unzip"}
     try { add-type -assembly "System.IO.Compression.FileSystem" -ea stop }
     catch { unzip_old $path $to; return } # for .net earlier than 4.5
+    $retries = 0
+    while ($retries -le 10) {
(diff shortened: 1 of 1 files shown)


r15ch13:
What happens when you install 7zip? Does it still not work?

r15ch13:
Can you try the following?

Add this function to core.ps1
```powershell
function isFileLocked([string]$path) {
    $file = New-Object System.IO.FileInfo $path

    if((Test-Path -Path $path) -eq $false) {
        return $false
    }

    try {
        $stream = $file.Open([System.IO.FileMode]::Open, [System.IO.FileAccess]::ReadWrite, [System.IO.FileShare]::None)
        if ($stream) {
            $stream.Close()
        }
        return $false
    } catch {
        # file is locked by a process.
        return $true
    }
}
```
And change unzip() to:
```powershell
function unzip($path,$to) {
    if(!(test-path $path)) { abort "can't find $path to unzip"}
    try { add-type -assembly "System.IO.Compression.FileSystem" -ea stop }
    catch { unzip_old $path $to; return } # for .net earlier than 4.5

    $retries = 0
    while($retries -le 10) {
        if($retries -eq 10) {
            abort "Unzip failed: Too many retries!"
        }
        if(isFileLocked $path) {
            write-host "Waiting for $path to be unlocked by another process... ($retries/10)"
            $retries++
            Start-Sleep -s 2
        } else {
            break
        }
    }

    try {
        [io.compression.zipfile]::extracttodirectory($path,$to)
    } catch [system.io.pathtoolongexception] {
        # try to fall back to 7zip if path is too long
        if(7zip_installed) {
            extract_7zip $path $to $false
            return
        } else {
            abort "Unzip failed: Windows can't handle the long paths in this zip file.`nRun 'scoop install 7zip' and try again."
        }
    } catch {
        abort "Unzip failed: $_"
    }
}
```

Hollow1838:
Aww... still getting the error for very big files with both my code and your's. 
I am trying to install idea-ultimate-eap from extras bucket. Seems like the file is too big and don't get released any time soon... AV must be scanning every dll in the _tmp folder of the app.

Hollow1838:
Currently trying with 30 retries

Hollow1838:
Well 30 was not enough, he blocked on a different file, retrying with 60 retries.

Hollow1838:
Well seems like scoop reaches a sort of timout and won't proceed further.

Hollow1838:
OK! It worked, instead of default unzip, I used the "extract_7zip $path $to $false" by default. Seems like 7zip can deal with the waiting.

Hollow1838:
Thanks ^^

rasa:
@r15ch13: this is ready to merge, yes?

PR #1824, opened 2017-11-08
Remove runcmda.exe from smartmontools

rasa:


rasa:
Installs ok with this patch

PR #1831, opened 2017-11-11
create go shims instead of adding dir to path var

jfmherokiller:
This change makes the script create shims in the scoop shims directory instead of adding another directory to the path env variable. I believe this might be prefered instead of adding to the path variable.


r15ch13:
Good idea. Thanks 👍

jfmherokiller:
i did it because i had my path environment variable cleaned and lost access to everything that appends to the path var

PR #1841, opened 2017-11-12
Update sqlite to use semver and fix hash

### LINKED ISSUE #1840: Hash check failed for sqlite
```
Installing 'sqlite' (3210000).
sqlite-tools-win32-x86-3210000.zip (1.6 MB) [=================================================] 100%
Checking hash of sqlite-tools-win32-x86-3210000.zip... Hash check failed for 'https://www.sqlite.org/2017/sqlite-tools-win32-x86-3210000.zip'.
Expected:
    bee05e6e3d6c15a4b149cf879596e602a025c0c622b862bc2a2a056003d2a64b
Actual:
    515b0230860e3557cafc9b1e1207349c10a8a823f82af6d769d86fee99df4681
```

brandon93s:
Fixes #1840 

- Update to use the actual semver (`3.21.0`)
- Fix hash

### DIFF
--- a/bucket/sqlite.json
+++ b/bucket/sqlite.json
@@ -1,9 +1,9 @@
 {
     "homepage": "https://www.sqlite.org/",
-    "version": "3210000",
+    "version": "3.21.0",
     "license": "Public Domain",
     "url": "https://www.sqlite.org/2017/sqlite-tools-win32-x86-3210000.zip",
-    "hash": "bee05e6e3d6c15a4b149cf879596e602a025c0c622b862bc2a2a056003d2a64b",
+    "hash": "515b0230860e3557cafc9b1e1207349c10a8a823f82af6d769d86fee99df4681",
     "extract_dir": "sqlite-tools-win32-x86-3210000",
     "bin": [
         "sqlite3.exe",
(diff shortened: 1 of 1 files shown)


brandon93s:
@r15ch13 - Do you know if this is intentionally locked to x86? If not, I'll open a separate PR to add x64/x86 architecture support.

r15ch13:
`$cleanVersion00` should work (See [sqlite.json history](https://github.com/lukesampson/scoop/commits/master/bucket/sqlite.json))

brandon93s:
I considered that, but don't think it's quite that simple in practice. Here is the format that they publish:

> For version 3.X.Y the filename encoding is 3XXYY00. For branch version 3.X.Y.Z, the encoding is 3XXYYZZ

`$cleanVersion` just strips the periods, but the file versions from sqlite will pad single digit values with a zero. Examples:

```sh
v3.21.0 -> 3-21-00-00 -> 3210000
# would add '000' to the end

v3.20.1 -> 3-20-01-00 -> 3200100
# would add '0' before '1' and add '00' to the end

v3.20.1.2 -> 3-20-01-02 -> 3200102
# would add '0' before '1' and '0' before '2'
```

If there is concern over the ugly RegEx being stable, I can revert and stick with the long-form version numbers instead of flipping to use the actual version. Thoughts?

r15ch13:
`[\s\S]+` that's clever! Just checked your regex on regex101.com 😁 
You're right about the version. Did read it myself some time ago and decided to use only the number. But with this nifty regex, we can change it. 👍

PR #1842, opened 2017-11-13
Maven: remove M2_HOME variable, do not persist conf

oltolm:
I have removed the `M2_HOME` environment variable, it is not used by Maven. Maven uses the `%USERPROFILE%\.m2` directory. The `conf` directory does not have to be persisted. It contains only examples and is not used by Maven. Maven uses configuration files in the `%USERPROFILE%\.m2` directory.

### DIFF
--- a/bucket/maven.json
+++ b/bucket/maven.json
@@ -6,9 +6,6 @@
     "hash": "90e9f0d700743516d1f610fcd63a5d49751d0743db713909f9ef933747a38b8a",
     "extract_dir": "apache-maven-3.5.2",
     "env_add_path": "bin",
-    "env_set": {
-        "M2_HOME": "$dir"
-    },
     "suggest": {
         "JDK": [
             "extras/oraclejdk",


bucket/maven.json
```diff
@@ -25,8 +22,5 @@
         "hash": {
             "url": "$url.sha1"
         }
-    },
-    "persist": [
-        "conf"
```
brandon93s: > The conf directory does not have to be persisted. It contains only examples and is not used by Maven. Maven uses configuration files in the %USERPROFILE%\.m2 directory.

I believe the `conf` directory is the _global_ configuration directory used by Maven and should be persisted:

image

bucket/maven.json
```diff
@@ -6,9 +6,6 @@
     "hash": "90e9f0d700743516d1f610fcd63a5d49751d0743db713909f9ef933747a38b8a",
     "extract_dir": "apache-maven-3.5.2",
     "env_add_path": "bin",
-    "env_set": {
-        "M2_HOME": "$dir"
```
brandon93s: `M2_HOME` was [deprecated](https://maven.apache.org/docs/3.5.0-alpha-1/release-notes.html) in 3.5.0-alpha and is no longer used by the `current` version.

> Based on problems in using `M2_HOME` related to different Maven versions installed and to simplify things, the usage of `M2_HOME` has been removed and is not supported any more MNG-5823, MNG-5836, MNG-5607.

With that being said, if a scoop user decided to run `scoop install maven@3.3.9` they would prefer to have `M2_HOME`. However, since version support isn't widely publicized, it's probably safe to go ahead and remove it. 👍

oltolm:
OK, I have reverted the `conf` change.

PR #1843, opened 2017-11-14
Gradle: remove GRADLE_HOME from manifest

oltolm:
This commit removes the `%GRADLE_HOME%` environment variable. It is not used by Gradle. See this [discussion](https://discuss.gradle.org/t/do-we-really-need-gradle-home-docs-site-inconsistency/21671).


oltolm:
Will this be merged?

PR #1847, opened 2017-11-15
Adding hash values to jq.json

Ghasan:
I installed jq for work today, and realized it does not contain hashes as scoop gave warnings.

The official project repo does not contain hashes, so I downloaded the binaries twice, and calculated the hashes and put them here.

I tested it with 'scoop install jq.json' locally and it worked fine for the 64


r15ch13:
Totally missed that one. 😁

PR #1855, opened 2017-11-17
Adding configuration support for nano.

Ghasan:
This is to add support for nano configuration. nano Windows build expect a 'HOME' variable to point to the home folder of the current user, and it reads the configuration from there by looking for nano.rc file. https://www.nano-editor.org/dist/v2.8/faq.html#3.9b

There are no options to change nano to look at somewhere else.

I have added a check so that nano.rc is not copied to the home folder if it already exists there, and I put a note about the environment set, and the nano.rc placed at the home folder.

### DIFF
--- a/bucket/nano.json
+++ b/bucket/nano.json
@@ -5,14 +5,11 @@
     "url": "https://www.nano-editor.org/dist/v2.5/NT/nano-2.5.3.zip",
     "hash": "c012456b9f29e488b0e9a688aa43a34326dd573f9e99765cbdac2b81a4bc1060",
     "bin": "nano-2.5.3-win32\\nano.exe",
-    "env_set": {
-        "HOME": "$env:USERPROFILE"
-    },
-    "post_install": "(-not (Test-Path ~/nano.rc)) -and (Copy-Item $dir/nano-2.5.3-win32/nano.rc ~/nano.rc)",
+    "env_ensure_home": true,
+    "post_install": "if (-not (Test-Path ~/nano.rc)) { Copy-Item $dir/nano-2.5.3-win32/nano.rc ~/nano.rc }",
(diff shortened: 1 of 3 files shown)


oltolm:
If you change the value of `%HOME%` it will break other apps like Emacs which also use `%HOME%`. Emacs won't be able to find its configuration. Scoop should never change the value `%HOME%`.

Ghasan:
Doesn't Emacs `HOME` variable point to the user profile? If it does, then it should be fine. Because `HOME` indicates that it points to the user profile, if not, then it defies its own meaning.

This is a screenshot of my user profile, in which many programs store their configurations.

user profile

This is if I am not mistaken mentioned [here](https://github.com/lukesampson/scoop/blob/master/lib/install.ps1#L55), but it is commented out for some reason.

r15ch13:
@Ghasan the line you mentioned was changed with this commit: https://github.com/lukesampson/scoop/commit/a04b69451a4efe771de11024addea56197b43d0b

Ghasan:
Thank you, @r15ch13. I have made some changes and now I use "env_ensure_home" flag. I tested it, and it worked. However, this probably needs some approval as it enables a new property.

r15ch13:
@Ghasan did you read the issue https://github.com/lukesampson/scoop/issues/687?
I would recommend adding a note that explains what to do, instead of adding `HOME`.

Ghasan:
Thank you, @r15ch13. I read that, and I checked Yarn, Vim, and Emacs, and they seem to not require %HOME% variable to be present, and they can find user profile themselves, which I think now is the correct way. So I removed the changes I made, and it now only copies nano.rc from the installation folder to user profile, and gives a note about setting %HOME% for nano.

PR #1856, opened 2017-11-20
Use official GnuPG binaries, not GPG4Win, and revert version from 3.0 to 2.2.3.

Ghasan:
I tried installing gpg yesterday, and I ran into some errors. It seemed neither gpg.exe, nor gpg2.exe actually existed in the package downloaded. GPG4Win actually contains the installer of GnuPG, but it does not contain the binaries themselves ready for referencing.

Moreover, GnuPG is in version 2.2.3, not 3.0.0. That is GPG4Win software bundle version.

So I made these "breaking changes", as it reverts version from 3.0.0 to 2.2.3, but this is necessary (as the previous one is not the version of GnuPG, but GPG4Win), and I referenced the official binaries from gnupg.org, and updated checkver and autoupdate accordingly.

I tested it, and it worked fine.

### DIFF
--- a/bucket/gpg.json
+++ b/bucket/gpg.json
@@ -1,22 +1,18 @@
 {
-    "version": "3.0.0",
-    "license": "GNU",
-    "url": "https://files.gpg4win.org/gpg4win-3.0.0.exe#/dl.7z",
-    "hash": "2565bf6faf8defb8fa61b0b1a30f0e68e2ca5ceb3177d08516e00ca1620252bf",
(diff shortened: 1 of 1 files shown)


r15ch13:
GnuPG is already in the extras bucket [gnupg.json](https://github.com/lukesampson/scoop-extras/blob/master/gnupg.json) and there is a PR https://github.com/lukesampson/scoop/pull/1730

Ghasan:
Thanks, @r15ch13. It looks good now.

Issue #1237, opened 2017-01-20
[maven] add sh version of mvn to bin


jinahya:
Apache Maven works just find.
Is it possible to add `mvn` which has no extension along with `mvn.cmd`?
I tend to use cygwin which is a linux emulation and it's not possible to use `mvn` command.
`alias mvn=mvn.cmd` [doesn't expand in non-interactive shell](http://unix.stackexchange.com/a/1499/8557).

inosik:
I prefer using MINGW most of the time, and noticed the same for Yarn, Gradle and others apps, which don't have an EXE-File.

@jinahya The solution described in the ["Can I use Scoop in bash?" wiki page][wiki-page] works here as well. Just replace `scoop.ps1` with `mvn.ps1`, for example.

  [wiki-page]: https://github.com/lukesampson/scoop/wiki/Can-I-Use-Scoop-In-Bash%3F

jinahya:
FIY, I put `alias mvn="powershell -ex unrestricted mvn.ps1"` into `~/.bashrs` and invoking `mvn` launches a new command console window. Thanks, anyway.

rasa:
This issue belongs in the [Main](https://github.com/ScoopInstaller/Main) bucket. Unfortunately, we cannot move this issue on this end, so it is being closed.
If this is still an issue, please create a new issue in the Main bucket via https://github.com/ScoopInstaller/Main/issues/new?body=Moved+from+https%3A%2F%2Fgithub.com%2Flukesampson%2Fscoop%2Fissues%2F1237

PR #1857, opened 2017-11-20
Maven: do not add directory to PATH

oltolm:
I have removed the `bin` directory from `%PATH%` and added `mvn.cmd` to it.

### DIFF
--- a/bucket/maven.json
+++ b/bucket/maven.json
@@ -5,7 +5,10 @@
     "url": "https://archive.apache.org/dist/maven/maven-3/3.5.2/binaries/apache-maven-3.5.2-bin.zip",
     "hash": "90e9f0d700743516d1f610fcd63a5d49751d0743db713909f9ef933747a38b8a",
     "extract_dir": "apache-maven-3.5.2",
-    "env_add_path": "bin",
+    "bin": [
+        "bin\\mvn.cmd",
+        "bin\\mvnDebug.cmd"
+    ],
     "suggest": {
         "JDK": [
             "extras/oraclejdk",


brandon93s:
`mvnDebug.cmd` is often used for remote debugging. Please add it to the `bin` as well since this PR will make it unavailable.

oltolm:
OK, I have added `mvnDebug.cmd` to `bin`.

r15ch13:
The `env_add_path` property is used intenionally because Cygwin users can only execute `mvn` bash script which doesn't get a shim from scoop. #1237 #1415 #1544

rasa:

PlotCitizen:
Hi all, I have just installed maven, however a shim is not created. Can you please advise if this is per expectation?

image
image

rashil2000:
Shims are not created because the `bin` folder is appended to the PATH

tauinger-de:
It would help me a lot if the scoop installer would notify me about changes to the PATH. 

It would allow me to realize why e.g. the `mvn` command isn't available even after installation (-> need to re-open terminal to make PATH changes apply)

Should I open a new issue for this?

rashil2000:
When you install a package which modifies PATH, it notifies you (for each folder that is appended to PATH).

tauinger-de:
Sorry, but I cannot confirm that:
```
C:\Temp>scoop install maven
Installing 'maven' (3.9.6) [64bit] from main bucket
apache-maven-3.9.6-bin.zip (9,1 MB) [===============================================================================] 100%
Checking hash of apache-maven-3.9.6-bin.zip ... ok.
Extracting apache-maven-3.9.6-bin.zip ... done.
Linking ~\scoop\apps\maven\current => ~\scoop\apps\maven\3.9.6
Persisting conf
Persisting lib\ext
'maven' (3.9.6) was installed successfully!
```

PR #1858, opened 2017-11-20
Ant: do not add directory to PATH

oltolm:
I have removed the `bin` directory from `%PATH%` and added `ant.cmd` to it.


brandon93s:
Should we add any of  `antenv.cmd`, `setenv.cmd`, `envset.cmd` or  `runrc.cmd` to bin as well? These would've been previously available but not after this PR.

oltolm:
I can add them, although I've never used any of the myself. What do others think?

rasa:

PR #1862, opened 2017-11-21
Yarn: use the preset location for cache and global bin

h404bi:
It should be a better way, I think.
(this will break previous presist.)

### DIFF
--- a/bucket/yarn.json
+++ b/bucket/yarn.json
@@ -13,16 +13,13 @@
     },
     "url": "https://yarnpkg.com/downloads/1.3.2/yarn-1.3.2.msi",
     "hash": "d2a4f6887381df152b0595fcf3b5590c6a7bc7b380ed364ad9f11595d099664d",
-    "persist": [
-        "cache",
-        "bin"
-    ],
-    "post_install": "
-        yarn config set cache-folder \"$dir\\cache\"
-        yarn config set prefix \"$dir\"
-    ",
+    "installer": {
+        "script": "add_first_in_path \"$env:LOCALAPPDATA\\Yarn\\bin\" $global"
+    },
+    "uninstaller": {
+        "script": "remove_from_path \"$env:LOCALAPPDATA\\Yarn\\bin\" $global"
+    },
     "env_add_path": [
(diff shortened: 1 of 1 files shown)


brandon93s:
Are these changes addressing an issue you're experiencing with yarn? 

> (this will break previous presist.)

It will also break all currently installed global applications. This manifest was just updated to fix global installs with https://github.com/lukesampson/scoop/commit/f783c38d58ba648e2031769d8043843c5441ef15. This will also move the yarn cache and shims out of `scoop` and into the users `localappdata`. For predictability, I think I prefer them to live within the persist dir.

h404bi:
Cuz I found that those **real** global bin, more specifically, all global installed node modules are still store within `%LOCALAPPDATA%\\Yarn`, in `%LOCALAPPDATA%\\Yarn\\config\\global`, though we relocate `bin` and `cache` directories into scoop.

e.g.: `~\scoop\persist\yarn\bin\gulp.cmd`
``` bat
@"%~dp0\..\..\..\..\..\AppData\Local\Yarn\config\global\node_modules\.bin\gulp.cmd"   %*
```

> Yarn keeps adding stuff to `%LOCALAPPDATA%` anyway, so there is no real benefit for adding persistence to bin and cache.
> —— [@r15ch13](https://github.com/lukesampson/scoop/commit/e4cc2c3c3115a51b6cba60a652c7ee66b85a9b1b#commitcomment-22621511)

It's a little tricky that we seperate cache and global bin from the actual config things to let scoop can presist them for us to use global applications. Now for that, we can just add `%LOCALAPPDATA%\\Yarn\\bin`, the default directory of shims of Yarn, to PATH. Keep its original feature, and simpler, compared to current workaround which relies on `~\.yarnrc`.

Looks like I'm SO OCD. Lol.

oltolm:
I have created a bucket for [nonportable](https://github.com/oltolm/scoop-nonportable) apps. It has a manifest for Node where I did something similar, so that npm stores its packages in `APPDATA\npm` instead of the nonstandard `scoop\persist` directory.

PR #1880, opened 2017-11-28
Update gcc to 7.2.0

kuhsaft:


kuhsaft:
It seems that I forgot to actually bump the version number. 😦

PR #1881, opened 2017-11-28
Adding Cake build tool.

Ghasan:


r15ch13:
Cool!

bucket/cake.json
```diff
@@ -0,0 +1,20 @@
...
+    "license": "MIT",
+    "url": "https://github.com/cake-build/cake/releases/download/v0.23.0/Cake-bin-net461-v0.23.0.zip",
+    "hash": "1ad334de4188e6dd320a8501dd0d76f5efc8f2f4e83a562fb3f309a298f39a4a",
+    "bin": "cake.exe",
+    "suggest": {
+        "NuGet": [
+            "nuget"
+        ]
+    },
+    "checkver": {
```
rasa:
````
”checkver”: {
“github”: “https://github.com/cake-build/cake”
}
````
Would,work, as well

r15ch13: yup https://github.com/lukesampson/scoop/commit/3cb03bec49daf7d9fa38068ba3d333b3bfc9c204

Ghasan: All the builds currently are marked as pre-release, and I thought it is better an explicit regex, as I read that "github" property ignores pre-releases.

https://github.com/lukesampson/scoop/wiki/App-Manifest-Autoupdate

rasa: Good point. @r15ch13?

r15ch13: https://github.com/cake-build/cake/releases/latest will fall back to the list of pre-releases and then the first result is used. I see no problem there 😄

Ghasan: Interesting. Thanks all for the help 👍

PR #1902, opened 2017-12-11
Initial implementation of command to query virustotal.com

pcrama:
Before installing packages, I like to check they are clean...  Luckily, the hash of the package can be used to retrieve the status on https://virustotal.com and more often than not, they have been analyzed already.

I don't know what to test for this (I looked at scoop-home.ps1 for inspiration, but it doesn't seem there are any tests either).

Best regards,

Philippe Crama

### DIFF
--- a/libexec/scoop-virustotal.ps1
+++ b/libexec/scoop-virustotal.ps1
@@ -1,0 +1,45 @@
+# Usage: scoop virustotal <app> [options]
+# Summary: Look for app's SHA256 on virustotal.com
+# Help: Look for app's SHA256 on virustotal.com
+#
+# The download's hash is also a key to access VirusTotal's scan results.
+# This allows to check the safety of the files without even downloading
+# them in many cases.
+#
+# Options:
+#   -a, --arch <32bit|64bit>  Use the specified architecture, if the app supports it
+
+. "$psscriptroot\..\lib\core.ps1"
+. "$psscriptroot\..\lib\help.ps1"
+. "$psscriptroot\..\lib\getopt.ps1"
+. "$psscriptroot\..\lib\manifest.ps1"
+. "$psscriptroot\..\lib\buckets.ps1"
+
+reset_aliases
+
+$opt, $apps, $err = getopt $args 'a:' 'arch='
+if($err) { "scoop virustotal: $err"; exit 1 }
+$architecture = ensure_architecture ($opt.a + $opt.arch)
+
+if($apps) {
+    $apps | % {
+        $manifest, $bucket = find_manifest $_
+        if($manifest) {
+            if([string]::isnullorempty($manifest.homepage)) {
+                abort "Could not find homepage in manifest for '$_'."
+            }
+            $h = hash $manifest $architecture
+            if ($h) {
+                $h | % { start "https://www.virustotal.com/#/file/$_/detection" }
+            }
+            else {
+                write-host -f darkred "No hash information for $_"
+            }
+        }
+        else {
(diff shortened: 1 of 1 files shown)


r15ch13:
Nice. Had this in mind for a while. Showing the results in the console would be cool.
Checking all programs would be nice too, using the API (but a key is needed for that)

r15ch13:
This works quite nicely without the API limitation 😁 

```powershell
. "$psscriptroot\..\lib\json.ps1"

$result = (new-object net.webclient).downloadstring("https://www.virustotal.com/ui/files/$hash")
$stats = json_path $result '$.data.attributes.last_analysis_stats'
$malicious = json_path $stats '$.malicious'
$suspicious = json_path $stats '$.suspicious'
$undetected = json_path $stats '$.undetected'
$unsafe = [int]$malicious + [int]$suspicious
if($unsafe -gt 0) {
    write-host -f red "$apps`: $unsafe/$undetected"
} else {
    write-host -f green "$apps`: $unsafe/$undetected"
}
```

pcrama:
Hi,

Thank you for your suggestion & research into making this directly available to the command line output.  I have updated my patch to work across more packages (apparently, VirusTotal allows to query using different hash alogrithms).
Next time I have 30 min, I will integrate your proposal.

Regards,

Philippe Crama

pcrama:
Hi,

I have no further ideas for improving this PR.  Is there anything else I would have to do to get this merged?

Thanks in advance,

Philippe Crama

r15ch13:
Just one small thing. Instead of showing a warning that the hash is not supported, it could just compute it and send it anyway. [compute_hash($file, $algname)](https://github.com/lukesampson/scoop/blob/master/lib/install.ps1#L458)

pcrama:
I'll look into it.  Though doesn't this imply that the package needs to be downloaded first?  This makes the runtime of the command hard to predict as sometimes a download is needed, sometimes not.  Maybe with an option to disable/enable it?  But then what should be the default?

r15ch13:
Yes, you're right. Let's add it like this.

I would like to propose the following change to reduce parentheses nesting a bit :smile:
```powershell
if(!$apps) {
    my_usage; exit 1
}

$apps | % {
    $app = $_
    $manifest, $bucket = find_manifest $app
    if(!$manifest) {
        $exit_code = $exit_code -bor $_ERR_NO_INFO
        write-host -f darkred "Could not find manifest for '$app'"
        continue
    }

    $hash = hash $manifest $architecture
    if (!$hash) {
        $exit_code = $exit_code -bor $_ERR_NO_INFO
        write-host -f darkred "No hash information for $app"
        continue
    }

    $url = url $manifest $architecture

    # Hacky way to see if $h is an array (i.e. there was a
    # list of hashes in the manifest) or a string
    # (i.e. there was 1! hash in the manifest).
    if ($hash[0].Length -eq 1) {
        # Wrap download URL in array to traverse it in
        # lockstep with the loop over the hash.
        $url = @(url $manifest $architecture)
    }

    $hash | % { $i = 0 } {
        try {
            $exit_code = $exit_code -bor (Start-VirusTotal $_ $app)
        } catch [Exception] {
            $exit_code = $exit_code -bor $_ERR_EXCEPTION
            if ($_.Exception.Message -like "*(404)*") {
                SubmitMaybe-ToVirusTotal $u[$i] $app ($opt.scan -or $opt.s)
            }
            else {
                write-host -f darkred "$app`: error fetching information`: $($_.Exception.Message)"
            }
        }
        $i = $i + 1
    }
}
```

PR #1903, opened 2017-12-12
Add manifest for kitematic

caveman-dick:
Tested autoupdate and works fine:

image

### DIFF
--- a/bucket/kitematic.json
+++ b/bucket/kitematic.json
@@ -1,0 +1,21 @@
+{
+    "version": "0.17.3",
+    "license": "Apache License 2.0",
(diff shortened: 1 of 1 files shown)


r15ch13:
This is better suited for the [extras bucket](https://github.com/lukesampson/scoop-extras). Please create a new PR over there. 👍 
The `depends` property can be removed if empty.

caveman-dick:
@r15ch13 I can move over no problem, I'm just wondering why that rather than in the main bucket? Because it's not a widely known/used tool?

r15ch13:
Because the main bucket is mostly for self-containing CLI programs. Everything with a Window goes to extras 😄

caveman-dick:
Moved to extras as requested: https://github.com/lukesampson/scoop-extras/pull/692

PR #1914, opened 2017-12-17
Add digdag (distributed workflow tool)

klauern:
Digdag is a distributed workflow tool, pointed out in the "assess" branch of Languages and Frameworks on Thoughtworks 2017 radar: https://www.thoughtworks.com/radar/languages-and-frameworks


rasa:
@klauern Can you add a return to the end of the last line, to make our CI happy?

klauern:
fixed.  I also noticed that it wasn't even in the right path....

rasa:
Sorry, but the CI now says: The following files have utf-8 BOM. Can you fix?

klauern:
Sorry about that.  You'd think that editing it with VSCode (and the editorconfig extension installed/enabled) would have fixed these by default for me.  OH well.  In any event, I think things are working.

PR #1918, opened 2017-12-21
Update virustotal command

rasa:
Add * option to check all installed apps
Add --global option to check globally installed apps
Add 60+ sec. pause after 4 queries/min to avoid 429/Too Many Requests
Add immediate abort if query limit is exceeded
Add INFO/WARN/ERROR prefixes to messages
Add DarkGray/DarkCyan/DarkRed colors to INFO/WARN/ERROR messages
Add using apikey if stored via: scoop config virustotal_api_key apikey
Change colors to DarkGreen/DarkYellow/Yellow/Red for 0/1/2/3+ hit count
Remove fragment from URL when submitting to virustotal.com
Standardized INFO/WARN/ERROR messages

### DIFF
--- a/libexec/scoop-virustotal.ps1
+++ b/libexec/scoop-virustotal.ps1
@@ -20,6 +20,7 @@
 #
 # Options:
 #   -a, --arch <32bit|64bit>  Use the specified architecture, if the app supports it
+#   -g, --global Scan globally installed apps too
 #   -s, --scan For packages where VirusTotal has no information, send download URL
 #              for analysis (and future retrieval)
 
@@ -29,12 +30,14 @@
 . "$psscriptroot\..\lib\manifest.ps1"
 . "$psscriptroot\..\lib\buckets.ps1"
 . "$psscriptroot\..\lib\json.ps1"
+. "$psscriptroot\..\lib\config.ps1"
 
 reset_aliases
 
-$opt, $apps, $err = getopt $args 'sa:' @('arch=', 'scan')
+$opt, $apps, $err = getopt $args 'a:gs' @('arch=', 'global', 'scan')
 if($err) { "scoop virustotal: $err"; exit 1 }
 $architecture = ensure_architecture ($opt.a + $opt.arch)
+$global = $opt.g -or $opt.global
 
 $_ERR_UNSAFE = 2
 $_ERR_EXCEPTION = 4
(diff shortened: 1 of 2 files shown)


libexec/scoop-virustotal.ps1
```diff
@@ -104,45 +128,75 @@ Function SubmitMaybe-ToVirusTotal ($url, $app, $do_scan) {
...
     my_usage; exit 1
 }
 
+if($global -and !(is_admin)) {
+    Error('You need admin rights to update global apps.')
+    exit 1
+}
+
+if(is_scoop_outdated) {
+    update_scoop
```
r15ch13: Should be `scoop update` instead

libexec/scoop-virustotal.ps1
```diff
@@ -104,45 +128,75 @@ Function SubmitMaybe-ToVirusTotal ($url, $app, $do_scan) {
...
-        write-host -f darkred "$app`: unknown, download $url & submit to https`://www.virustotal.com/"
+        Warn("$app`: not found`: manually submit $url")
     }
 }
 
 if(!$apps) {
     my_usage; exit 1
 }
 
+if($global -and !(is_admin)) {
```
r15ch13: `--global` option isn't necessary because it only uses the manifest files. It doesn't check the installed program itself.

libexec/scoop-virustotal.ps1
r15ch13: Should be `scoop update` instead

libexec/scoop-virustotal.ps1
```diff
@@ -29,35 +30,58 @@
...
 $architecture = ensure_architecture ($opt.a + $opt.arch)
+$global = $opt.g -or $opt.global
 
 $_ERR_UNSAFE = 2
 $_ERR_EXCEPTION = 4
 $_ERR_NO_INFO = 8
 
 $exit_code = 0
 
+Function Info($msg) {
```
r15ch13: INFO/WARN/ERROR functions should be moved to core.ps1 (we should add a better logging/output system anyway 😁)

rasa: Agreed.

libexec/scoop-virustotal.ps1
rasa: Gotcha. I'll remove.

libexec/scoop-virustotal.ps1
rasa: Gotcha. I'll fix.

pcrama:
@rasa & @r15ch13, thank you for carrying this work further!

I didn't see this update and will work on it a bit more.

One thing that already stands out as strange to me is that only installed applications may be checked.  My idea (or use case) is to check, preferrably even before downloading, that the software isn't infected.  

If I can do it, I would even try to check dependencies that would be installed, too [e.g. checking youtube-dl should automatically check ffmpeg], again, before they are installed.

I like the idea of using '*' to test all installed apps, but I am not too sure that this is entirely clear for the user: AFAIU, the hashes of the most up to date versions will be checked in VirusTotal, independently of the exact version actually currently installed.  I struggle with how to make this clear (in the documentation or even better in the command output).

I will open a PR with changes/proposals soon (WIP is in https://github.com/pcrama/scoop/tree/virustotal-apikey).

rasa:
One additional thought. What happens if a user installs a particular version with `scoop install 7zip@16.04` Do we support that now, and if not, can we?

pcrama:
As far as I know, this is not supported.  I have currently no idea if it can be.

pcrama:
I have submitted #1934 (https://github.com/pcrama/scoop/tree/virustotal-apikey-fixes-20171231), the other branch will be deleted.

PR #1923, opened 2017-12-23
Update cygwin to 2.883

rasa:
Removed checkver and autoupdate as these don't work. The version number isn't even in the metadata for setup.exe, so we'll use the date of the file for the version.

### DIFF
--- a/bucket/cygwin.json
+++ b/bucket/cygwin.json
@@ -1,5 +1,5 @@
 {
-    "version": "2017.12.11",
+    "version": "2.883",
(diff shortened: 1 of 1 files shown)


r15ch13:
The actual setup version can be found here as tags: https://cygwin.com/git/?p=cygwin-apps/setup.git;a=tags
It's only shown on the first page of the setup wizard.
The auto update property can stay in my opinion. :smile:

rasa:
Agreed. I will re-add and fix the checkver to grab the 2.883 version.

PR #1934, opened 2017-12-31
Followup to PR #1918: scoop-virustotal.ps1 fixes

pcrama:
As stated in #1918, I have made some more changes.  I believe that most of them are uncontroversial, except maybe 238efbc5.

- 8804c4e0 Check dependencies with VirusTotal, too (by default)
- 238efbc5 Remove requirement to only verify installed apps
- bdc6409b Color undecided VirusTotal information the same way as `dangerous' files
- d42668a1 Use API to submit download link to VirusTotal, rate limited in EAFP fashion
- 8bb1f5c6 Warn once when virustotal_api_key's absence prevents VirusTotal submission
- 42dde1a3 Refactor logic to warn user about apps unknown to VirusTotal
- 3e00f7ef Never use virustotal_api_key to query if a package is safe
- 54f00055 Improve documentation of virustotal subcommand


pcrama:
@r15ch13, is there anything I need to do to get this merged?

As far as I can tell from @rasa's 👍 , he doesn't seem to mind that I reverted the limitation that only pre-installed apps could be scanned.

Thanks in advance!

rasa:
LGTM. Thanks for a great addition.

rasa:
@r15ch13: assigning to you to merge, but if you want to have it reviewed more, please reassign.

libexec/scoop-virustotal.ps1
```diff
@@ -155,15 +198,20 @@ if(is_scoop_outdated) {
 $apps_param = $apps
 
 if($apps_param -eq '*') {
-    $apps = applist (installed_apps $false) $false
-} else {
-    $apps = ensure_all_installed $apps_param
+    $apps = (applist (installed_apps $false) $false) | ForEach-Object {
+                ($a, $_global_flag_to_drop) = $_
+                $a
```
r15ch13: What does this `$a` do?

pcrama: `applist` returns a list of tuples: the application name and a flag saying whether the application is installed globally.  For my purposes, I only need the application name.

It is a way to transform the list `@(@("app1", flag1), @("app2", flag2) )` into `@("app1", "app2")`.

The change was needed because I now have two sources of the list of apps to check with the VirusTotal service `installed_apps` (returning a list of tuples, as discussed above) and `install_order` (see [line 215 (was 208)](https://github.com/pcrama/scoop/blob/b3221133785be10b394fb6cd7ce69885a252e6e4/libexec/scoop-virustotal.ps1#L215)) to also include the list of dependencies in the scan.

The flag returned by `applist` for each application used to be dropped from that tuple in the final loop (see change from [(previously) line 166](https://github.com/pcrama/scoop/blob/af41fc6efb0fdf65943b8639ca5d44a406445452/libexec/scoop-virustotal.ps1#L166) to [line 221](https://github.com/pcrama/scoop/blob/b3221133785be10b394fb6cd7ce69885a252e6e4/libexec/scoop-virustotal.ps1#L221))

As you can see I pushed a comment in the source code about this.

If there's a better way in PowerShell to do this transformation, I'm all ears.

Thank you for looking at my PR!

r15ch13: `return $a` would have made it more clear I guess. I simply can't wrap my head around the fact that PowerShell returns the last variable by itself ... 😁

pcrama:
Thank you for your review.

One thing I wonder about is [line 205](https://github.com/pcrama/scoop/blob/18a59630e856ba9297fccb5e52b924be2dabd5f0/libexec/scoop-virustotal.ps1#L205): isn't that redundant after [line 63](https://github.com/pcrama/scoop/blob/18a59630e856ba9297fccb5e52b924be2dabd5f0/libexec/scoop-virustotal.ps1#L63)?

Note that I haven't tested the code, so this is just from reading the diffs.

r15ch13:
It's required for https://github.com/pcrama/scoop/blob/18a59630e856ba9297fccb5e52b924be2dabd5f0/libexec/scoop-virustotal.ps1#L69

pcrama:
Thank you for taking the time to polish my code.

I don't know if any comment is expected from my side, but just in case: I don't have anything to add to it anymore at this stage.

The automated submission part is still a bit flaky in the sense that packages aren't always scanned, but I believe that this is due to some size limit imposed on the submission API because for smaller packages it has worked for me.  Manual submission of the packages is still possible.  Anyway, I have no intention of working on that in the coming weeks.

Thanks again for making and keeping `scoop` running.

PR #1951, opened 2018-01-13
Create sh shim

### LINKED ISSUE #1949: Create Shell Shims
Users of sh-terminals cannot really use Scoop because it is too Windows-specific in the way it operates. There is the [“Can I use Scoop in bash?” Wiki-page](https://github.com/lukesampson/scoop/wiki/Can-I-Use-Scoop-In-Bash%3F) that offers a simple solution for running Scoop command from a sh-terminal, however, any software that is installed via Scoop is not usable within the sh-terminal, well, unless it provides a `.exe`.

The solution to this problem is actually very simple. We could add a shell script to the shims if the software does not come along with an executable. The contents of that one is actually very simple:

```sh
#!/bin/sh
powershell -ex unrestricted scoop.ps1 "$@"
```

Obviously the `scoop.ps1` part would need to be replaced by the actual name of the software. That be all. 😊 

See also #1237

Fleshgrinder:
Closes #1949

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -236,7 +236,7 @@ function shim($path, $global, $name, $arg) {
         "@`"$(resolve-path $path)`" $arg %*" | out-file $shim_cmd -encoding ascii
 
         $shim_sh = "$(strip_ext($shim))"
-        "#!/bin/sh`ncmd `"$(resolve-path $path)`" `"$@`"" | out-file $shim_sh -encoding ascii
+        "#!/bin/sh`ncmd `"$(resolve-path $path)`" $arg `"$@`"" | out-file $shim_sh -encoding ascii
     } elseif($path -match '\.ps1$') {
         # make ps1 accessible from cmd.exe
         $shim_cmd = "$(strip_ext($shim)).cmd"
@@ -253,7 +253,7 @@ if !args! == !invalid! ( set args= )
 powershell -noprofile -ex unrestricted `"& '$(resolve-path $path)' %args%;exit `$lastexitcode`"" | out-file $shim_cmd -encoding ascii
 
         $shim_sh = "$(strip_ext($shim))"
-        "#!/bin/sh`npowershell -ex unrestricted `"$(resolve-path $path)`" `"$@`"" | out-file $shim_sh -encoding ascii
+        "#!/bin/sh`npowershell -ex unrestricted `"$(resolve-path $path)`" $arg `"$@`"" | out-file $shim_sh -encoding ascii
     }
 }
 
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -236,7 +236,7 @@ function shim($path, $global, $name, $arg) {
         "@`"$(resolve-path $path)`" $arg %*" | out-file $shim_cmd -encoding ascii
 
         $shim_sh = "$(strip_ext($shim))"
-        "#!/bin/sh`ncmd `"$(resolve-path $path)`" $arg `"$@`"" | out-file $shim_sh -encoding ascii
+        "#!/bin/sh`ncmd //C `"$(resolve-path $path)`" $arg `"$@`"" | out-file $shim_sh -encoding ascii
     } elseif($path -match '\.ps1$') {
         # make ps1 accessible from cmd.exe
         $shim_cmd = "$(strip_ext($shim)).cmd"
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -234,6 +234,9 @@ function shim($path, $global, $name, $arg) {
         # shim .bat, .cmd so they can be used by programs with no awareness of PSH
         $shim_cmd = "$(strip_ext($shim)).cmd"
         "@`"$(resolve-path $path)`" $arg %*" | out-file $shim_cmd -encoding ascii
+
+        $shim_sh = "$(strip_ext($shim))"
+        "#!/bin/sh`ncmd `"$(resolve-path $path)`" `"$@`"" | out-file $shim_sh -encoding ascii
     } elseif($path -match '\.ps1$') {
         # make ps1 accessible from cmd.exe
         $shim_cmd = "$(strip_ext($shim)).cmd"
@@ -248,11 +251,10 @@ set args=%args:)=``)%
 set invalid=`"='
 if !args! == !invalid! ( set args= )
 powershell -noprofile -ex unrestricted `"& '$(resolve-path $path)' %args%;exit `$lastexitcode`"" | out-file $shim_cmd -encoding ascii
-    }
 
-    # make command accessible from sh terminals
-    $shim_cmd = $(strip_ext($shim))
-    "#!/bin/sh`npowershell -ex unrestricted `"$(resolve-path $path)`" `"$@`"" | out-file $shim_cmd -encoding ascii
+        $shim_sh = "$(strip_ext($shim))"
+        "#!/bin/sh`npowershell -ex unrestricted `"$(resolve-path $path)`" `"$@`"" | out-file $shim_sh -encoding ascii
+    }
 }
 
 function ensure_in_path($dir, $global) {
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -249,6 +249,10 @@ set invalid=`"='
 if !args! == !invalid! ( set args= )
 powershell -noprofile -ex unrestricted `"& '$(resolve-path $path)' %args%;exit `$lastexitcode`"" | out-file $shim_cmd -encoding ascii
     }
+
+    # make command accessible from sh terminals
+    $shim_cmd = $(strip_ext($shim))
+    "#!/bin/sh`npowershell -ex unrestricted `"$(resolve-path $path)`" `"$@`"" | out-file $shim_cmd -encoding ascii
 }
 
 function ensure_in_path($dir, $global) {
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -203,45 +203,46 @@ function shim($path, $global, $name, $arg) {
     $abs_shimdir = ensure (shimdir $global)
     if(!$name) { $name = strip_ext (fname $path) }
 
-    $shim = "$abs_shimdir\$($name.tolower()).ps1"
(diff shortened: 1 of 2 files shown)


lib/install.ps1
```diff
@@ -701,7 +701,7 @@ function rm_shim($name, $shimdir) {
     }
 
     # other shim types might be present
-    '.exe', '.shim', '.cmd' | % {
+    '', ''.exe', '.shim', '.cmd' | % {
```
stkb: Syntax error - there's an extra `'` before the `.exe`

stkb:
I would like to see something like this implemented too, and after correcting the syntax error it works with the (limited) testing I've one. However I think the shellscript shims should only be created if the target is a .ps1 or a .cmd/.bat file. For the .exe's they're just redundant.

Fleshgrinder:
@stkb fixed the syntax error. I agree and it's what I wrote in #1951 too. We could simply `return` from the `exe` block, what do you think?

stkb:
@Fleshgrinder 👍 I would have put the shellscript shim creation in both the .bat and the .ps1 blocks instead, even if it's a bit of repetition, but your suggestion works just as well. I don't really have a preference.

stkb:
Wait sorry I just noticed something, we do need separate shim code for the .bat/.cmd and the .ps1 paths, otherwise we get this for .cmd shims:

```
#!/bin/sh
powershell -ex unrestricted "path/to/gvim.cmd" "$@"
```

Now starting powershell to run a .cmd file does work, but it's not really the best, especially given how slow PS is starting up :)

Just this is apparently enough for .cmd shims

```
#!/bin/sh
"path/to/cmd_or_bat.cmd" "$@"
```

Testing this with Git Bash btw, I don't know if any of this stuff works in WSL.

Fleshgrinder:
Your proposed CMD shim cannot work in WSL because CMD files have no shebang, hence, Linux won't be able to figure out what to start it with. However, we can simply use `cmd "path/to/script.bat" "$@"` in this case which always works. Will update the PR in a bit.

lib/core.ps1
```diff
@@ -234,6 +234,9 @@ function shim($path, $global, $name, $arg) {
         # shim .bat, .cmd so they can be used by programs with no awareness of PSH
         $shim_cmd = "$(strip_ext($shim)).cmd"
         "@`"$(resolve-path $path)`" $arg %*" | out-file $shim_cmd -encoding ascii
+
+        $shim_sh = "$(strip_ext($shim))"
+        "#!/bin/sh`ncmd `"$(resolve-path $path)`" `"$@`"" | out-file $shim_sh -encoding ascii
```
stkb: Shouldn't $arg be in here too?
```"#!/bin/sh`ncmd `"$(resolve-path $path)`" $arg `"$@`""```

File: lib/core.ps1 [hunk 3]
```diff
@@ -248,11 +251,10 @@ set args=%args:)=``)%
 set invalid=`"='
 if !args! == !invalid! ( set args= )
 powershell -noprofile -ex unrestricted `"& '$(resolve-path $path)' %args%;exit `$lastexitcode`"" | out-file $shim_cmd -encoding ascii
-    }
 
-    # make command accessible from sh terminals
-    $shim_cmd = $(strip_ext($shim))
-    "#!/bin/sh`npowershell -ex unrestricted `"$(resolve-path $path)`" `"$@`"" | out-file $shim_cmd -encoding ascii
+        $shim_sh = "$(strip_ext($shim))"
+        "#!/bin/sh`npowershell -ex unrestricted `"$(resolve-path $path)`" `"$@`"" | out-file $shim_sh -encoding ascii
```
stkb: Here too
```"#!/bin/sh`npowershell -ex unrestricted `"$(resolve-path $path)`" $arg `"$@`""```

**Comment by stkb** [comment]
Nice. We just forgot to put in $arg there for the shim args I think (haven't tested my suggestions)

**Comment by Fleshgrinder** [comment]
Only question left is if we have to quote `$arg`? Is this always a single thing or is it actually `$args` and multiple things?

**Comment by stkb** [comment]
It's not a very good name IMO, but it's either a single string or an array of strings.*
They're taken straight from the manifest, where they should already be quoted if necessary I guess. So I don't think it's necessary to quote them. [Existing code doesn't either](https://github.com/lukesampson/scoop/blob/87b2b089336ee4ed690c523e1e0fecb85cabfd74/lib/core.ps1#L236)

\* There's [some code](https://github.com/lukesampson/scoop/blob/87b2b089336ee4ed690c523e1e0fecb85cabfd74/lib/core.ps1#L221) that treats `$arg` as possibly an array, but from some quick testing I'm not sure it can be. Will look into that after this PR.

**Comment by stkb** [comment]
Yeah, `$arg` is always a string (of args joined with spaces), never an array. So it doesn't need to be joined/quoted. The .ps1 shim code that tries to do that leads to a bug. But that's a separate issue.

**Comment by stkb** [comment]
Ok one more thing: I don't know about WSL but just `cmd "path/to/script"` isn't working in Git Bash. It needs the `/C` switch. After changing [line 239](https://github.com/Fleshgrinder/scoop/blob/bf78d6125fdf07cc7d8d5502dd5aa6004c7d4b1d/lib/core.ps1#L239) to this it's all working for me: (note the double `//`)

``` powershell
        "#!/bin/sh`ncmd //C `"$(resolve-path $path)`" $arg `"$@`"" | out-file $shim_sh -encoding ascii
```

Fleshgrinder:
I added the switch for script execution and immediate termination. I guess we're fine now. 👍

stkb:
👍  (cc @r15ch13)

r15ch13:
Some more changes I made:
* remove excessive `strip_ext` calls
* remove excessive `resolve-path` calls
* add `.jar` support (see [selenium](https://github.com/lukesampson/scoop/blob/master/bucket/selenium.json)/[digdag](https://github.com/lukesampson/scoop/blob/master/bucket/digdag.json))

```powershell

function shim($path, $global, $name, $arg) {
    if(!(test-path $path)) { abort "Can't shim '$(fname $path)': couldn't find '$path'." }
    $abs_shimdir = ensure (shimdir $global)
    if(!$name) { $name = strip_ext (fname $path) }

    $shim = "$abs_shimdir\$($name.tolower())"

    # convert to relative path
    pushd $abs_shimdir
    $relative_path = resolve-path -relative $path
    popd
    $resolved_path = resolve-path $path

    # if $path points to another drive resolve-path prepends .\ which could break shims
    if($relative_path -match "^(.\\[\w]:).*$") {
        write-output "`$path = `"$path`"" | out-file "$shim.ps1" -encoding utf8
    } else {
        write-output "`$path = join-path `"`$psscriptroot`" `"$relative_path`"" | out-file "$shim.ps1" -encoding utf8
    }
    if($arg) {
        write-output "`$args = '$($arg -join "', '")', `$args" | out-file "$shim.ps1" -encoding utf8 -append
    }
    if($path -match '\.jar$') {
        "if(`$myinvocation.expectingInput) { `$input | & java -jar `$path @args } else { & java -jar `$path @args }" | out-file "$shim.ps1" -encoding utf8 -append
    } else {
        "if(`$myinvocation.expectingInput) { `$input | & `$path @args } else { & `$path @args }" | out-file "$shim.ps1" -encoding utf8 -append
    }

    if($path -match '\.exe$') {
        # for programs with no awareness of any shell
        cp "$(versiondir 'scoop' 'current')\supporting\shimexe\shim.exe" "$shim.exe" -force
        write-output "path = $resolved_path" | out-file "$shim.shim" -encoding utf8
        if($arg) {
            write-output "args = $arg" | out-file "$shim.shim" -encoding utf8 -append
        }
    } elseif($path -match '\.((bat)|(cmd))$') {
        # shim .bat, .cmd so they can be used by programs with no awareness of PSH
        "@`"$resolved_path`" $arg %*" | out-file "$shim.cmd" -encoding ascii

        "#!/bin/sh`ncmd //C `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
    } elseif($path -match '\.ps1$') {
        # make ps1 accessible from cmd.exe
"@echo off
setlocal enabledelayedexpansion
set args=%*
:: replace problem characters in arguments
set args=%args:`"='%
set args=%args:(=``(%
set args=%args:)=``)%
set invalid=`"='
if !args! == !invalid! ( set args= )
powershell -noprofile -ex unrestricted `"& '$resolved_path' %args%;exit `$lastexitcode`"" | out-file "$shim.cmd" -encoding ascii

        "#!/bin/sh`npowershell -ex unrestricted `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
    } elseif($path -match '\.jar$') {
        "@java -jar `"$resolved_path`" $arg %*" | out-file "$shim.cmd" -encoding ascii
        "#!/bin/sh`njava -jar `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
    }
}
```

r15ch13:
Add `-PathType leaf` to `test-path` because if `$name` and `$_` are empty it will delete everything from the `shims` directory. This can happen when there is no shim name in the manifest (killed my scoop installation like that, which lead to https://github.com/lukesampson/scoop/commit/7176719501ddb20344e0d0d329b9d80d630386a7 😁)
```json
    "bin": [
        ["main.exe", "", "-somearg"]
    ],
```

```powershell
    # other shim types might be present
    '', '.exe', '.shim', '.cmd' | % {
        if(test-path -Path "$shimdir\$name$_" -PathType leaf ) {
            rm "$shimdir\$name$_"
        }
    }
```

Fleshgrinder:
@r15ch13 two questions:

0. Is it a good idea to provide the JAR support with this PR?
0. Why would we wrap the JAR execution in a PowerShell script on SH? We can directly call Java and skip PowerShell.

r15ch13:
@Fleshgrinder yeah, there are some programs that only provide a `*.jar` and everyone of them has a custom `post_install` property that just creates a `ps1` script (see my examples above)
The code above creates a bash script for `.jar` file (see last two lines :smile:)

Fleshgrinder:
Oh, missed the last condition and only looked at the first `jar` check, all good, commit coming up. Many thanks for your great help here.

PR #1958, opened 2018-01-18
upgrade yarn to 1.4.0

nueko:


brandon93s:
1.4.0 is still in pre-release. Once the yarn team marks it stable, it will be autoupdated in scoop:

https://yarnpkg.com/latest-version

PR #1965, opened 2018-01-20
Fix yasm checkver

brandon93s:
Patches the checkver included with #1964.

- Change `https` to `http`
- Fix `re` to grab version

### DIFF
--- a/bucket/yasm.json
+++ b/bucket/yasm.json
@@ -17,8 +17,8 @@
         "yasm.exe"
     ],
     "checkver": {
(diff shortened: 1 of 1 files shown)


Calinou:
The website doesn't have a valid TLS certificate indeed, I was testing it and forgot that my browser automatically redirected me to HTTPS (since I added an exception). So, this should be merged. :+1:

brandon93s:
Awesome, thanks for confirming @Calinou!

PR #1971, opened 2018-01-22
Add dependency walker version 2.2.

matthewjberger:
This adds Dependency Walker version 2.2, which allows you to list the dependencies and imported/exported functions of executables and assemblies.


r15ch13:
Can you add it to the extras bucket?

rasa:
It does have a CLI feature, but it should live in extras

r15ch13:
Closed with: https://github.com/lukesampson/scoop-extras/pull/756

PR #1974, opened 2018-01-23
Revert wget to 1.9.2

rasa:
See
http://git.savannah.gnu.org/cgit/wget.git/commit/?id=ab8a3c80dfc981194ca2ea42d9f709ac198b222c

### DIFF
--- a/bucket/wget.json
+++ b/bucket/wget.json
@@ -1,22 +1,22 @@
 {
     "homepage": "https://eternallybored.org/misc/wget/",
     "license": "GPL3",
-    "version": "1.19.3",
+    "version": "1.19.2",
     "architecture": {
         "64bit": {
             "url": [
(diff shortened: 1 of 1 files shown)


r15ch13:
This will get reverted by the update service to `1.9.3`.

rasa:
I tried `scoop install wget@1.9.2` but it failed, as the file is at https://eternallybored.org/misc/wget/releases/old/wget-1.19.2-win64.zip not https://eternallybored.org/misc/wget/releases/wget-1.19.2-win64.zip as expected by https://github.com/lukesampson/scoop/blob/master/bucket/wget.json

rasa:
Fixed in https://github.com/lukesampson/scoop/commit/85ec2a7b0a020cd764620eea38563c683f14fc56

PR #1980, opened 2018-01-24
Add multi-line support to pre/post_install (fixes #1890)

### LINKED ISSUE #1890: Multiple lines in post_install script
I am trying to put multiple lines in my post install section like below,

"post_install": [
    "Write-Host \"Configuring Git for...\"",
    "git config --global credential.helper wincred",
    "git config --global core.autocrlf true"
]

It turns out an error "% : Cannot convert 'System.Object[]' to the type 'System.String' required by parameter 'Command'. Specified method is not supported
+ $apps | % { install_app $_ $architecture $global $suggested $use_cach ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [ForEach-Object], ParameterBindingException
    + FullyQualifiedErrorId : CannotConvertArgument,Microsoft.PowerShell.Commands.ForEachObjectCommand

Tried with different apps for post_install, but still it is not accepting multiple scripts but if I give only one install script it runs smooth

rasa:


rasa:
@r15ch13 Should I mod the PR to:

````
iex (@($pre_install) -join '`r`n')
...
iex (@($post_install) -join '`r`n')
````

brandon93s:
+1 for `\r\n` - consider the use case of an if statement which wouldn't work with `;`. 

The current implementation is a bit easier to understand to me, but my ps is pretty weak. Checking for an explicit array instead of not string would be even more clear.

stkb:
To explain, the "array subexpression operator" (`@(   )`) does the check for an array already. It's like a "ensure array" operator. If the value passed in to it is already an array, it leaves it as it is. If it's not an array, it creates a singleton array from it. From there it's just a `-join` to ensure we always have a string again.

PR #1986, opened 2018-01-27
Added basic substitutions to bin::args attribute of the manifest

excitoon:


excitoon:
Hi @lukesampson . Can you take a look on my fix? I think it is pretty useful to use substitutions for creating shims. For example:
```
    "bin": [
        [ "usr/bin/sh.exe", "findssl.sh", "$dir/usr/bin/findssl.sh" ]
    ]
```

r15ch13:
Most of these cases can be solved by using the `extract_dir` property (which already supports substitutions).
Do you have an example for which this solution doesn't work?

excitoon:
Yeah. For example we want to install some script (in example above, `findssl.sh`) and run it using some command interpreter (`sh`). `sh` requires full path to the script and we know its path relative to `$dir`. I guess there is nothing common with `extract_dir` (though it was really helpful in the other case).

excitoon:
Hi. Can you give more attention to this ticket? I would like to have this or any other alternative change in main scoop because some of my packages require calling non-executable scripts using external programs.

Please find an example here:
https://github.com/excitoon/scoop-user/blob/master/openssh.json#L44

PR #1989, opened 2018-01-27
suggest cacert after installing openssl

robjarsen:


rasa:
Does openssl need to be told where to find the cert bundle?

robjarsen:
OpenSSL will check the default certificate store on *nix operating systems, but this is not the case for Windows. The user must always specify the path to your certificate bundle using the `CAfile` or `CApath` argument. You can also set the environment variables `SSL_CERT_FILE=/path/to/ca_bundle.crt` or `SSL_CERT_DIR=/path/to/ca/dir` then you do not have to specify CAfile or CApath in every OpenSSL command.

robjarsen:
This is also the case for cURL and wget. You must always specify the path to your CA bundle.

PR #1990, opened 2018-01-28
Fix: 'scoop cache' command not using SCOOP_CACHE

stkb:
I think this was just overlooked? `scoop-cache.ps1` doesn't use the `$cachedir` variable, so it doesn't get the SCOOP_CACHE support added in #1364.

The result being that while the install command correctly uses the folder in SCOOP_CACHE, all *scoop cache ...* commands error because they can't find a `...scoop\cache` folder.


r15ch13:
Good one!

PR #1994, opened 2018-01-29
add YARA

gheoan:
Could not find hash information for the binaries.

### DIFF
--- a/bucket/yara.json
+++ b/bucket/yara.json
@@ -1,14 +1,38 @@
 {
-	"version": "3.7.0",
-	"architecture": {
-		"64bit": {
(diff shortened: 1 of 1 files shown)


r15ch13:
Currently, it's not possible to use our auto-update feature because I can't find a download URL without random strings in it ...
No idea why they don't use GitHub releases for the AppVeyor artifacts, or why they use Dropbox ... 😕

gheoan:
@r15ch13 Thanks for the improvements! What is the reason Dropbox is used instead of AppVoyer in the manifest?

r15ch13:
It's the latest stable version. The files from AppVeyor were build from the latest change.

gheoan:
My understanding is that AppVeyor provides artifacts for every commit. The manifest url could be pointed at the commit which was tagged as latest version.

PR #2003, opened 2018-02-03
 Remove errant quotes around $dir var

rasa:


rasa:
See http://nsis.sourceforge.net/Docs/Chapter3.html#installerusagecommon :
/D sets the default installation directory ($INSTDIR), overriding InstallDir and InstallDirRegKey. It must be the last parameter used in the command line and must not contain any quotes, even if the path contains spaces. Only absolute paths are supported.

PR #2011, opened 2018-02-09
Add bucket_file to installer/uninstaller

comp500:
This allows installers/uninstallers (e.g. powershell scripts) to be stored in the bucket, instead of having to download them twice, or store them in the json file (which VS Code doesn't like).

### DIFF
--- a/bin/uninstall.ps1
+++ b/bin/uninstall.ps1
@@ -23,7 +23,7 @@ function do_uninstall($app, $global) {
     $architecture = $install.architecture
 
     echo "Uninstalling '$app'"
-    run_uninstaller $manifest $architecture $dir
+    run_uninstaller $manifest $architecture $dir $install.bucket
     rm_shims $manifest $global $architecture
 
     # If a junction was used during install, that will have been used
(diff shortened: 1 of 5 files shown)


r15ch13:
Can you give an example how this works or the manifest file should look like? What does vscode not like?

stkb:
> What does vscode not like?

I can answer this. VSCode doesn't like the multiline scripts because they are invalid json. However now that string arrays are supported for `pre_install` and `post_install`, they can be converted to string arrays instead :) Still not ideal for very long scripts though.

comp500:
As stkb said, VSCode doesn't like them as they are invalid json, as control characters must be escaped. Putting them in a seperate file also allows for syntax highlighting, code suggestions etc. that the VSCode powershell extension provides.

I have implemented and tested it in my personal bucket, [commit here](https://github.com/comp500/scoop-comp500/commit/6150825d63c3bfff9e3628f671699c89ff267bf0), with the [installers in the bucket](https://github.com/comp500/scoop-comp500/tree/master/installers).

rashil2000:
No longer required. The current way of doing this is calling `$bucketsdir\bucket\scripts\app\installer.ps1`

PR #2012, opened 2018-02-09
Vim: Create gvim.cmd file for gvim shim (fixes #2006)

### LINKED ISSUE #2006: windows cmd gvim.exe shim waits for subprocess
I am using scoop in a windows cmd (perhaps why i am the only person seeing this). The `gvim.exe` shim waits for the subprocess to terminate, and this is not desirable. The `gvim.ps1` shim works just fine: running `gvim` opens a new windows and a new prompt is available.

I'm not sure how the `gvim.exe` shim is created but it would be nice if it did something equivalent to `start / b ..\path\to\current\gvim.exe $*`.

I'm happy to provide any additional information, though I'm not sure what may be needed or how to get it ;-). I'm sure there is a way to make `.ps1` files executable from windows cmd, but regardless `gvim.exe` from powershell exhibits the same undesired behavior.

stkb:
See #2006 for discussion. Creates a gvim.cmd file in pre_install

For some reason, this didn't work (wouldn't create the .cmd file) when I first added it to post_install instead.

This might also need adding to the extras/gvim package (I don't actually know what that package is for? It seems almost identical to the vim package)

**Edit**: I should emphasize that, this breaks the `gvim` shim for those using (Git) Bash, until something like #1951 is implemented.

### DIFF
--- a/bucket/vim.json
+++ b/bucket/vim.json
@@ -19,7 +19,46 @@
             "vim.exe",
             "vi"
         ],
-        "gvim.exe"
+        [
+            "gvim.cmd",
+            "gvim"
+        ]
+    ],
+    "pre_install": [
+        "# Create gvim.cmd",
+        "@\"",
+        "@echo off",
+        "setlocal",
+        "set VIM_EXE_DIR=$dir",
+        "if exist \"%VIM_EXE_DIR%\\gvim.exe\" goto havevim",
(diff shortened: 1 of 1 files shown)


rasa:
I shouldn’t review as I’m not a vim expert.

rasa:

PR #2019, opened 2018-02-13
Update latex to version 2.9.6615

brandon93s:
- Update to version 2.9.6615
- Extract SHA-256 from website; 190 mb download to compute hash!


r15ch13:
Ah, salvation for my server! 😁

PR #2029, opened 2018-02-19
Proposal: Use write-output in scoop list.

bannmoore:
This is a minor quality of life thing, but currently `scoop list` has no output and cannot be piped. Coming from `homebrew`, I expect this to work:

```
scoop list | grep "putty"
```

This is useful for workstation setup scripts that try to check if something is already installed. Listing the apps with `write-output` instead of `write-host` means losing the colored text, but that seems like a fair tradeoff.

### DIFF
--- a/libexec/scoop-list.ps1
+++ b/libexec/scoop-list.ps1
@@ -17,29 +17,29 @@ $global = installed_apps $true | % { @{ name = $_; global = $true } }
 $apps = @($local) + @($global)
 
 if($apps) {
-    write-host "Installed apps$(if($query) { `" matching '$query'`"}): `n"
+    write-output ""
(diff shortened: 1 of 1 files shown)


rasa:
Maybe `scoop list` is colored, but `scoop export` is not?

bannmoore:
I'll admit, I did not realize that was what `scoop export ` was doing. My bad! Reading is hard.

That being said, I'm not sure having both is necessary, for a couple of reasons:
- I can't think of any instance where printing it with colors is more useful than just printing a plain output (but I understand that's just personal preference)
- It deviates from the `grep`-able `<command> list` convention in the tools this one is based on.

Maybe it could be a flag instead of two separate commands? `scoop list` and `scoop list --pretty`, or some such. Regardless, `scoop export` suits my needs so I'll close this.

PR #2032, opened 2018-02-20
Clean up some error messages

stkb:
Cleans up code that outputs error messages:
- Removes the extra 'ERROR' text in the case of `error 'ERROR...`
- Uses the `error` function instead in some cases of `write-host '...' -f Red/DarkRed`...
- ...Except when an `exit` comes after it, in which case it's changed to use `abort` (in some cases this changes the text from darkred to red). The `abort` method doesn't prepend any text.

Haven't touched these cases:
- Code that (effectively) uses write-output instead of write-host to output error messages.
- Code that uses `[console]::error.writeline` instead, since I don't know exactly what this does/why it was used
- Lots of code uses `write-host -f DarkRed "..."`, but I haven't changed them all to use the error function because I don't know if preceding every error message with "ERROR " was really desired.

----

Additional note: For the scoop commands there's some inconsistency in what they do when arguments are missing. Some use `error` (in turn using `write-host`), eg:
https://github.com/lukesampson/scoop/blob/35eb73b2dbc4a8cde6189c8b46eccf03113b9f3b/libexec/scoop-install.ps1#L62

Others just output the text, eg:
https://github.com/lukesampson/scoop/blob/35eb73b2dbc4a8cde6189c8b46eccf03113b9f3b/libexec/scoop-cache.ps1#L41

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -65,7 +65,7 @@ $queue | % {
 
     if ($json.checkver -eq "github") {
         if (!$json.homepage.StartsWith("https://github.com/")) {
-            write-host "ERROR: $name checkver expects the homepage to be a github repository" -f DarkYellow
+            error "$name checkver expects the homepage to be a github repository"
         }
         $url = $json.homepage + "/releases/latest"
         $regex = $githubRegex
--- a/libexec/scoop-depends.ps1
+++ b/libexec/scoop-depends.ps1
@@ -15,13 +15,13 @@ reset_aliases
 $opt, $apps, $err = getopt $args 'a:' 'arch='
 $app = $apps[0]
 
-if(!$app) { error 'ERROR: <app> missing'; my_usage; exit 1 }
+if(!$app) { error '<app> missing'; my_usage; exit 1 }
 
 $architecture = default_architecture
 try {
     $architecture = ensure_architecture ($opt.a + $opt.arch)
 } catch {
-    error "ERROR: $_"; exit 1
+    abort "ERROR: $_"
 }
 
 $deps = @(deps $app $architecture)
--- a/bin/auto-pr.ps1
+++ b/bin/auto-pr.ps1
@@ -53,8 +53,7 @@ if ((!$push -and !$request) -or $help) {
 }
 
 if(!($upstream -match "^(.*)\/(.*):(.*)$")) {
-    Write-Host -f DarkRed "Upstream must have this format: <user>/<repo>:<branch>"
-    exit 1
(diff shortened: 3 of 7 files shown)


bin/checkver.ps1
```diff
@@ -65,7 +65,7 @@ $queue | % {
 
     if ($json.checkver -eq "github") {
         if (!$json.homepage.StartsWith("https://github.com/")) {
-            write-host "ERROR: $name checkver expects the homepage to be a github repository" -f DarkYellow
+            error "$name checkver expects the homepage to be a github repository"
```
stkb: Even though the original was darkyellow text, changed to `error` instead of `warn`, since it seemed like an error condition.

r15ch13:
Regarding the `scoop-cache.ps1` it's clearly an error message so adding the `error` function is reasonable.
`auto-pr.ps1` and `checkver.ps1` are mainly used for maintenance, so they aren't that important :grin:

PS: there goes my plan to remove the `abort` function :sob::grin:

stkb:
> PS: there goes my plan to remove the abort function 😭😁

If you want to go another way, it's fine by me :)

r15ch13:
The complete error handling of scoop has to be changed in order to remove the `abort` function. So it's still okay to use it :smile:

libexec/scoop-depends.ps1
```diff
@@ -15,13 +15,13 @@ reset_aliases
...
 
-if(!$app) { error 'ERROR: <app> missing'; my_usage; exit 1 }
+if(!$app) { error '<app> missing'; my_usage; exit 1 }
 
 $architecture = default_architecture
 try {
     $architecture = ensure_architecture ($opt.a + $opt.arch)
 } catch {
-    error "ERROR: $_"; exit 1
+    abort "ERROR: $_"
```
rasa: This is fine, but kinda confusing. Could abort either not prefix 'ERROR: ', or use a different prefix, like 'FATAL: '?

stkb: It could, but I wouldn't worry too much about it. As @r15ch13 said, the whole error handling needs to be changed anyway. The worst cases ("ERROR ERROR...") have been taken care of.

rasa: Agreed. thx.

stkb:
Having thought about it some more, yeah this should all really be done a different way.

IMO most inner functions shouldn't be writing error messages, or possibly anything to the screen directly, but rather returning a success value or throwing errors if something goes wrong. Only at the last moment should the result be translated into success or error messages on-screen. This makes it _much_ easier to test and to know what's going on in the code.

PR #2033, opened 2018-02-20
Warn on shim overwrite

rasa:
Tested and working for me

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -198,12 +198,44 @@ function movedir($from, $to) {
     }
 }
 
+function get_app_name($path) {
+    if ($path -match '([^/\\]+)[/\\]current[/\\]') {
+        return $Matches[1].tolower()
+    }
+    return ""
+}
+
+$warned_on_overwrite = @{}
+
+function warn_on_overwrite($shim_ps1, $path) {
+    if (!([System.IO.File]::Exists($shim_ps1))) {
+        return
+    }
+    $shim = [System.IO.Path]::GetFileNameWithoutExtension($shim_ps1).tolower()
+    if ($warned_on_overwrite.ContainsKey($shim)) {
+        return
+    }
+    $reader = [System.IO.File]::OpenText($shim_ps1)
+    $line = $reader.ReadLine().replace("`r","").replace("`n","")
+    $reader.Close()
+    $shim_app = get_app_name $line
+    $path_app = get_app_name $path
+    if ($shim_app -eq $path_app) {
+        return
+    }
+    $filename = [System.IO.Path]::GetFileName($path)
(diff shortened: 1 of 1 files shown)


lib/core.ps1
```diff
@@ -198,13 +198,45 @@ function movedir($from, $to) {
...
+    $line = $reader.ReadLine().replace("`r","").replace("`n","")
+    $reader.Close()
+    $shim_app = get_app_name $line
+    $path_app = get_app_name $path
+    if ($shim_app -eq $path_app) {
+        return
+    }
+    $filename = [System.IO.Path]::GetFileName($path)
+    warn "Overwriting shim to $filename installed from $shim_app"
+    $warned_on_overwrite[$shim] = $shim_ps1
```
stkb: Not sure what the warning flags were meant to do, but at the moment, since this function is only called once for any given shim name, it's not doing anything.

rasa: I appreciate the feedback. The warming’s are to inform the user when a program overwrites another program’s shim, as the two have the same name. My apologies if that wasn’t clear. I’m not sure what you mean by “it’s not doing anything,” I’ve tested it, and it’s reports when a shim is overwritten. Are you suggesting it report if the overwritten shim is then overwritten by a subsequent app?

stkb: Sorry, I myself wasn't clear. I'm not talking about the warning feature itself (which works and is a good feature!)

I'm talking about where you set up a hashtable:
``` powershell
$warned_on_overwrite = @{}
```

And later check if it already contains a shim:
``` powershell
if ($warned_on_overwrite.ContainsKey($shim)) {
        return
}
```

And later set a flag for this shim:
``` powershell
$warned_on_overwrite[$shim] = $shim_ps1
```

This is the code that, as far as I can tell, isn't doing anything, since the function is only run once for any `$shim`. `$warned_on_overwrite.ContainsKey($shim)` will always be false.* What is this code supposed to be preventing?

\*Technically a manifest could contain two shims with the same name; there's no code preventing it. But that's a different problem that should be handled elsewehere.

stkb: I've just realized, that you might be thinking that the `shim` function gets called once for each individual shim file (.exe, .ps1, .cmd...)? If so then the code would make sense. But it isn't: it's only called once for each named shim (eg `shim1`) and all files (`shim1.ps1`, `shim1.exe` ...) are created in the one method call.

rasa: You're right, that's leftover code that should be stripped out. Thanks for catching it. I will update the PR.

rasa: OK, I removed the dead code. It should be good to go now. Thanks again.

PR #2034, opened 2018-02-21
Adapt to path change of Newtonsoft.Json.dll from #2024 (see f0a15886)

pcrama:
The move of the Newtonsoft.Json.dll breaks the virustotal subcommand.  This PR fixes it.

Other similar PRs in the buckets `versions' and `extras' are forthcoming.

Thanks in advance,
-- 
Philippe Crama


r15ch13:
Oh shit, forgot about the tests... Thanks for fixing it :+1:

PR #2053, opened 2018-02-27
Enabled applications which require elevated privileges.

excitoon:
Hi.

Sometimes shim does not work because `CreateProcess()` returns `ERROR_ELEVATION_REQUIRED` error (e.g., `scoop install rufus && rufus`). We can't completely replace code with `ShellExecute()` because it would break running programs in the same console window, that is why I first check for `ERROR_ELEVATION_REQUIRED`, and then use `ShellExecute()` (in .NET way).

I guess, this really fixes #1319 (#579) for UAC-enabled systems.


excitoon:
Hi. @lukesampson . Can you take a look at it?

excitoon:
Rebased.

r15ch13:
Quick to make this work, execute `scoop reset rufus` or `scoop reset *`

PR #2054, opened 2018-02-27
Converted llvm download to 7z to avoid elevated permissions request

SimplyKnownAsG:
The installer and uninstaller do not appear to have `/noadmin` options. converting to a 7z and adding bin appears to work. I'm not sure what the installer was attempting to do, so I believe this is an acceptable change.


SimplyKnownAsG:
@r15ch13, is there something I should do to get this reviewed? Thanks!

PR #2056, opened 2018-02-28
added descritpion field

gbursson:
while browsing thru scoop buckets I sorely miss description on what is the package supposed to do, so I will try to provide description tags for those

### DIFF
--- a/bucket/minisign.json
+++ b/bucket/minisign.json
@@ -1,5 +1,6 @@
 {
     "homepage": "https://jedisct1.github.io/minisign/",
+    "description": "A dead simple tool to sign files and verify signatures",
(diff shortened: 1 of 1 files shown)


r15ch13:
The `description` property currently isn't used anywhere. So it's only available by looking at the manifest files.

One pull-request for all the changes would have been enough :grin:

gbursson:
that's exactly where I look for a quick glance what is the package about.
forgive me numerous pull requests, I am brand new to the whole github thing

rasa:
Once most of the apps have descriptions, we can run [makeindex.py](https://github.com/rasa/scoops/blob/master/makeindex.py) on them, and generate a summary table, as is shown here: https://github.com/rasa/scoops#apps

PR #2065, opened 2018-02-28
Support TLS 1.2 for scoop search

### LINKED ISSUE #2063: Invoke-WebRequest Could not create SSL/TLS secure channel
On my machine `scoop search` fails with error:

```
invoke-webrequest : The request was aborted: Could not create SSL/TLS secure channel.
At C:\Users\sekost\scoop\apps\scoop\current\libexec\scoop-search.ps1:53 char:15
+     $result = invoke-webrequest $url -UseBasicParsing | select -exp c ...
```

It fails to access GitHub API endpoints.

I guess it can be because of this:
[StackOverflow: Powershell Invoke-WebRequest Fails with SSL/TLS Secure Channel](https://stackoverflow.com/questions/41618766/powershell-invoke-webrequest-fails-with-ssl-tls-secure-channel)

**UPDATE:**

Yes, I added this line into `scoop-seach.ps1` and it solved the problem:
```ps1
[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
```

kostrse:
`scoop-search` fails with the error below when trying to access GitHub API endpoint.

```
invoke-webrequest : The request was aborted: Could not create SSL/TLS secure channel.
At C:\Users\sekost\scoop\apps\scoop\current\libexec\scoop-search.ps1:53 char:15
+     $result = invoke-webrequest $url -UseBasicParsing | select -exp c ...
```

This fix tells .NET runtime to use TLS 1.2 security protocol for HTTPS.
Fixes the problem on Windows 10 1709, PowerShell 5.1

Not sure if it introduces any compatibility issues on older versions of Windows / PowerShell.

**Possibly this fix needs to be placed in another file.**

Fixes #2063

### DIFF
--- a/libexec/scoop-search.ps1
+++ b/libexec/scoop-search.ps1
@@ -113,6 +113,8 @@ function search_remotes($query) {
     }
 }
 
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
 if (!$local_results -and !(github_ratelimit_reached)) {
     $remote_results = search_remotes $query
     if(!$remote_results) { [console]::error.writeline("No matches found."); exit 1 }


r15ch13:
Probably related to https://github.com/lukesampson/scoop/issues/2040, we need to work on a global fix for that.

masaeedu:
I think the correct fix is to set:

```
[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
```

since this avoids mysteriously breaking your own access to sites that do not yet support TLS 1.2.

@r15ch13 could we just put this in `core.ps1`? There's not a lot of places where you're doing `net.webclient` or `invoke-webrequest`, and all the ones I can find have an import of `core.ps1`.

r15ch13:
superseded by https://github.com/lukesampson/scoop/pull/2074

PR #2072, opened 2018-03-01
Enabled shimming of external applications

excitoon:
Hi.

Now one can make shims for external applications (system, other applications installed by scoop and so on).

First, `$bin` is to be searched in `$dir`, then globally, then in user and system `PATH`.

Examples:
```
    "bin": [ [ "winver.exe", "myver", "" ] ]
```

```
    "depends": [ "python" ],
    "bin": [
        [ "python.exe", "myscript", "$dir/myscript.py" ],
        [ "python.exe", "myscript.py", "$dir/myscript.py" ]
    ]
```

```
    "depends": [ "bash" ],
    "bin": [
        [ "bash.exe", "myscript", "$dir/myscript.sh" ],
        [ "bash.exe", "myscript.sh", "$dir/myscript.sh" ]
    ]
```

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -679,14 +679,14 @@ function create_shims($manifest, $dir, $global, $arch) {
         $target, $name, $arg = shim_def $_
         write-output "Creating shim for '$name'."
 
-        if(test-path "$dir\$target") {
+        if(test-path "$dir\$target" -pathType leaf) {
             $bin = "$dir\$target"
-        } elseif(test-path $target) {
+        } elseif(test-path $target -pathType leaf) {
             $bin = $target
         } else {
             $bin = search_in_path $target
         }
-        if(!(test-path $bin)) { abort "Can't shim '$target': File doesn't exist."}
+        if(!$bin) { abort "Can't shim '$target': File doesn't exist."}
 
         shim $bin $global $name (substitute $arg @{ '$dir' = $dir; '$original_dir' = $original_dir; '$persist_dir' = $persist_dir})
     }
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -679,14 +679,16 @@ function create_shims($manifest, $dir, $global, $arch) {
         $target, $name, $arg = shim_def $_
         write-output "Creating shim for '$name'."
 
-        # check valid bin
-        $bin = "$dir\$target"
-        if(!(is_in_dir $dir $bin)) {
-            abort "Error in manifest: bin '$target' is outside the app directory."
+        if(test-path "$dir\$target") {
+            $bin = "$dir\$target"
+        } elseif(test-path $target) {
+            $bin = $target
+        } else {
+            $bin = search_in_path $target
         }
         if(!(test-path $bin)) { abort "Can't shim '$target': File doesn't exist."}
 
-        shim "$dir\$target" $global $name (substitute $arg @{ '$dir' = $dir; '$original_dir' = $original_dir; '$persist_dir' = $persist_dir})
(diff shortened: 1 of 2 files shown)


r15ch13:
Do you have an example for which this could be useful?

excitoon:
It could be useful for virtually any case where script is being shipped without interpreter. For example, `openssh` ships with few Bash scripts (`findssl.sh` and `ssh-copy-id`) and it forces maintainer to include `bash` and its dependencies into manifest (and to file system of every user).
More examples (Python, Perl, Bash):
- getmail
- SimpleHTTPServer (one would shim it without any real files, just Python 2)
- deluge
- newspipe
- [https://github.com/jpoliv/wakeonlan](wakeonlan)
- [https://github.com/jpoliv/wakeonlan](swaddle)

excitoon:
Basically my point is that every programmer who writes scripts and want to ship it to Windows systems need to make some `cmd` wrapper around them in order to run them from command line. This patch will allow run these scripts in Windows in the same way as in Unix.

gbursson:
excitoon, just to let you know, two links (for wakeonlan and swaddle) are incorrect

stkb:
I've also found this patch useful for a tool I've been using (which is actually a php script).

lib/install.ps1
```diff
@@ -679,14 +679,16 @@ function create_shims($manifest, $dir, $global, $arch) {
         $target, $name, $arg = shim_def $_
         write-output "Creating shim for '$name'."
 
-        # check valid bin
-        $bin = "$dir\$target"
-        if(!(is_in_dir $dir $bin)) {
-            abort "Error in manifest: bin '$target' is outside the app directory."
+        if(test-path "$dir\$target") {
```
stkb: `test-path` should probably have `-pathType leaf` to be safe

lib/install.ps1
```diff
@@ -679,14 +679,16 @@ function create_shims($manifest, $dir, $global, $arch) {
...
+            $bin = "$dir\$target"
+        } elseif(test-path $target) {
```
stkb: Not sure what this check is for (if $target is absolute?)

lib/install.ps1
```diff
@@ -679,14 +679,16 @@ function create_shims($manifest, $dir, $global, $arch) {
...
+            $bin = $target
+        } else {
+            $bin = search_in_path $target
         }
         if(!(test-path $bin)) { abort "Can't shim '$target': File doesn't exist."}
```
stkb: This needs to be something like `if(!$bin) { ... }` instead. Currently, if we don't find the file, `$bin` is null and this check blows up, causing lots of errors as the null value propogates. 
Also there's no need to do the `test-path` again.

lib/install.ps1
excitoon: Yeah, just because absolute path is better than invalid path.

lib/install.ps1
excitoon: Yeah, you're right.

lib/install.ps1
excitoon: Thanks. Actually I did not use Powershell before.

excitoon:
Rebased first commit on current master.

excitoon:
One more thing about `For-Each` and piping:
```
    $path.split(';') | % {
        if(test-path "$_\$target" -pathType leaf) {
            return "$_\$target"
        }
    }
```

If you write it like that, function may return a list consisting of joined return values from each element. That's totally counter-intuitive. If one replace that with regular `foreach` everything works as expected.

excitoon:
I guess it is ready to merge/review again. Tested on all mentioned cases.

excitoon:
Hi @r15ch13. Can you merge it (and the other 3 pull requests: #2079, #2118, #2120)?

PR #2075, opened 2018-03-02
Lint: PSAvoidUsingCmdletAliases

h404bi:
Replace all cmdlet aliases with their full content.

### DIFF
--- a/bin/auto-pr.ps1
+++ b/bin/auto-pr.ps1
@@ -59,7 +59,7 @@ if(!($upstream -match "^(.*)\/(.*):(.*)$")) {
 
 function execute($cmd) {
     Write-Host -f Green $cmd
-    $output = iex $cmd
+    $output = Invoke-Expression $cmd
 
     if($LASTEXITCODE -gt 0) {
         Write-Host -f Red "^^^ Error! See above ^^^ (last command: $cmd)"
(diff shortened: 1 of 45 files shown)


r15ch13:
Nice! Is there a way to enforce it by adding it to the tests?

h404bi:
@r15ch13 
I think we should lint the whole project for future mantainance, `PSAvoidUsingCmdletAliases` is the first step of this big task.

The official PowerShell script *linter* is the [PSScriptAnalyzer](https://github.com/PowerShell/PSScriptAnalyzer), currently we could use it through:

1. Install [From PowerShell Gallery](https://github.com/PowerShell/PSScriptAnalyzer#from-powershell-gallery), then use it through console.
2. Install [VSCode PowerShell extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.PowerShell), then use it in VSCode.

But the PSScriptAnalyzer is still problematic, some rules are false positive. e.g. [PSUseDeclaredVarsMoreThanAssignment](https://github.com/PowerShell/PSScriptAnalyzer/issues/903).

It seems that PowerShell script linting toolchain remain to be perfected, but of course, the automatic linting and tests should be improved. Here I just give a prompt to set the ball rolling. 😂

rasa:
These changes make the code much easier to read and maintain. Can we add a lint check to our test suite, and/or build chain?

bergmeister:
Just a heads up: The next upcoming release of PSScriptAnalyzer will have a `-Fix` switch that can autofix you warnings like e.g. `PSAvoidUsingCmdletAliases`
We are aware that `PSUseDeclaredVarsMoreThanAssignment` has many false positives. Some enhancements will come in the next version but this rule is a tricky one (even from a theoretical point of view, it is difficult to assess when a variable is being used and not in terms of scoping). But we are open to PRs :-)

PR #2079, opened 2018-03-02
Better error handling during uninstallation.

excitoon:
If one uninstalls more than one application and have an error, she will receive pretty random behaviour - scoop will either stops (skipping other applications) or continue to uninstall the current. This is so because `continue` doesn't works with pipes and some `continue` were supposed to exit current iteration of nested loop. When one calls `continue` from `For-Each`, PowerShell simply exits `For-Each` (the same effect as if it was `break`:
```
PS C:\Users\Chebotarev_V\scoop\apps\scoop\current> @(1,2,3,4) | % { echo $_; continue }
1
PS C:\Users\Chebotarev_V\scoop\apps\scoop\current> foreach($_ in @(1,2,3,4)) { echo $_; continue; }
1
2
3
4
```

### DIFF
--- a/libexec/scoop-uninstall.ps1
+++ b/libexec/scoop-uninstall.ps1
@@ -36,7 +36,7 @@ if($apps -eq 'scoop') {
 $apps = ensure_all_installed $apps $global
 if(!$apps) { exit 0 }
 
-$apps | ForEach-Object {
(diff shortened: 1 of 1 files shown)


excitoon:
Done.

PR #2087, opened 2018-03-05
Fix hash in exiftool

rasa:


rasa:
We *HAVE* to move of sourceforget.

PR #2095, opened 2018-03-05
Fix llvm download rename to 7z

SimplyKnownAsG:
The llvm download rename wasn't quite correct. This fixes it, and has been tested. 

Is there a post-install test that should be run to confirm this works?

### DIFF
--- a/bucket/llvm.json
+++ b/bucket/llvm.json
@@ -4,11 +4,11 @@
     "license": "University of Illinois/NCSA Open Source License",
     "architecture": {
         "32bit": {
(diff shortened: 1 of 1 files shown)


SimplyKnownAsG:
@r15ch13, unfortunately it looks like something whacky had happened with #2054. It is a bit difficult to check the scoop/bucket/*.json definitions without breaking my currently install. would be nice to minimize what is in scoop/bucket/*.json to the very minimum requirements for getting scoop operational (7z, git?), and then maybe having a bucket for scoop-core or something

PR #2103, opened 2018-03-08
Fixing OpenSSH SourceForge Address!

yigitemres:
Fixing OpenSSH SourceForge Address! #2064

PR #2104, opened 2018-03-08
Update openssh to 7.6p1-2 (fixes #2103)

rasa:
Step 1 of 60 in moving off of sourceforget.

## scoop (msys related):
````
coreutils
(diffutils has already been migrated to msys2)
findutils
gawk
gzip
msys
msys2
patch
tar
telnet
````
## scoop (all sourceforget hosted):
````
atomicparsley
bochs
cdrtools
core
coreutils
ctags
dosbox
editorconfig
exiftool
file
findutils
gawk
gcc
gdb
grep
gzip
iconv
inadyn-mt
innounp
install
msys
msys2
optipng
patch
pngcrush
premake4
qpdf
sbcl
sed
smartmontools
tar
telnet
vim
which
xming
xmlstarlet
````
## scoop-extras:
```
console2
ditto
doublecmd
filezilla
gvim
joe
jpegview
kdiff3
keepass
mc
mplayer
nsis
pidgin
plantuml
processhacker
qbittorrent
sfk
SweetHome3D
tortoisemerge
tortoisesvn
util-linux-ng
vcxsrv
winmerge
winscp
````

### DIFF
--- a/bucket/openssh.json
+++ b/bucket/openssh.json
@@ -1,30 +1,80 @@
 {
-    "version": "5.4p1-1",
+    "homepage": "https://www.openssh.com/",
+    "version": "7.6p1-2",
     "license": "https://www.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/LICENCE?rev=HEAD",
-    "url": [
-        "https://sourceforge.net/projects/mingw/files/MSYS/Extension/openssh/openssh-5.4p1-1/openssh-5.4p1-1-msys-1.0.13-bin.tar.lzma",
-        "https://sourceforge.net/projects/mingw/files/MSYS/Extension/openssh/openssh-5.4p1-1/openssh-5.4p1-1-msys-1.0.13-lic.tar.lzma",
-        "https://sourceforge.net/projects/mingw/files/MSYS/Base/msys-core/msys-1.0.13-2/msysCORE-1.0.13-2-msys-1.0.13-bin.tar.lzma",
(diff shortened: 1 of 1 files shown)


r15ch13:
Please don't create 60 single pull requests :grin:
Updating this one is enough :smile:

rasa:
I promise this is the only one. If we want to be able to install _any_ msys2 package, perhaps we create a separate `msys` bucket, and have a script create and update the packages in it. Shouldn't take more than a few hours. Not high on my priority list, though.

excitoon:
Can someone explain what exactly was wrong with SourceForge?
I guess the problem was in deprecated URL format, which worked for a while and then stopped.
Look at #2103, from my perspective that is better solution.

rasa:
@excitoon: Sorry, I was (semi) jesting about moving all 60 apps off sourceforge. We’ve been having issues with them, so I thought it helpful to document what apps are affected when sf.net goes down.

Ignoring my joke, I think upgrading openssh from 5.4 to 6.7 is a good idea, to capture all security fixes. Also, upgrading openssh to the version that git-for-windows uses internally will minimize conflicts git might have with version 5.4.

rasa:
https://github.com/excitoon/scoop-user/blob/master/openssh.json may be a better solution. I’m gonna close this for now.

excitoon:
> Ignoring my joke, I think upgrading openssh from 5.4 to 6.7 is a good idea, to capture all security fixes. Also, upgrading openssh to the version that git-for-windows uses internally will minimize conflicts git might have with version 5.4.

That's it for sure. Actually I wrote my own manifest in order to connect to Git-TFS who did not accept old 5.4 authentication mechanisms.

Issue #1863, opened 2017-11-23
query: scoop export


jas99:
I see one can use `scoop export` to export the state; but is thr any way to import exported file.
Similar to `brew bundle dump` which produces Brewfile that can be imported.

Thx

r15ch13:
No there is no way to import it right now. You have to create your own script to install everything from the exported list.

rasa:
Closing as a duplicate of #2105.

PR #2105, opened 2018-03-08
Added basic importing function for exported lists!

yigitemres:
Added basic importing function for exported lists! #1863

### DIFF
--- a/libexec/scoop-import.ps1
+++ b/libexec/scoop-import.ps1
@@ -1,0 +1,38 @@
+# Usage: scoop import <filepath/filepaths> [options]
+# Summary: Import exported list!
+# Help: The basic way to install exported list:
+#      scoop import exportedfile.txt
+
+
+function createAppsList($filePaths){
+    if($filePaths){
+        foreach ($file in $filePaths){
+	          foreach ($line in Get-Content $file)
+	          {
+                if($line -ne "") #Checks empty line!
+                {
+                    $appslist += $line.Split()[0] + " "
+                }
+	          }
+        }
+    }else{
+	      foreach ($line in Get-Content $filePaths)
+	      {
+            if($line -ne "")
+            {
+                $appslist += $line.Split()[0] + " "
+            }
+	      }
+    }
+	return $appslist
+}
+
+function importApps($appslist)
+{
+    scoop install $appslist
+}
+
+
+$filePaths = $args
+importApps(createAppsList($filepaths))
+


libexec/scoop-import.ps1
```diff
@@ -0,0 +1,44 @@
...
+    if($filePaths){
+        foreach ($file in $filePaths){
+            $ext = [IO.Path]::GetExtension($file)
+            if($ext -eq ".txt")
+            {
+                foreach ($line in Get-Content $file)
+                {
+                    if($line -ne "") #Checks empty line!
+                    {
+                        $appslist += $line.Split()[0] + " "
```
excitoon: What if we have duplicate packages on the list or dependency of some package and the package itself?

libexec/scoop-import.ps1
```diff
@@ -0,0 +1,44 @@
...
+function importApps($appslist)
+{
+    $appslist.Split() | ForEach-Object{
+        scoop install $_
+    }
+}
+
+
+$filePaths = $args
+importApps(createAppsList($filepaths))
```
excitoon: I think we should report `%ERRORLEVEL%` with result of the operation.

libexec/scoop-import.ps1
yigitemres: Duplicate packages are also going to into list but, because of the nature ``scoop install`` they are not installed. But you are right, duplicates are trying to install by this script. I will look into it. Thx for your comments! 🥂

libexec/scoop-import.ps1
yigitemres: I will add this too. Thx again! 👍

rasa:
Can import ignore lines that start with a # or ; char?

yigitemres:
@rasa Nope currently, but thx to you it will. 🥂

rasa:
Since `scoop export` says it [`Exports (an importable) list of installed apps`](https://github.com/lukesampson/scoop/blob/master/libexec/scoop-export.ps1#L2), `scoop import` should be able to consume its output, and act accordingly. To that end, we may want to consider that `scoop import` should:
1. Install globally, if [`*global*`](https://github.com/lukesampson/scoop/blob/master/libexec/scoop-export.ps1#L26) is found.
2. Install the 32bit version, if [`{32bit}`](https://github.com/lukesampson/scoop/blob/master/libexec/scoop-export.ps1#L37) is found.
3. Add the appropriate bucket, if [`[bucket]`](https://github.com/lukesampson/scoop/blob/master/libexec/scoop-export.ps1#L32) is found. For non-standard buckets, this should fail (until we add the bucket location to the export file).
4. Install “url” apps, if a [`[url]`](https://github.com/lukesampson/scoop/blob/master/libexec/scoop-export.ps1#L34) app is found.
5. Consider supporting the alternate bucket convention `bucket/appname`.
6. Consider supporting pinning of specific versions, using `appname@version`.
7. Consider maybe adding flags to quit on the first failure, or ignore all failures and plow on.
8. Gracefully ignore apps that are already installed.
9. Consider adding switches to either report/fail, or uninstall/reinstall, if the installed app is not the same as the one listed in the import file (different bucket/url, architecture, or globalness ).

Have I captured everything? @r15ch13?

yigitemres:
@rasa Thx for suggestions. 1 and 2 is done!  As I said earlier, I'm pretty new to PS scripting but I try my best. 🥂

r15ch13:
@rasa sounds about right.
6. export has to be modified so it will output this info
7. "plowing" through might be difficult because of the bad error handling 😢
8. `scoop install` does this already

@yigitemres I don't want to be rude, but this https://blog.codinghorror.com/flattening-arrow-code/ might be a good read for you. 😃

yigitemres:
@r15ch13 It could be worse. It could be spaghetti with switch cases and arrow codes! 🤣 My first ever C++ homework is resulted with spaghetti code and I had to rewrite it just day before the deadline!

I just write this sake of QoL. It will be possibly deleted or changed (if it is ever merged!) in time. I maybe calling these ifs as function (or found clever way to check string for multiple case with ease) in order to become a little more easy to read.

Any fix or suggestions are welcomed as always!
(BTW, If someone had already create import function, please pull request. I will be happily delete this. All I want to achieve is that eliminating issues and contributing this wonderful project.)

yigitemres:
@rasa * for wildcards. We are both wrong. 😆 It should be `*{32bit}*`. (When apps name like "dasapp32bit", it was passed as --arch 32bit)

Thx for warning. 🥂

libexec/scoop-import.ps1
```diff
@@ -0,0 +1,176 @@
...
+                if (checkArch($_)) {
+                    $archlArgs += "--arch 32bit "
+                }
+
+                $_ -match '\[(?<url>\S+.json)\]' > $null
+
+                try {
+                    scoop install $Matches.url $globalArgs.TrimEnd(" ") $archArgs.TrimEnd(" ")
+                }
+                catch {
```
yigitemres: I cannot figure out how to catch if url not exist? If anybody have suggestions, I'm very happy to hear it!

mimica12345: [super-julho-enigma-5918cf3ba9c24ff3e1e543798e55923952b1b2ab.zip](https://github.com/lukesampson/scoop/files/2432325/super-julho-enigma-5918cf3ba9c24ff3e1e543798e55923952b1b2ab.zip)

nitrocode:
Will there be a `Scoopfile` similar to `Brewfile` where it will contain both command line installs and installs from its extras and other buckets? also if that's true, will the exported file also auto-add whatever buckets are required?

libexec/scoop-import.ps1
```diff
@@ -0,0 +1,176 @@
...
+
+function checkFileExt ($file) {
+    $ext = [IO.Path]::GetExtension($file)
+    if ($ext -eq ".txt") {
+        return $True
+    }
+    else {
+        return $false
+    }
+}
```
nitrocode: I believe above can be reduced to this.

```
function checkFileExt ($file) {
    return ([IO.Path]::GetExtension($file) -eq ".txt")
}
```

The above simplification can be applied to functions below and would make the code more concise and thus make it more readable.

scadu:
What's the status of this merge request? I see that @yigitemres removed his repository.

I started playing with `scoop export` trying to install package from array when item not installed.
The only property available from `scoop export` seems to be `Length` which is not very useful, but most probably I'm missing something.

`import` would be really useful for provisioning. I might take a look if no one is interested in introducing this feature. Please note that I haven't touched PowerShell before except my [deploy script](https://gist.github.com/scadu/15851975cfb8e57f6b026a3ac5fe1426).

libexec/scoop-import.ps1
nitrocode: You could check status code of the url `HEAD`

```
PS > Invoke-WebRequest -Uri 'community.spiceworks.com' -Method 'HEAD' | select StatusCode

StatusCode
----------
       200
```

yigitemres:
> What's the status of this merge request? I see that @yigitemres removed his repository.

"I will be Bach" Arnold, probably...

I will fork again and work on it. If anyone comes with cleaner version I will be happy.

Thanks for all ideas, suggestions and warnings.

yigitemres:
NecroBump?! 

Hi guys, It's been a long time. I remember that we have unfinished business with this unimplemented feature. I will be very happy if you can help me. I'm gonna try to add all previously mentioned features with clean code. If you have suggestions, please comment. 🥂 For a testing purposes I need something. Can somebody export of their scoop list and paste it here if;

- The person installed app from json file.
- The person installed app from manifest url.

I'm using WSLGentoo as a scoop replacement. Because of this, I need these exports for testing purposes. Installing scoop apps with many different conditions is a real burden for me for now. 😢 

I hope that this will finished really soon with your helps.

Have a nice day! 🤗

Issue #2115, opened 2018-03-11
GitHub regex is very opinionated about version string


andreysaksonov:
https://github.com/lukesampson/scoop/blob/79bf99c3c110494d799e147263db7b6f2f921d4e/bin/checkver.ps1#L56

This doesn't work for this repo: https://github.com/spring-projects/spring-boot/releases (because version is not purely numeric).

I workarounded this for particular app manifest using custom regex (https://github.com/andreysaksonov/scoop-bucket/blob/master/springboot.json) and I think this regex is more safe. Should it be used by default for github checkver?

r15ch13:
Using a default regex that allows non numeric characters could lead to automatic update for `1.2.3-alpha1` if they are published as releases.
In the case of sprint-boost I would recommend only matching the RELEASE versions `\/releases\/tag\/(?:v)?([\d.]+)\.RELEASE`

andreysaksonov:
Thanks! Now I got the idea, probably this concept is good candidate to be added to the https://github.com/lukesampson/scoop/wiki/App-Manifests.

PR #2116, opened 2018-03-11
Add springboot manifest

andreysaksonov:
There is a corresponding issue regarding "github" checkver.ps1 code: #2115

It is workarounded here by custom regex for now.

### DIFF
--- a/bucket/springboot.json
+++ b/bucket/springboot.json
@@ -1,0 +1,23 @@
+{
+    "homepage": "https://projects.spring.io/spring-boot/",
+    "version": "2.0.0.RELEASE",
+    "license": "Apache 2.0",
+    "hash": "45362FCB133E5DB84A771DD37AF4D8B9655D936877F8E537DBDE2C1D648E488A",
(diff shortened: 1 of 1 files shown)


rasa:
Does it make sense to have the word `.RELEASE` in the version?

bucket/springboot.json
```diff
@@ -0,0 +1,23 @@
+{
+    "homepage": "https://projects.spring.io/spring-boot/",
+    "version": "2.0.0.RELEASE",
+    "license": "Apache 2.0",
+    "hash": "45362FCB133E5DB84A771DD37AF4D8B9655D936877F8E537DBDE2C1D648E488A",
```
brandon93s: The has is available at `$url.sha256`. Add the following to autoupdate to update the hash without needing to download the file: 

```json
"hash": {
    "mode": "extract",
    "url": "$url.sha256"
}
```

r15ch13: The `mode` property can be omitted, it defaults to `extract`

andreysaksonov:
Replaced with PR to extras https://github.com/lukesampson/scoop-extras/pull/838

Issue #2129, opened 2018-03-15
Latex (miktex) Error!


yigitemres:
┌[⚡ yigit] [21:36]
└[~\Desktop]> scoop install latex
Installing 'latex' (2.9.6621) [64bit]
The remote server returned an error: (302) Found.
URL https://miktex.org/download/ctan/systems/win32/miktex/setup/windows-x86/miktex-portable-2.9.6621.exe#/dl.7z is not valid

r15ch13:
Yeah, this happens because they redirect to mirrors without HTTPS. Retrying it is the only solution. (I hope letsencrypt gets more reach...)

PR #2131, opened 2018-03-16
Changing miktex url to ctan!

yigitemres:
Miktex url is not working, we have to use CTAN. #2129

### DIFF
--- a/bucket/latex.json
+++ b/bucket/latex.json
@@ -2,7 +2,7 @@
     "homepage": "https://miktex.org",
     "version": "2.9.6621",
     "license": "https://miktex.org/copying",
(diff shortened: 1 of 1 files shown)


r15ch13:
We could fix the link to ftp.fau.de, it's the only mirror that supports HTTPS.
https://ftp.fau.de/ctan/systems/win32/miktex/setup/

brandon93s:
Existing urls appear to be working now.

yigitemres:
@r15ch13 and @brandon93s Sorry for late answers. If existing urls are still working then I will close this pull request.

r15ch13:
As long as the mirrors don't use HTTPS they will fail.
I would recommend using https://ftp.fau.de/ instead of downgrading to HTTP

yigitemres:
@r15ch13 I was looking for something and I forget to push it. Thx for double reminder and suggestion! 🥂

PR #2135, opened 2018-03-17
Added RefreshEnv

Nirudium:
We needed a quick environment variable reset for less closing command prompts and more dev-ing


r15ch13:
Wouldn't it be better to implement it directly into scoop's install routine? Or is this specifically for `cmd`?

Nirudium:
We could shove a version of this into the install routine

Nirudium:
Gonna close this and make a new one

PR #2136, opened 2018-03-18
Added Auto-Refresh to both cmd and the install script

DrMkdaddy:


r15ch13:
Hm, it seems you forgot the attach the script from https://github.com/lukesampson/scoop/pull/2135
Simply adding `cmd.exe /c "refreshenv"` to the bottom of `install.ps1` doesn't work. 😄
It should be added somewhere in `install_app()` https://github.com/lukesampson/scoop/blob/9cd585adf2886324abb641af8559f313634e239b/lib/install.ps1#L12-L74

Also, it would be cool to implement it in PowerShell directly without running a separate Batch script.

Nirudium:
I'm obviously a noob when it comes to powershell but, will do!

Nirudium:
`$env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User")` Would this work?

r15ch13:
@Nirudium This seems to be specifically for the case when running `scoop` from `cmd`. Now I know why it's a batch script. Powershell command won't work because they don't effect the `cmd` session.
Hm, how do we detect if `scoop` was called from `cmd`... 😕

Nirudium:
:thinking:

PR #2150, opened 2018-03-22
bucket for a complete gradle install

neonailol:
Sometimes `bin` version of `Gradle` is not enough, particularly when using `IntelliJ Idea` with `Gradle`, version `all` required to get autocomplete, analyze, and documentation features to work.


neonailol:
@rasa i agree with you

PR #2152, opened 2018-03-24
OpenJDK 10

se35710:
Closes #1727


rasa:
Instead of this PR, let’s add https://github.com/se35710/scoop-java to our bucket list, agreed?

se35710:
Yes, let's wait for the LTS version due September.

PR #2153, opened 2018-03-25
Fix some tests on linux

leonyu:
Fix some test on Linux. 

* Currently `Get-Content` on Linux doesn't support `-Encoding` , thus cannot read file as `Byte`. It is still possible to read the test using .NET functionality. 
* Adjust path to use `join-path` in another file


leonyu:
Also I am not sure if AppVeyor supports Linux with PowerShell, if not, it is fairly easy to use CircleCI, a Docker-based cloud CI, to do Linux CI by pulling `[microsoft/powershell](https://hub.docker.com/r/microsoft/powershell/)`

Issue #2161, opened 2018-03-29
Handle failing SourceForge mirrors for gcc


twogood:
My Appveyor build fails right now because of:
```
Installing 'gcc' (7.2.0) [64bit]
The remote server returned an error: (502) Bad Gateway.
URL https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/7.2.0/threads-posix/seh/x86_64-7.2.0-release-posix-seh-rt_v5-rev1.7z is not valid
```

See https://github.com/twogood/unshield/pull/76

r15ch13:
The download itself works. SourceForge is quite the problem child...
Currently, the only solution would be to simply retry it. 😞

twogood:
Actually, I can't get that URL to work on my desktop machine either, but the same gcc version worked on March 10.

A bit off topic but maybe Scoop itself could support mirrors or fallback URLs in the future?

asmgf:
It may be related to #2109 since we don't use `downloads.sourceforge.net` for GCC.

I don't see any error though, Scoop just loads a web-page instead of the archive and fails on `Checking hash`.

Changing the `-imatch` to a more generic one or changing the URL to `https://downloads.sourceforge.net/project/mingw-w64/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/7.2.0/threads-posix/seh/x86_64-7.2.0-release-posix-seh-rt_v5-rev1.7z` fixes the issue.

twogood:
Using downloads.sourceforge.net instead sounds like a good idea!

rasa:
Fixed in #2168

twogood:
Works fine now. You are awesome!

PR #2168, opened 2018-03-31
Use downloads.sourceforge.net URLs

asmgf:
Otherwise Scoop just loads a web-page instead of the archive.

#2109 #2161


yigitemres:
and lukesampson/scoop-extras/pull/867

PR #2179, opened 2018-04-05
Update gcc to 7.3.0

fadhlika:


r15ch13:
Please change the URLs from `excellmedia.dl.sourceforge.net` to `downloads.sourceforge.net`.
This way it's not tied to a specific mirror.

Issue #2186, opened 2018-04-09
Add Bazel to an official Scoop repository


excitoon:
Hi.

I have a package in my private repository `scoop-user`: https://github.com/excitoon/scoop-user/blob/master/bazel.json and I want to add it to official one.

1. Bazel is fast and correct build system, I guess it should belong to the main bucket (like CMake).
2. Bazel requires latest version of `infozip` (missing, but I can easily add it) and `versions/python27` (we need to solve #1466 for that).

Please, confirm that I can go this way.

r15ch13:
It's already in the extras bucket. https://github.com/lukesampson/scoop-extras/blob/master/bazel.json
Feel free to improve it. Moving it to the main bucket might be a good idea.

excitoon:
Okay, I will stick to that.

gbursson:
Lads,

What is the reasoning behind having extras bucket? Maintainability? 


Greg.

r15ch13:
@gbursson separation between console only programs and GUIs

h404bi:
main bucket should only contains console apps.

r15ch13:
Done: https://github.com/lukesampson/scoop/pull/2187

PR #2187, opened 2018-04-09
Fixed Bazel dependencies and added environment variable.

excitoon:
See #2186.

Zip is taken from http://www.paehl.com and repackaged to zip (sic!) in order to avoid `7z` dependency.

### DIFF
--- a/bucket/bazel.json
+++ b/bucket/bazel.json
@@ -1,0 +1,18 @@
+{
+    "homepage": "https://bazel.build",
+    "version": "0.11.1",
+    "license": "Apache-2.0",
+    "url": "https://github.com/bazelbuild/bazel/releases/download/0.11.1/bazel-0.11.1-windows-x86_64.zip",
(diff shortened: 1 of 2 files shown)


r15ch13:
Is there a reason for changing the `zip` version?
Depending on just `python27` will fail. It should be `versions/python27` or add it as a suggestion.

rasa:
I would suggest we push the zip change to the versions bucket as 3.1d is a beta version

excitoon:
Okay, I will try to find fresh non-beta version of zip. Current zip in main is outdated. Will come back soon.

rasa:
I’m pretty sure a dependency on 7z is fine.

rasa:
Added the zip 3.1 beta at https://github.com/scoopinstaller/versions/commit/db12fed879f13cc253d2281dcb8725fae4c2892c

excitoon:
Thanks!

PR #2210, opened 2018-04-19
Added MSMPI runtime environment

SimplyKnownAsG:
This adds to the path with `env_add_path`, because the `msmpi.dll` is
required to be in the path for MPI programs to properly work.

This replaces lukesampson/scoop-extras#895

### DIFF
--- a/bucket/msmpi.json
+++ b/bucket/msmpi.json
@@ -1,0 +1,32 @@
+{
+    "homepage": "https://www.microsoft.com/en-us/download/details.aspx?id=56727",
+    "version": "9.0.1",
+    "license": "license.rtf",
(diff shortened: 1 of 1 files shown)


r15ch13:
You added `add_env_path` instead of `env_add_path` 😄

SimplyKnownAsG:
darn, good catch. I guess my testing failed because `scoop uninstall msmpi` wouldn't have removed the path and `scoop install msmpi` didn't bother either, so I had it in my path from a previous change.

I forced a push to update my changes, interestingly the pull request seems to have updated and is already using the new/rewritten commit
```
To https://github.com/SimplyKnownAsG/scoop.git
 + 5e3038bc...6d0d19d1 master -> master (forced update)
```

PR #2215, opened 2018-04-22
minishift - fix version extract dir (again)

klauern:
It seems that this field keeps getting missed when people are bumping versions, causing the install to fail.  I'm fixing it, but it would be nice if there were more places where you could put in `$version`, but it doesn't seem to be something you can just put anywhere you want, though.

### DIFF
--- a/bucket/minishift.json
+++ b/bucket/minishift.json
@@ -19,10 +19,10 @@
     "autoupdate": {
         "architecture": {
             "64bit": {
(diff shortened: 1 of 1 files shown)


se35710:
`$version` is working, but it is in the wrong section. Replace the autoupdate with:

```
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": "https://github.com/minishift/minishift/releases/download/v$version/minishift-$version-windows-amd64.zip"
            }
        },
        "hash": {
            "url": "$url.sha256"
        },
        "extract_dir": "minishift-$version-windows-amd64"
    }
```

Issue #2219, opened 2018-04-24
Update Docker 


tejashah88:
From [docker.json](https://github.com/lukesampson/scoop/blob/master/bucket/docker.json), the listed version is v17.05.0, which was released about a year ago. The latest version is v18.03.0 which was released about a month ago.

rasa:
Is 18.03.0 provided in [docker-nightly](https://github.com/lukesampson/scoop/blob/master/bucket/docker-nightly.json)?

rasa:
Where do we find 18.03? The latest at https://download.docker.com/win/static/edge/x86_64/ is 17.10, which is what Chocolately is using. See https://chocolatey.org/packages/docker/

tejashah88:
@rasa You're right, I tried searching for 18.03, but it's only released for linux and mac. I guess the least that we can ask for is updating to 17.10.

tejashah88:
Closing due to #2235. Thanks @se35710!

PR #2235, opened 2018-05-01
Docker 18.03.0-ce

se35710:
Solves #2219 

Duplicates the [docker-compose.json](https://github.com/lukesampson/scoop/blob/master/bucket/docker-compose.json) and [docker-machine.json](https://github.com/lukesampson/scoop/blob/master/bucket/docker-machine.json) manifests, plus adding Kitematic.

Maybe not the prettiest, but it does it's job.

### DIFF
--- a/bucket/docker.json
+++ b/bucket/docker.json
@@ -29,13 +29,5 @@
             "url": "$baseurl/sha256sum.txt",
             "find": "([a-fA-F0-9]{64})\\s+DockerToolbox.exe"
         }
-    },
-    "notes": [
-        "This is the docker client only. Until the docker daemon can run natively",
-        "on Windows you will need a Virtual Machine such as Boot2Docker or",
-        "Docker Machine",
-        "",
-        "Boot2Docker: https://github.com/boot2docker/windows-installer/releases/",
-        "Docker Machine: https://docs.docker.com/machine/"
-    ]
+    }
 }
--- a/bucket/docker.json
+++ b/bucket/docker.json
@@ -9,7 +9,7 @@
     },
     "homepage": "https://docker.com",
     "innosetup": true,
-    "post_install": "rm -r $dir\\installers",
+    "post_install": "rm -r \"$dir\\installers\"",
     "bin": [
         "docker.exe",
         "docker-compose.exe",
@@ -22,7 +22,7 @@
         ]
     ],
     "checkver": {
-        "url": "https://github.com/docker/toolbox/releases/latest",
+        "github": "https://github.com/docker/toolbox/releases/latest",
         "re": "/releases/tag/(?:v)?([\\d\\w.-]+)"
     },
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


bucket/docker.json
```diff
@@ -1,41 +1,39 @@
...
-            "url": "https://get.docker.com/builds/Windows/i386/docker-17.05.0-ce.zip",
-            "hash": "99820746a362466700718d33ee727b9aea993b9c2e05fb2c53ed606e9f6780f1"
+            "url": "https://github.com/docker/toolbox/releases/download/v18.03.0-ce/DockerToolbox-18.03.0-ce.exe",
+            "hash": "7c412e867e4708ec1b36005a217161131a21430a0ef0b0db3591218ec9ec5091"
         }
     },
     "homepage": "https://docker.com",
-    "extract_dir": "docker",
+    "innosetup": true,
+    "post_install": "rm -r $dir\\installers",
```
rasa: Change
```
   "post_install": "rm -r $dir\\installers",
```
To
```
   "post_install": "rm -r \"$dir\\installers\"",
```

bucket/docker.json
```diff
@@ -1,41 +1,39 @@
...
+        "docker-compose.exe",
+        "docker-machine.exe"
+    ],
+    "shortcuts": [
+        [
+            "kitematic\\Kitematic.exe",
+            "Kitematic"
+        ]
     ],
     "checkver": {
```
rasa: This can be simplified to:
```
"github": "https://github.com/docker/toolbox"
```
right?

rasa: This can be simplified to:
```
"github": "https://github.com/docker/toolbox"
```
right?

se35710: @rasa Yes, correct! Thanks.

bucket/docker.json
```diff
@@ -22,7 +22,7 @@
         ]
     ],
     "checkver": {
-        "url": "https://github.com/docker/toolbox/releases/latest",
+        "github": "https://github.com/docker/toolbox/releases/latest",
```
rasa: Please remove the `/releases/latest` and the `re` line, as they are unneeded. Thx.

se35710: re is needed due to the `v` in the release tag, the /releases/latest was removed with the [commit](https://github.com/lukesampson/scoop/pull/2235/commits/3a2b6fae07f9d16887c4200bde256008b6305a3c) after.

rasa: `re` is not needed, as checkver already has that regex. See https://github.com/lukesampson/scoop/blob/master/bin/checkver.ps1#L56

se35710: Ahh, sorry. it's the `-ce` and possibly `-rc` parts of their version naming that the default regex doesn't catch. I could add `-ce` to the download urls?

rasa: You’re right, my bad. Let’s get this puppy merged.

PR #2236, opened 2018-05-02
Fixed rsync download URL

vivus-ignis:


r15ch13:
Thanks 👍
But this is still not working. `cwrsync` has to be all lowercase: https://itefix.net/dl/cwrsync_5.7.1_x86_free.zip
:smile:

vivus-ignis:
oops, you're right =)). fixed now

PR #2244, opened 2018-05-05
Adding PowerShell Core.

Ghasan:
For the check version I did not use "github" property, as the PowerShell Core preview releases are not marked so properly, and a version like "v6.1.0-preview.2" ends up as "v6.1.0" by scoop parser, which results in an invalid update url.

### DIFF
--- a/bucket/pwsh.json
+++ b/bucket/pwsh.json
@@ -1,0 +1,30 @@
+{
+    "homepage": "https://github.com/PowerShell/PowerShell",
+    "version": "6.1.0-preview.2",
+    "license": "MIT",
+    "architecture": {
+        "32bit": {
+            "url": "https://github.com/PowerShell/PowerShell/releases/download/v6.1.0-preview.2/PowerShell-6.1.0-preview.2-win-x86.zip",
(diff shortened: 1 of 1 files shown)


h404bi:
PowerShell Core (pwsh) is in extras bucket.

Ghasan:
Oh, interesting. But then I think there is no need for this pull request. But why is it in the extras? I think it should be in the main since its used by developers for most of the time and is cleanly portable.

h404bi:
AFAIK, main bucket should only contain console application in most case, and pwsh has a GUI.

Ghasan:
I thought pwsh is a shell environment and did not expect it to contain GUI? Do you mean PowerShell ISE? I don't think that is available PowerShell Core, only the standard which runs on Windows.

Do you have a command or something that displays the GUI? I tried to Google, but nothing.

Here is a link of an issue asking for GUI, since it is not available: https://techcommunity.microsoft.com/t5/PowerShell-AMA/PowerShell-Core-console-only-No-GUI/td-p/144001

r15ch13:
It makes more sense to add it to the main bucket. You're right @Ghasan. I merged the PR without realizing that it was added to the extras bucket. 😁

r15ch13:
The preview version will not be added, because the main bucket should only provide stable builds.

PR #2253, opened 2018-05-09
Updated descriptions

gbursson:
updated descriptions of most applications from A to C.


gbursson:
That's odd, seems like VSCode is auto-removing newline on save!

Will do it in Sublime maybe.

Ash258:
Vscode setting: `"files.insertFinalNewline": true,`

gbursson:
Thanks Ash :)

I fixed (I think) busybox and bfg manifests (oddly Sublime highlighted some parts of the "original" manifests as broken.

Can't fix composer.json issue, will leave it for now. Quick question, as I am just learning git - how to revert changes for one file only (i.e. composer manifest)?

r15ch13:
@gbursson try [GitKraken](https://www.gitkraken.com/) it has an undo button.

PR #2256, opened 2018-05-12
Updated descriptions

gbursson:


gbursson:
I kinda had to leave package name in **Cmake**'s description as it's a _compound_ of **C#** and **make**.
Other suggestions are incorporated now.

PR #2258, opened 2018-05-14
Add forego

alexbepple:
I didn't find any contribution guide. So this is an attempt to the best of my knowledge. Please let me know, if and what needs improving.


r15ch13:
Nice!
Wow, I didn't know about the `autoupdate.note` property 😲
Too bad that equinox uses random strings. 😞

alexbepple:
@r15ch13 Yeah, figuring out the random strings for Equinox was a bit annoying.

PR #2273, opened 2018-05-19
Added reference to environment variables

spadino:
I added some references to environment variables and its uses, as well as explain what the basic setup gives in term of directory layout.

### The WIKI should be updated in that regard, however I can't provide a PR request for wikis...


Ash258:
Wikipage could also contain this:

https://github.com/lukesampson/scoop/issues/52#issuecomment-385661479

PR #2300, opened 2018-06-12
add jfrog-cli-go to bucket

daveneeley:
jFrog makes Artifactory, Bintray, and a few other products in the Binary Repository management space. This cli tool is a go executable that interacts with their products.


rasa:
We like to have `license` and 'description`, if possible. Can you add to this PR?

PR #2319, opened 2018-06-19
Revert rsync to version 5.5.0 and disable autoupdate

### LINKED ISSUE #2295: download-url for rsync is incorrect
https://www.itefix.net/dl/cwrsync_5.7.2_x86_free.zip only gets the gpg signature instead of the installer executable. I was unable to find/guess the correct url.

h404bi:
Fix #2295


rasa:
http://repo.msys2.org/msys/x86_64/rsync-3.1.3-1-x86_64.pkg.tar.xz Is an alternative

PR #2327, opened 2018-06-23
Change imagemagick to not add its directory to PATH. Fix #2275

### LINKED ISSUE #2275: Maybe env_add_path should adjust the paths' order for some shims
### Reproduce
```cmd
scoop install ffmpeg imagemagick
```
```cmd
where ffmpeg
~\scoop\apps\imagemagick\current\ffmpeg.exe
~\scoop\shims\ffmpeg.exe
```
### What was expected
```cmd
where ffmpeg
~\scoop\shims\ffmpeg.exe
~\scoop\apps\imagemagick\current\ffmpeg.exe
```
### Explanation
The ffmpeg's version is 4.0 currently, but the one included in imagemagick is 3.4. Some people might want to use the latest version.

I know nothing about powershell, maybe we should adjust some functions around this
https://github.com/lukesampson/scoop/blob/bcab4710ac821ac9cad95ecbd5b289b0169a7010/lib/install.ps1#L814

lahcim8:
Imagemagick doesn't need to be added to PATH because all its .exe files are already shimmed.

### DIFF
--- a/bucket/imagemagick.json
+++ b/bucket/imagemagick.json
@@ -15,7 +15,6 @@
     "depends": [
         "ffmpeg"
     ],
-    "env_add_path": ".",
     "bin": [
         "compare.exe",
         "composite.exe",


rasa:
My guess is it’s done so imagemagick’s `convert.exe` is called instead of %windir%\system32’s?

lahcim8:
This is what I get by running `where`:

```
C:\Users\lahcim8>where convert
C:\Windows\System32\convert.exe
C:\Users\lahcim8\scoop\shims\convert.exe
```

You can see the env doesn't help. I however have scoop installed only as a user. Maybe the env behaves differently if global?

I am not really power user of `imagemagick` but I have always used only `magick` command, i.e.
```magick -density 300 in.pdf out.png```

My understanding is that you can use `magick <subcommand>` instead of `<sybcommand>`. This is confirmed by running `magick convert` and `convert` and getting almost same output.

**Comment by lahcim8** [comment]
After the last update of imagemagick ([8eb5d74](https://github.com/lukesampson/scoop/commit/8eb5d746aa40ffc457b5863ae42afe551693fd68)), the `ffmpeg`-`imagemagick` clash still lasts.

This is now my `where` output:
```
C:\Users\lahcim8>where ffmpeg
C:\Users\lahcim8\scoop\apps\imagemagick\current\ffmpeg.exe
C:\Users\lahcim8\scoop\shims\ffmpeg.exe
```

And also:

```
C:\Users\lahcim8>where magick
C:\Users\lahcim8\scoop\apps\imagemagick\current\magick.exe
C:\Users\lahcim8\scoop\shims\magick.exe
```

I think if we have the possibility to use shims (because the .exe files are known), we shouldn't resolve to using `"env_add_path"` , it pollutes `PATH` and in this specific case is mere duplicate of the shim behaviour.

PR #2343, opened 2018-06-28
download before uninstalling

gitolicious:
download before uninstalling to not end up without any version installed if download fails

### DIFF
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -148,6 +148,9 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
 
     $dir = versiondir $app $old_version $global
 
+    # download before uninstalling to not end up without any version installed if download fails
+    dl_urls $app $version $manifest $bucket $architecture $dir $use_cache $check_hash
+
     write-host "Uninstalling '$app' ($old_version)"
     run_uninstaller $old_manifest $architecture $dir
     rm_shims $old_manifest $global $architecture


r15ch13:
Keep in mind that `dl_urls()` also extracts all downloaded files. I guess this should be refactored first.

gitolicious:
Ok, probably want to extract the actual download from [lib/install.ps1#L295](https://github.com/lukesampson/scoop/blob/master/lib/install.ps1#L295) then.
Why do you want to enforce `$check_hash`? It's not enforced for the download during regular installation either...

r15ch13:
Because it would download the files, check/extract and delete them. Then run the uninstaller, and starts downloading the files again 😄

gitolicious:
But removal of a downloaded file is only happening if hash check fails and execution is aborted immediately then... Or am I missing something?

r15ch13:
Sorry, I am doing three things at once right now... just a bit confused ... 😁
I meant `$use_cache` should be set to `$true` because `dl_with_cache()` will copy downloaded files to the new version directory. `install_app()` will call `dl_urls()` again, which downloads the files anew.

gitolicious:
Ok, no problem.
Still I think it would actually be ok because the user explicitly states he doesn't want to use the cache. He tells us he is ok with some extra bandwidth being used. If we set `$use_cache=$true` the cache would be touched even though the user expressed he doesn't want that.
The solution would not be optimal because theoretically it would be possible to reuse the downloaded file of the first attempt in the installation phase. Though this requires quite a bit more refactoring which I need to dig my head deeper into the code to understand it.

excitoon:
Would `scoop update pkg` help?

PR #2350, opened 2018-07-04
Add Ammonite Scala REPL

nightscape:


rasa:
The URL https://github.com/lihaoyi/Ammonite/releases/download/1.1.2/2.12-1.1.2#/amm.bat is a .jar file. Why give it a .bat suffix?

nightscape:
It's a tweaked JAR file that contains some lines of script to allow it to be used as `.bat` file.
See the docs: http://ammonite.io/#UnstableVersions

rasa:
You're right. I tested it and it works!

nightscape:
Yeah, actually quite a nice technique. It even seems to work across shells 😄

PR #2358, opened 2018-07-07
Create python2.json

npenin:


rasa:
It currently lives at https://github.com/scoopinstaller/versions/blob/master/python27.json, but users may expect it to live in main. @r15ch13 ?

r15ch13:
@rasa I think we should keep older versions in the versions bucket. On this note, ruby19 should be moved too.

rasa:
@r15ch13 Agreed.

PR #2363, opened 2018-07-07
Change multiline strings to arrays

rasa:


rasa:
See https://github.com/lukesampson/scoop/pull/2351

PR #2376, opened 2018-07-09
R - add library folder to persistent data

benplm:
I've got no deep knowledge of R, but as far as I know all downloaded packages data goes into the "library" folder, so should be safe to add it to persist, in order to preserve downloaded packages when updating to a new version.

EDIT: can anybody check the error? The manifest installs fine on my system.

### DIFF
--- a/bucket/r.json
+++ b/bucket/r.json
@@ -20,6 +20,7 @@
         "bin\\curr_arch\\rterm.exe",
         "bin\\curr_arch\\rscript.exe"
     ],
+    "persist": "library",
     "notes": [
         "You'll need to type 'r.ps1' or 'r.cmd' to run R, because in Powershell 'r' runs the last command. Alternatively 'rterm' can be used to start the interactive R terminal session.",
         "",


r15ch13:
Currently, it's not possible to merge the existing libraries with an updated library directory when updating R.

(The error was introduced by me through another commit. It has nothing to do with this PR :smile:)

benplm:
Oh, I see! I'm closing this PR, then. Thanks! 😅

thorstenkampe:
Please reopen this issue @r15ch13 . R has a multitude of ways to split system and user libraries. We don't need to merge them, R does that automatically. The options are for instance: setting `$Env:R_LIBS_USER` to `$Env:SCOOP/persist/r/library`, setting `$Env:R_ENVIRON_USER` to `$Env:SCOOP/persist/r/.Renviron` which contains `R_LIBS_USER=[...]/persist/r/library`.

Same goes for R's user profile file `.Renviron` (`$Env:R_PROFILE_USER`).

PR #2377, opened 2018-07-09
Fix license attributes to SPDX standard

rasa:
Final (?) attempt. I had to change errant `*-or-later`s to `*-only`s.

### DIFF
--- a/bucket/7zip.json
+++ b/bucket/7zip.json
@@ -1,7 +1,10 @@
 {
     "homepage": "http://www.7-zip.org/",
-    "description": "A free and open-source file archiver with a high compression ratio",
-    "license": "http://www.7-zip.org/license.txt",
+    "description": "A multi-format file archiver with high compression ratios",
+    "license": {
+        "identifier": "LGPL-2.0-only,BSD-3-Clause",
+        "url": "https://www.7-zip.org/license.txt"
+    },
     "version": "18.05",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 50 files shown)


rasa:
Once this is merged, I’ll submit https://github.com/rasa/scoop-extras/tree/rasa/fix-licenses-v2 and https://github.com/rasa/scoop-extras/tree/rasa/fix-licenses-m-z

rasa:
I documented it [here](https://github.com/lukesampson/scoop/wiki/App-Manifests#optional-properties):
license: The software license for the program. For well-known licenses, please use the identifier found at https://spdx.org/licenses/ For other licenses, use the URL of the license, if available. Otherwise, use “Freeware”, “Proprietary”, “Public Domain”, or “Shareware”, as appropriate. If multiple licenses, seperate licenses with a comma (,). If dual licensed, seperate licenses with a pipe symbol (|).

We should update for the new schema.

r15ch13:
Oh, okay. I missed that 😁

h404bi:
So is it a backward compatibility update? What about user-defined buckets? And is there any tool/script can merge this change, for 3rd-party buckets?

rasa:
It’s backwards compatible, so nothing will break. The `license` property can remain a string of any value.

h404bi:
Ok, well done. It looks like the `scoop info` sub-command should also be enhanced to handle the new license property.

rasa:
See #2380

PR #2420, opened 2018-07-22
Add TrID

Gargauth:
As this is CLI-only tool, I hope it's correct to assume it belongs to main bucket.

>  TrID is an utility designed to identify file types from their binary signatures. While there are similar utilities with hard coded logic, TrID has no fixed rules. Instead, it's extensible and can be trained to recognize new formats in a fast and automatic way. TrID has many uses: identify what kind of file was sent to you via e-mail, aid in forensic analysis, support in file recovery, etc.

http://mark0.net/soft-trid-e.html

Sadly I couldn't manage to make autoupdate to update both the main application AND definitions, as `autoupdate` accepts only single URL (hence the note). Perhaps there could be a helper function to fix this, but the main program seems to not be updated very often, so I don't think it's necessary.

### DIFF
--- a/bucket/trid.json
+++ b/bucket/trid.json
@@ -1,0 +1,24 @@
+{
+    "homepage": "http://mark0.net/soft-trid-e.html",
+    "version": "v2.24-19.07.18",
+    "url": [
+        "http://mark0.net/download/triddefs.zip",
+        "http://mark0.net/download/trid_w32.zip"
+    ],
+    "hash": [
+        "7327767055e3bfe546ea3456f51c3dce15708e3fdfc51871fed73b9309934145",
+        "ea7f82363912f5b3c79217ba8716425ec3f2514887f788dcd5a2821d0b1fc83f"
+    ],
+    "bin": [
+        "trid.exe"
+    ],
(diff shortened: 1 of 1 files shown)


r15ch13:
I like that regex! 😍 
Creating a JSON file for the program and the definition could be a way to solve this.
`trid.json` would depend on `trid-definition.json` and the definition JSON would copy `triddefs.trd` to trid by using a `post_install` script.

Gargauth:
I like that idea. But how to handle uninstall? You can't chain these two apps together, so that if you'd un- or re-install main `trid` app, so would `trid-definitions`.

bucket/trid.json
```diff
@@ -0,0 +1,24 @@
+{
+    "homepage": "http://mark0.net/soft-trid-e.html",
+    "version": "v2.24-19.07.18",
```
rasa: Shouldn’t `dd-mm-yy` be `yy-mm-dd` for semantic reasons?Should the dash be a period?

r15ch13: And `v` should be removed 👍

PR #2435, opened 2018-07-24
Removed the bucket from the app name when checking directories.

Ardelean-Calin:
In the app manifest, specifying dependencies with the bucket on which to find that dependency is supported (see the [flutter](https://github.com/lukesampson/scoop-extras/blob/master/flutter.json) manifest for an example). The problem, however, is that when running `scoop status` after installing an app such as flutter, we get an error message that some dependencies are supposedly missing:

image

This message is however wrong, since oraclejdk8 was successfully installed. The problem is that flutter checks if a dependency is installed by checking whether a directory with the name of that dependency exists in the apps folder. In this case it checks for the folder `java/oraclejdk8`, doesn't find it, since it actually exists as `oraclejdk8` and marks the dependency as being not satisfied.

This small pull request fixes that:

image

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -118,7 +118,10 @@ function app_status($app, $global) {
     }
 
     $status.missing_deps = @()
-    $deps = @(runtime_deps $manifest) | Where-Object { !(installed $_) }
+    $deps = @(runtime_deps $manifest) | Where-Object {
+        $app, $bucket, $null = parse_app $_
+        return !(installed $app)
+    }
     if($deps) {
         $status.missing_deps += ,$deps
     }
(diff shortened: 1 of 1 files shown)


r15ch13:
`installed()` is doing the right thing, but `app_status()` is passing in the wrong data:
https://github.com/lukesampson/scoop/blob/88040972a30b459a3859c7c2f883e47e19da9f84/lib/core.ps1#L117

Changing it to the following would be better.
```powershell
    $deps = @(runtime_deps $manifest) | Where-Object {
        $app, $bucket, $null = parse_app $_
        return !(installed $app)
    }
```
Your change can still be used as a fallback.

PR #2441, opened 2018-07-27
Update yarn to 1.9.2

JingYenLoh:
Welp, 1 minute later master branch updates, not the best timing 😆

Would someone like me to close this?


h404bi:
Aha! because there is auto update task running.

PR #2455, opened 2018-08-03
 Implement SourceForge checkver functionality

se35710:
In it's most simple form:
```
    "checkver": "sourceforge",
```
This will find the project name from homepage, or if that fails, from the manifest name.

A more comples example:
```
    "checkver": {
        "sourceforge": {
            "project": "pmt",
            "path": "/pngcrush-executables"
        }
    },
```
Here we provide the project name, and a path into the files section (to limit number of downloads). `project` is optional and only needed if `homepage` or manifest name is not equal to the SourceForge project name.

Regex can be applied as normal, on the `checkver` property, or on the `re` property. The default regex is `"(?:.*)?\/(?:v)?([\d.]+)\/"` plus the path to the project.

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -54,6 +54,7 @@ $queue | ForEach-Object {
     $name, $json = $_
 
     $githubRegex = "\/releases\/tag\/(?:v)?([\d.]+)"
+    $sourceforgeRegex = "(?:.*)?\/(?:v)?([\d.]+)\/"
 
     $url = $json.homepage
     if($json.checkver.url) {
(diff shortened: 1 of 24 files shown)


rasa:
@r15ch13 if you like this work, I will fix the conflicts.

niheaven:
It should be noticed that `editorconfig` is not suitable for this method, since it didn't release windows binary from 0.12.1.

rashil2000:
@se35710 any updates?

se35710:
@rashil2000 Working on it now, slowly...

se35710:
Replaced with #4560

PR #2542, opened 2018-08-29
Update dotnet-sdk to version 2.1.401

ggobbe:
It should also fix the auto update

### DIFF
--- a/bucket/dotnet-sdk.json
+++ b/bucket/dotnet-sdk.json
@@ -1,15 +1,15 @@
 {
-    "version": "2.1.301",
+    "version": "2.1.401",
     "homepage": "https://www.microsoft.com/net/",
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://download.microsoft.com/download/D/0/4/D04C5489-278D-4C11-9BD3-6128472A7626/dotnet-sdk-2.1.301-win-x64.zip",
-            "hash": "sha256:FF49924A2C7FCE5284B69F1F11D6973C30A1F79C7F058EF591042B38A4000C52"
+            "url": "https://download.microsoft.com/download/E/8/A/E8AF2EE0-5DDA-4420-A395-D1A50EEFD83E/dotnet-sdk-2.1.401-win-x64.zip",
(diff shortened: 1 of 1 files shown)


r15ch13:
Just found this `https://dotnetcli.blob.core.windows.net/dotnet/Sdk/2.1.401/` URL in https://github.com/dotnet/core/blob/master/release-notes/releases.csv
Appending `dotnet-sdk-2.1.401-win-x86.zip` works.

`https://dotnetcli.blob.core.windows.net/dotnet/Sdk/2.1.401/dotnet-sdk-2.1.401-win-x86.zip`
`https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$version/dotnet-sdk-$version-win-x86.zip`

r15ch13:
Some version check variants:
```json
"checkver": {
    "url": "https://raw.githubusercontent.com/dotnet/core/master/release-notes/releases.json"
    "re": "\"version-sdk\":\\s+\"(?<version>[\\d.]+)\","
}
```

```json
"checkver": {
    "url": "https://raw.githubusercontent.com/dotnet/core/master/release-notes/releases.csv"
    "re": "[\\d.]+,(?<version>[\\d.]+),"
}
```

ggobbe:
The only issue with those links is that is it hard to filter betas, preview, rc releases from stable ones. There is a flag lts-runtime but it can be true even for a preview version if the runtime hasn't changed. 

Whereas on their home page, their main download link is always pointing to the latest stable version.

Let me know your thoughts. Thanks

ggobbe:
@r15ch13 any update regarding this? The sdk is already two versions old and it would be nice to get the updates.

r15ch13:
Betas, previews, release candidates all have some string in the version. Matching only digits and dots eliminates those versions. The SDK also has no LTS version, so this can also be ignored.

ggobbe:
@r15ch13 I have applied your comments and updated the PR accordingly. Not sure about the auto-update because I sometimes see $matchVersion and sometimes $version.

r15ch13:
`$version` is the right way to use it.
But `$matchVersion` is the same as `$version` :grin:

PR #2543, opened 2018-08-30
Add traefik 

warrickhunter:
This pull-request adds an app manifest for traefik (https://github.com/containous/traefik).

- Releases from Github.
- Separate binaries for each environment.


r15ch13:
Nice :+1:
By adding `#/traefik.exe` to the end of all URLs it's possible to rename the file on download.

warrickhunter:
Thanks. Updated to use the file rename rather than shim.

PR #2546, opened 2018-08-31
add scoop-nsis

idleberg:


r15ch13:
Nice!
Some suggestions:
- `shortcuts` property requires a name
```json
"shortcuts": [
        [
            "NSIS.exe",
            "NSIS"
        ]
    ]
```
- the `hash` property should only contain one checksum per item in `url`. In this case, Scoop will only check with the `md5`. The `sha1` from SourceForge should be preferred (it can be extracted from the HTML source).

se35710:
What about `scoop install nsic@<version>`, eg `scoop install ncis@2.11`? Would that not remove the necessity of a separate nsic repository?

idleberg:
If it's possible, I would like that much better. I looked at other buckets (e.g. PHP) and how they do it and followed their model.

se35710:
That used to be possible with scoop, but installing multiple versions alongside seems to be broken now @r15ch13 @rasa 

Also, [this issue](https://github.com/lukesampson/scoop/issues/1550) needs to be fixed. And [this](https://github.com/lukesampson/scoop/pull/2456#issuecomment-410508157) would also be a nice feature to have.

```
PS C:> scoop install nsis@2.11
WARN  Given version (2.11) does not match manifest (3.03)
WARN  Attempting to generate manifest for 'nsis' (2.11)
Autoupdating nsis
Found: sha1:67114175c443c3b2b3181a83ca6ed5614ff81e05 using Sourceforge Mode
Writing updated nsis manifest
Installing 'nsis' (2.11) [64bit]
[...]
Download: (OK):download completed.
Checking hash of nsis-2.11.zip ... ok.
Extracting nsis-2.11.zip ... done.
Linking ~\scoop\apps\nsis\current => ~\scoop\apps\nsis\2.11
Creating shim for 'makensis'.
Creating shim for 'nsis'.
Creating shortcut for NSIS (NSIS.exe)
Creating shortcut for NSIS Compiler Interface (makensisw.exe)
Creating shortcut for NSIS Zip2Exe (zip2exe.exe)
Persisting nsisconf.nsh
'nsis' (2.11) was installed successfully!
PS C:> scoop install nsis@2.12
WARN  'nsis' (2.11) is already installed.
Use 'scoop update nsis' to install a new version.
PS C:> scoop install nsis@2.10
WARN  'nsis' (2.11) is already installed.
Use 'scoop update nsis' to install a new version.
```

idleberg:
Where can I find detailed information how this can be done? Do I simply tag my manifests?

rashil2000:
A bucket for a single app does not make much sense. [nsis](https://github.com/ScoopInstaller/Extras/blob/master/bucket/nsis.json) is present in the Extras bucket. For older versions, please make PR in Versions bucket.

PR #2555, opened 2018-09-02
Add bat.exe to scoop

meltinglava:
bat is a replacement for cat
it has:

- support for multiple files
- syntax highlighting
- line numbering by default.

It is written by the same person that maintains fd(which is in scoop). 

atm bat does not have a 32-bit binary.


r15ch13:
Thanks 👍

PR #2556, opened 2018-09-02
Rename webp.json to libwebp.json

vaguiners:


r15ch13:
Some information about why it should be renamed?

vaguiners:
this is unprofessional

vaguiners:
anyone who tries to download this package of webp utilities will think about looking for "libwebp", not "webp". you're suggesting a new pattern by just leaving webp

PR #2559, opened 2018-09-03
Create png2html.json

vaguiners:


bucket/png2html.json
```diff
@@ -1,8 +1,7 @@
 {
     "homepage": "https://www.engr.mun.ca/~holden/png2html.html",
-    "description": "transform a PNG image and a text file into pure html",
     "version": "1.1",
-    "license": "GPL-2.0-only",
```
r15ch13: It should be `GPL-2.0-only` because we use the short identifiers from https://spdx.org/licenses/

bucket/png2html.json
```diff
@@ -1,8 +1,7 @@
 {
     "homepage": "https://www.engr.mun.ca/~holden/png2html.html",
-    "description": "transform a PNG image and a text file into pure html",
```
r15ch13: This should stay

bucket/png2html.json
vaguiners: why this is causing error

r15ch13: AppVeyor had an issue (https://ci.appveyor.com/project/lukesampson/scoop/build/6481-master/messages) that was not related to this change :smile:

PR #2564, opened 2018-09-06
Fix .ps1 shim parsing logic due to my prior PR

leonyu:
Due to my prior PR #2562 which changed the content of .ps1 shims, the code that attempts to warn about shims being overwritten no longer detects app name correctly. This is to correct that. 

Since now the path of the app can be on either first line or 2nd line, my implementation just assumes shims will be utf8 files and try to find first the app name in the first 1000 characters. This works for both the old shim and the new shim, and allow for some forward flexibility towards how shim code is written going forward.

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -332,14 +332,19 @@ function get_app_name($path) {
     return ""
 }
 
+function get_app_name_from_ps1_shim($shim_ps1) {
+    if (!([System.IO.File]::Exists($shim_ps1))) {
+        return ""
+    }
+    $first_1k_text = (Get-Content $shim_ps1 -Encoding utf8 -TotalCount 1000) -join ' '
+    return get_app_name $first_1k_text
+}
+
 function warn_on_overwrite($shim_ps1, $path) {
     if (!([System.IO.File]::Exists($shim_ps1))) {
         return
     }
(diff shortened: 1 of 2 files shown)


lib/core.ps1
```diff
@@ -332,14 +332,19 @@ function get_app_name($path) {
...
+}
+
 function warn_on_overwrite($shim_ps1, $path) {
     if (!([System.IO.File]::Exists($shim_ps1))) {
         return
     }
-    $reader = [System.IO.File]::OpenText($shim_ps1)
-    $line = $reader.ReadLine().replace("`r","").replace("`n","")
-    $reader.Close()
-    $shim_app = get_app_name $line
```
r15ch13: Wouldn't it be enough to read the whole file instead of one line? The `regex` will still match.
```powershell
$line = Get-Content $shim_ps1 -Encoding utf8
$shim_app = get_app_name $line
```

leonyu: It's just in case the file is arbitrarily large, but I guess it is not a scenario we need to guard for. 

I chose it cause prior implementation used .Net to read the file line-by-line instead of Powershell.

leonyu: I push a simplified change when I get home later today.

PR #2570, opened 2018-09-08
Add deno

MarkTiedemann:


h404bi:
How about adding the `description` property?  `A secure TypeScript runtime on V8`

MarkTiedemann:
@h404bi Didn't know about that property. Thanks!

Issue #2573, opened 2018-09-09
Autoupdate error when used directory name with multibyte characters


yukihane:
I create custeom app-manifest([here](https://github.com/yukihane/scoop-bucket-yukihane-games/blob/91d960886de11fc41137918394b9de2f943da616/densya.json#L7)).

Following error occurs when updating, using manifest generated by checkver.ps1.

It seems that "extract_dir" directory name is garbled, and this problem is solved by changing encoding to UTF-8 with BOM.

> <pre>
> $ scoop update densya
> densya: 1.04 -> 1.06
> Updating one outdated app:
> Updating 'densya' (1.04 -> 1.06)
> Uninstalling 'densya' (1.04)
> Removing shim for 'densya'.
> Unlinking ~\scoop\apps\densya\current
> Installing 'densya' (1.06) [64bit]
> Loading dcnt.cgi from cache
> Checking hash of dcnt.cgi ... ok.
> Extracting densya.zip ... DEBUG
> -------------------------------------------------------------------------------
>    ROBOCOPY     ::     Windows の堅牢性の高いファイル コピー
> -------------------------------------------------------------------------------
> 
>   開始: 2018年9月9日 9:24:35
>    コピー元 : C:\Users\yuki\scoop\apps\densya\1.06\_tmp\繝・Φ繧ｷ繝｣\
>      コピー先 : C:\Users\yuki\scoop\apps\densya\1.06\
> 
>     ファイル: *.*
> 
>   オプション: *.* /S /E /DCOPY:DA /COPY:DAT /MOVE /R:1000000 /W:30
> 
> ------------------------------------------------------------------------------
> 
> 2018/09/09 09:24:35 エラー 2 (0x00000002) コピー元ディレクトリにアクセスしています C:\Users\yuki\scoop\apps\densya\1.06\_tmp\繝・Φ繧ｷ繝｣\
> 指定されたファイルが見つかりません。
> 
> ERROR Could not find '繝・Φ繧ｷ繝｣'! (error 16)
> 
> Please create a new issue by using the following link and paste your console output:
> https://github.com/lukesampson/scoop/issues/new?title=densya%401.06%3a+extract_dir+error
> True
> </pre>

yukihane:
It occurs when targeted local manifest file only.

> scoop install ./densya.json

yukihane:
[...]

PR #2574, opened 2018-09-09
use UTF-8 with BOM

yukihane:
refs. #2573


yukihane:
I'm sorry for my bad fix.

https://tools.ietf.org/html/rfc8259#section-8.1

>  Implementations MUST NOT add a byte order mark (U+FEFF) to the
   beginning of a networked-transmitted JSON text.

PR #2588, opened 2018-09-13
Added ARMClient bucket

snobu:
https://github.com/projectkudu/ARMClient
ARMClient is a simple command line tool to invoke the Azure Resource Manager API.


snobu:
Thanks for the improvements!

PR #2599, opened 2018-09-16
 Update new repo url of hugo

ManassehZhou:
hugo has changed its repo to another account. 
Fix the path issues when installing.


ManassehZhou:
ty

PR #2613, opened 2018-09-22
pstools: Add pstools

MarkTiedemann:


leonyu:
Version should be `v2.45`. You can find it on [this page](https://docs.microsoft.com/en-us/sysinternals/downloads/)

Also, I think we should package the entire Sysinternals Suite

brandon93s:
sysinternals - including pstools - is included in the extras bucket: https://github.com/lukesampson/scoop-extras/blob/master/sysinternals.json

```sh
scoop bucket add extras
scoop install sysinternals
```

PR #2623, opened 2018-09-26
add Service Bus Explorer

mattb-atomic:


rasa:
Since this is a GUI app, perhaps it should live in the extras bucket?

mattb-atomic:
Sure, I could move it there.

I did discover the extras bucket after-the-fact, but also wasn't sure that it qualified.

Issue #2414, opened 2018-07-20
Discussion about recent OpenSSH issues


r15ch13:
Wanted to discuss the issues OpenSSH is causing lately.

- Connection issues with `repo.msys2.org`: #2388
- Mismatching checksums for various files: #2412 #2375 #2373 #2372 #2371 #2366 #2353 #2414
- Extractions not working: #2386

Possible solutions:

1. leave the manifest like it is now? [openssh.json](https://github.com/ScoopInstaller/Main/blob/master/bucket/openssh.json)
2. use another download source than `repo.msys2.org`
3. swap it with the `git-with-openssh` manifest silently
  a. `version` property shows GIT version instead of OpenSSH
  b. auto update works
4. change it to an fixed version of `git-with-openssh` (Example: [rivy/openssh](https://github.com/rivy/scoop-bucket/blob/master/openssh.json))
  a. `version` property shows OpenSSH version
  b. no auto update
5. change it to `win32-openssh` silently
  a. it's still in beta
6. change it to a meta package that installs `git-with-openssh` via `depends` property?
7. change it to an empty meta package that does nothing
  a. will most likely kill all automated processes
    ```json
    {
        "homepage": "https://git-for-windows.github.io/",
        "url": "https://git-for-windows.github.io/",
        "license": "GPL-2.0",
        "version": "7.7p1",
        "suggest": {
            "Git with OpenSSH": [
                "git-with-openssh"
            ],
            "Portable OpenSSH": [
                "win32-openssh"
            ]
        },
        "notes": "This is a placeholder! Please install 'git-with-openssh' or 'win32-openssh' instead!"
    }
    ```

rasa:
Another possibility is openssh depend on git and then shim the usr/bin/ssh*.exes in the post_install. But I guess that would frustrate users who want openssh without git, or who later decide to uninstall git.

gbursson:
If I can add my 0.02PLN:
_change it to **win32-openssh** silently_

The fact that it's still in beta should not matter, I believe it's mature enough to be used.

h404bi:
Changing `openssh` it to `win32-openssh` is my suggestion. It does not matter that it's still in beta, because Windows 10 RS4 has included it as a built-in tool. This is a good enough reason.

rasa:
I added an rsync-via-git.json to show how we can easily install other apps inside the git app's directory. Initial testing is encouraging.

se35710:
[...]

PR #2632, opened 2018-09-30
Add mls-software-openssh

h404bi:
Another choice of openssh. #2414

### DIFF
--- a/bucket/mls-software-openssh.json
+++ b/bucket/mls-software-openssh.json
@@ -1,0 +1,39 @@
+{
+    "homepage": "https://www.mls-software.com/opensshd.html",
+    "description": "A Windows installer package of Openssh for Windows that is actively maintained and current.",
+    "license": "ISC",
+    "version": "7.8p1-1",
+    "url": "https://www.mls-software.com/files/setupssh-7.8p1-1.exe#/setupssh.exe",
+    "hash": "sha1:ef12f4c3148e95134614e2dd0a2b71cedb545027",
+    "installer": {
+        "args": [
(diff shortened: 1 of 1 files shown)


Ash258:
I would suggest split parameters into separate strings and rename it to something like openssh-mls

h404bi:
@Ash258 
1. I have tested that. If args are split, the installation would be broken, you could try it by making a copy locally.
2. The manifest name is the same as the choco's one, which I prefer to keep uniform app name. But if you like.

Ash258:
Work as expected. Did you followed this [page](https://www.mls-software.com/opensshd-silent.html)?

proof

Ash258:
For that naming. It would be best to keep all these openssh variations together.

Ash258:
Also isntaller for 32bit could useful. 
> /x86=#   specify set to 1 to force x86 install on x64.
Based on this it's used by default when on 32bit machine, but someone want to install 32bit while using 64bit machine.

h404bi:
I can't reproduce now, the last time I installed without using admin rights, but now it requires admin rights. Besides, there is no naming standard, so there is no best choice. As an user, I will search apps by typing `scoop search openssh`, so `openssh-mls` has no difference with `mls-software-openssh` for me. But if all openssh variations need to be kept together, it's ok to rename it. Like from `win32-openssh` to `openssh-win32`...

Ash258:
Tested with separate arguments and unelevated prompt. Work as expected, but UAC popped.
Unelevated

h404bi:
Just changed to extract mode, now say goodbye to UAC, everything works like a charm.

PR #2662, opened 2018-10-11
Allow letters in github version s

gitolicious:
There are project which include letters in their version names. E.g. _Fritzing_ (https://github.com/fritzing/fritzing-app/releases/latest) is now on `0.9.3b`. When only looking for numbers, the version is not extracted correctly.


r15ch13:
The regex is restricted to normal versions only to prevent the update service from including alpha-/beta-/pre-releases. In cases were letters are required, it's recommended to use a specific regex for the app.

PR #2696, opened 2018-10-21
Add Clang version 7.0.0

Calinou:
This makes it easy to install [Clang](https://clang.llvm.org/) and related tools (such as `clang-format`) using Scoop.


brandon93s:
@Calinou I believe this is already covered by https://github.com/lukesampson/scoop/blob/master/bucket/llvm.json.

PR #2703, opened 2018-10-25
Fix the check for ExecutionPolicy scope

bdangit:
Under certain terminals like cmder, `Get-ExecutionPolicy` will show the scope of `Process`.  This fix explicitly checks the scope of `CurrentUser`.

Signed-off-by: Ben Dang <me@bdang.it>

### DIFF
--- a/bin/install.ps1
+++ b/bin/install.ps1
@@ -12,7 +12,7 @@ if(($PSVersionTable.PSVersion.Major) -lt 3) {
 }
 
 # show notification to change execution policy:
-if((get-executionpolicy) -gt 'RemoteSigned') {
+if((get-executionpolicy -scope CurrentUser) -gt 'RemoteSigned') {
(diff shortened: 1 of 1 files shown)


r15ch13:
Makes sense. Thanks 👍

BjoernPetersen:
How _does_ it make sense though? If an ExecutionPolicy for Process is defined, it takes precedence over the policy for CurrentUser.

Because of this patch, I'm currently facing an issue with my Azure Pipeline manifest update checker. Azure starts Powershell processes with an Unrestricted policy on the Process Scope, the rest of the scopes are undefined. In my opinion, that should absolutely be enough for scoop and this patch should be reverted.

r15ch13:
This https://github.com/lukesampson/scoop/commit/a3cae34b13ac7af739eee76fd1bdc886bc9f9fe8 should work for both cases.

BjoernPetersen:
Thanks, it does

PR #2708, opened 2018-10-27
Add manifest for git-istage

johannesegger:
git-istage is a [dotnet tool](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools) that improves the staging experience for git on the command line.

dotnet tools are distributed as platform-independent NuGet packages. The client then creates platform-specific executables out of the NuGet package. This manifest describes exactly that: it's just a wrapper around installing the package with `dotnet tool` (with some quirks I must admit). The advantage is that I can manage git-istage with scoop, as most of my other software. So I don't have to think about which package I installed with which package manager.

There are already [a lot of dotnet tools](https://github.com/natemcmaster/dotnet-tools) that I would like to install with scoop instead and I might imagine that scoop manifests could be automatically generated for those. However I'm not sure what the scoop maintainers think about that.

### DIFF
--- a/bucket/git-istage.json
+++ b/bucket/git-istage.json
@@ -17,13 +17,13 @@
     "license": "MIT",
     "installer": {
         "script": [
-            "Move-Item $dir\\$version $dir\\git-istage.$version.nupkg",
+            "Move-Item \"$dir\\$version\" \"$dir\\git-istage.$version.nupkg\"",
             "$nugetConfig = \"<?xml version=`\"1.0`\" encoding=`\"utf-8`\"?><configuration><packageSources><clear /><add key=`\"Local`\" value=`\"$dir`\" /></packageSources></configuration>\"",
             "$nugetConfigFile = \"$dir\\nuget.config\"",
-            "$nugetConfig | Out-File -Encoding UTF8 $nugetConfigFile",
-            "dotnet tool install git-istage --tool-path $dir --configfile $nugetConfigFile",
+            "$nugetConfig | Out-File -Encoding UTF8 \"$nugetConfigFile\"",
+            "dotnet tool install git-istage --tool-path \"$dir\" --configfile \"$nugetConfigFile\"",
             "$runCommand = \"if (-not (Test-Path env:DOTNET_ROOT)) { `$env:DOTNET_ROOT=Split-Path -Parent (Get-Command dotnet).Path } & `\"$dir\\git-istage.exe`\" `$args\"",
-            "$runCommand | Out-File -Encoding UTF8 $dir\\run.ps1"
+            "$runCommand | Out-File -Encoding UTF8 \"$dir\\run.ps1\""
         ]
     },
     "bin": [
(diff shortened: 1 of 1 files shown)


bucket/git-istage.json
```diff
@@ -0,0 +1,35 @@
+{
+    "homepage": "https://github.com/terrajobst/git-istage",
+    "version": "0.2.29",
+    "url": "https://www.nuget.org/api/v2/package/git-istage/0.2.29",
```
Ash258: Use `"url": "https://www.nuget.org/api/v2/package/git-istage/0.2.29#/dl.nupkg"` And remove that Move-Itemon L20.

bucket/git-istage.json
```diff
@@ -0,0 +1,35 @@
...
+    ],
+    "description": "A better git add -p",
+    "license": "MIT",
+    "installer": {
+        "script": [
+            "Move-Item $dir\\$version $dir\\git-istage.$version.nupkg",
+            "$nugetConfig = \"<?xml version=`\"1.0`\" encoding=`\"utf-8`\"?><configuration><packageSources><clear /><add key=`\"Local`\" value=`\"$dir`\" /></packageSources></configuration>\"",
+            "$nugetConfigFile = \"$dir\\nuget.config\"",
+            "$nugetConfig | Out-File -Encoding UTF8 $nugetConfigFile",
+            "dotnet tool install git-istage --tool-path $dir --configfile $nugetConfigFile",
```
Ash258: Surround `$dir` with \"\" to preserve spaces in path.

bucket/git-istage.json
```diff
@@ -0,0 +1,35 @@
...
+            "$runCommand = \"if (-not (Test-Path env:DOTNET_ROOT)) { `$env:DOTNET_ROOT=Split-Path -Parent (Get-Command dotnet).Path } & `\"$dir\\git-istage.exe`\" `$args\"",
```
Ash258: Missing $ between `env:DOTNET..`

bucket/git-istage.json
johannesegger: Thanks for your comments.

I tried that first, but then scoop automatically runs 7zip and extracts the package. But the compressed package is needed when `dotnet tool install ...` is run later on. Actually I was quite surprised that scoop "knows" that .nupkg is a zip file, or maybe it just tries?

bucket/git-istage.json
johannesegger: It seems to work better without $.

`Test-Path env:DOESNT_EXIST` gives me `false` while `Test-Path $env:DOESNT_EXIST` throws an exception.

Ash258: Then make it `-not $env:DOTNET -eq $null`

bucket/git-istage.json
Ash258: Sorry my bad I forgot. Does dotnet cli check for extension?? If not just rename it to smething that is not any of these:

https://github.com/lukesampson/scoop/blob/915eef0d0210c976840e30685f1e0dace674914c/lib/decompress.ps1#L17

johannesegger:
Awesome, thanks a lot 👍

PR #2709, opened 2018-10-27
Add LinqPadless

DavoBR:


brandon93s:
linqpadless is available in the extras bucket: https://github.com/lukesampson/scoop-extras/blob/master/linqpadless.json

```sh
scoop bucket add extras
scoop install linqpadless
```

Issue #2724, opened 2018-11-02
Persisting bug with already existing folder


Ash258:
## How to reproduce

1. Install `vscode-insiders-portable`
    1. `scoop install vscode-iniders-portable -k`
1. Should be OK
    1. Installing
    1. Executing
1. Twice force update vscode-insiders (simulate two updates a day)
    1. `scoop update vscode-insiders-portable -kf`
    1. `scoop update vscode-insiders-portable -kf`
1. See error

![Bug](https://i.imgur.com/l9BhZLJ.png)

## Relates

- lukesampson/scoop-extras#1300
- lukesampson/scoop-extras#1370
- lukesampson/scoop#2671

Ash258:
Reopen please.

![Reopen](https://i.imgur.com/NY4wvuV.png)

PR #2732, opened 2018-11-04
Fix persisting bug with already existing folder.

PakornVl:
#2724


h404bi:
This could be merged i think. @r15ch13

Ash258:
Add closes statements in description for all relates listed in #2724, before merging.

dzxs:
if persist should be `System.IO.FileInfo`, then `scoop install stuff` will create the persist as `System.IO.DirectoryInfo` if there is nothing in `$persist_dir\$target`. Maybe should review.

PR #2744, opened 2018-11-05
Fix and improve svtplay-dl

krokofant:
* Fix bin reference
* Add ffmpeg suggestion
* Format json


Ash258:
Why don't rather add `"extract_dir": "svtplay-dl"?

krokofant:
@Ash258 Updated with your suggestion.

PR #2749, opened 2018-11-06
mysql: Update to 8.0.13

rasa:


bucket/mysql.json
```diff
@@ -1,43 +1,60 @@
...
-        "#Initialize data directory (without generating root password)",
-        "if (!(Test-Path \"$dir\\data\\auto.cnf\")) {",
-        "    mysqld --initialize-insecure",
-        "}",
+        "# Using $dir instead of $persist_dir, causes this error:",
+        "# [ERROR] [MY-010187] [Server] Could not open file 'C:/Users/ex/scoop/apps/mysql/current/data/host.err' for error logging: Permission denied",
+        "$dir_u = ($persist_dir -replace '\\\\', '/')",
         "#Output client configuration to my.ini file so no username is required when connecting",
-        "echo \"\" | out-file \"$dir/my.ini\" -Encoding ASCII -Append",
+        "echo \"[mysqld]\" | out-file \"$dir/my.ini\" -Encoding ASCII -Append",
```
Ash258: Rather use `Add-Content \"$dir\\my.ini\" '[mysqld]' -Encoding Ascii` instead of aliases echo and out-file

rasa: OK, I will fix. I was just cut-and-pasting.

Ash258:
Hash extraction should be possible with:

`autoupdate.64bit.hash = "$basename\\)</td>\\s+<td.*>MD5: <code class=\"md5\">([A-Fa-f\\d]{32})"`

Or just could work `md5\">([A-Fa-f\\d]{32})` since it's first match

rasa:
```
"$basename\\)</td>\\s+<td.*>MD5: <code class=\"md5\">([A-Fa-f\\d]{32})"
```
didn't work (multi-line?) but
```
md5\">([A-Fa-f\\d]{32})
```
did. Thanks!

Ash258:
You could always use `(?m)` (see [modifiers](https://www.regular-expressions.info/modifiers.html)) for multiline match. But since it should be first match all the time, it's unnecessary.

Issue #2736, opened 2018-11-04
Some autoupdates are not getting processed


rasa:
Why are some autoupdates not getting processed?
```
allure: 2.8.0 (scoop version is 2.7.0) autoupdate available
artifact: 2.0.1 (scoop version is 1.0.1) autoupdate available
mysql: 8.0.13 (scoop version is 5.7.22) autoupdate available
postgresql: 11.0 (scoop version is 10.5) autoupdate available
resharper-clt: 2018.2.3 (scoop version is 2018.1.4) autoupdate available
gource: 0.49 (scoop version is 0.47) autoupdate available
```
It's not working in the extras bucket, too:
```
chromium-dev: 71.0.3576.0-r597978 (scoop version is 70.0.3518.0-r581881) autoupdate available
fsviewer: 6.7 (scoop version is 6.6) autoupdate available
kitty: 0.70.0.6 (scoop version is 0.70.0.5) autoupdate available
mc: 4.8.21 (scoop version is 4.8.19) autoupdate available
kdenlive: 18.08.2 (scoop version is 18.04.1) autoupdate available

**Comment by r15ch13** [comment]
Broken URL or regex.
Last log of the update service:
```
Updating Scoop ...
Already up-to-date.
Cleaning Scoop cache ...
Excavating ...
Updating ...
hub pull origin master
From https://github.com/lukesampson/scoop
 * branch            master     -> FETCH_HEAD
Already up-to-date.
hub checkout master
Already on 'master'
Your branch is up-to-date with 'origin/master'.
7zip: 18.05
ant: 1.10.5
autoit: 3.3.14.5
brotli: 1.0.7
chromedriver: 2.43
cacert: 2018-10-17
ack: 2.24
astyle: 3.1
apache: 2.4.37
cockroachdb: 2.1.0
casperjs: 1.1.4-1
busybox: 2358-g25a1bcec7
bfg: 1.13.0
curl: 7.62.0
consul: 1.3.0
ag: 2.2.0-2-gbd82cd3
dos2unix: 7.4.0
ahoy: 2.0.0
dart: 2.0.0
allure: 2.8.0 (scoop version is 2.7.0) autoupdate available
Autoupdating allure
Downloading allure-2.8.0.zip to compute hashes!
The remote server returned an error: (404) Not Found.
URL https://github.com/allure-framework/allure2/releases/download/2.8.0/allure-2.8.0.zip is not valid
Could not find hash!
ERROR Could not update allure
[...]

PR #2751, opened 2018-11-07
Add or fix autoupdate in some app manifest

moosad:
Partially solves #2736


h404bi:
please fix the ci test.

moosad:
> I am not fan of this multiple manfiets PRs, but in NASM i would check version from https://nasm.us/doc/nasmdocc.html

Sorry, but I just checked the bin\checkver.ps1 output and fixed what I easily could. 
Changed the nasm checkver as suggested.

PR #2764, opened 2018-11-13
7-zip installation error fix

queses:
There is an error during 7-zip installation:
```
URL https://7-zip.org/a/7z1805-x64.msi is not valid
```
I'm not using firewall or proxy, but still getting this error. The URL is realty unavailable for my network, so I replaced it with URL's provided by SourgeForge.


rasa:
The original url is working for me.

PR #2769, opened 2018-11-14
Add the git-archive-all formula.

Kentzo:
The git-archive-all script extends git with a tool to archive repo with all its submodules.

### DIFF
--- a/bucket/git-archive-all.json
+++ b/bucket/git-archive-all.json
@@ -1,0 +1,25 @@
+{
+    "homepage": "https://github.com/Kentzo/git-archive-all",
+    "license": "MIT",
+    "version": "1.19.2",
+    "url": "https://github.com/Kentzo/git-archive-all/archive/1.19.2.zip",
+    "hash": "386d095d4b462b29cb177959e15210569e9fd78f9dffb7d1cb7129645ae8facf",
+    "extract_dir": "git-archive-all-1.19.2",
+    "installer": {
+        "script": [
+            "Push-Location \"$dir\"",
+            "try {",
+            "    python --version | Out-Null",
+            "} catch {",
+            "    Write-Host 'No Python is found' -ForegroundColor Red",
+            "}",
+            "python setup.py install",
+            "Pop-Location"
+        ]
+    },
+    "checkver": "github",
+    "autoupdate": {
+        "url": "https://github.com/Kentzo/git-archive-all/archive/$version.zip",
+        "extract_dir": "git-archive-all-$version"
+    }
+}


bucket/git-archive-all.json
```diff
@@ -0,0 +1,24 @@
...
+    "license": "MIT",
+    "version": "1.19.2",
+    "url": "https://github.com/Kentzo/git-archive-all/archive/1.19.2.zip",
+    "hash": "386d095d4b462b29cb177959e15210569e9fd78f9dffb7d1cb7129645ae8facf",
+    "extract_dir": "git-archive-all-1.19.2",
+    "depends": "python",
+    "post_install": [
+        "pushd \"$dir\"",
+        "try {",
+        "    scoop reset python",
```
Ash258: Why this reset??

Kentzo: The formula is based on `https://github.com/lukesampson/scoop/blob/master/bucket/git-up.json` and the trick is borrowed there. My idea was to use an existing recipe for the identical installation process.

Ash258: Reset is nonsense and could cause problems to users, which don't want to have actual python in path or whatever. 

It's already in try block, so doing `python --version | Out-Null` should be enough.

Ash258: Add catch block with some message like `No python found` or something like that

Kentzo: What if there is a non-scoop Python in user's PATH? Or this situation is not possible (i.e. scoop overrides PATH to avoid such conflicts)?

Ash258: That's reason why this should not depend on python. It force installing new python (Not preserving use-case when user have other python installed already).

So best way possible is move that try block inside pre_install (in catch block write message and exit with non zero code) and then keep installer block with only setup.py and context changes.

Kentzo: Isn't it expected for software (in general) installed from scoop to depend on scoop-managed software? I don't want to make assumption about the kind of Python installed by the user (e.g. too old version).

The script gets installed into Python's Scripts directory, therefore it's necessary to ensure that the python executable is scoop's Python.

Kentzo:
@Ash258 Updated per suggestions.

Kentzo:
Any further feedback?

rasa:
Shouldn’t this suggest python or versions/python2?

rasa:
If this change is still desired and viable, please open a PR in https://github.com/ScoopInstaller/Main as we cannot move it.

PR #2774, opened 2018-11-15
Fixed persisting bug when force update app with same version.

PakornVl:
Relate issue : #2724

### DIFF
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -156,6 +156,18 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
     # directory.
     $refdir = unlink_current $dir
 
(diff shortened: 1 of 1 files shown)


Ash258:
Fix persisting files, which was broken in your previous PR. https://github.com/lukesampson/scoop/pull/2732#issuecomment-438811196

proof

PakornVl:
Could you explain to me. Which package was broken.

Ash258:
Every manifest, which specify persist of file, which does not exists on persisting. In my custom [UnrealCommander](https://github.com/Ash258/scoop-Ash258/blob/master/UnrealCommander.json) it happens even if I precreate that file. 

When that file does not exists, it's automatically created as folder (Now). Before persisting failed, which is better for not messing up that file.

Fsviewer
Unreal

PR #2786, opened 2018-11-19
Add gnirehtet version 2.3

Calinou:
[gnirehtet](https://github.com/Genymobile/gnirehtet) is a tool for reverse tethering on Android. This is useful if you want to share a PC's connection to a mobile device (e.g. in an area where you cannot connect your device to a Wi-Fi network).


r15ch13:
Hash extraction can be added to `autoupdate`
```json
"hash": {
    "url": "$baseurl/SHA256SUMS.txt"
}

Calinou:
@r15ch13 Added. I also added a dependency on `adb` which is required to use gnirehtet.

PR #2797, opened 2018-11-22
pshazz: Update to version 0.2018.09.25

niheaven:
Since current version of pshazz in scoop is 0.2018.04.22, and it is in conflict with PowerShell Core v6.1.x. Recent commits solve this, and this is the update with checkver.


r15ch13:
Nice use of the `checkver` functionality 👍

PR #2799, opened 2018-11-22
Show bucket url when installation failed even it's not in buckets.json

Ash258:
When you install app, installation failed and your bucket is not in buckets.json you will get default bucket url (lukesampson/scoop) or just none link for creating issue.

## Pure powershell

Now

## pwsh

Now

### DIFF
--- a/lib/buckets.ps1
+++ b/lib/buckets.ps1
@@ -54,7 +54,7 @@ function new_issue_msg($app, $bucket, $title, $body) {
         Push-Location $bucket_path
         $remote = git config --get remote.origin.url
         $remote -match 'git@(?<url>.*):(?<username>.*)/(?<repository>.*)\.git' | Out-Null # match git@github.com:lukesampson/scoop-extras.git
-        $url = "https://www.$($matches.url)/$($matches.username)/$($matches.repository)"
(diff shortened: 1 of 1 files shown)


lib/buckets.ps1
```diff
@@ -46,11 +46,19 @@ function find_manifest($app, $bucket) {
...
-    if($manifest -and $null -eq $url -and $null -eq $bucket) {
+    $bucket_path = "$bucketsdir\$bucket"
+
+    if($manifest -and ($null -eq $url) -and ($null -eq $bucket)) {
         $url = 'https://github.com/lukesampson/scoop'
+    } elseif (Test-Path $bucket_path) {
+        Push-Location $bucket_path
+        $remote = git config --get remote.origin.url
+        $remote -match 'git@(?<url>.*):(?<username>.*)/(?<repository>.*)\.git' | Out-Null # match git@github.com:lukesampson/scoop-extras.git
+        $url = "https://www.$($matches.url)/$($matches.username)/$($matches.repository)"
```
r15ch13: `www.` should be removed

r15ch13:
Hm, maybe we should limit the ticket URL to GitHub only and display the remote URL/Repo `as is` for other services (e.g. Gitlab/Gitea/self-hosted).

Issue #2757, opened 2018-11-10
Suport autoupdate of multiple urls


Ash258:
When autoupdating manifests, binary should be able to update multiple files in manifest.

[Example manifest](https://github.com/Ash258/scoop-Ash258/blob/master/PostInstall.json)

I have manifest with custom powershell scripts. One is main with version (for checkver) and second is just file, for which i want to calculate hashes too.

There should be option like this:

```json
"autoupdate": {
    "url": [
        "https://gist.githubusercontent.com/Ash258/e143436e2f9183a650424d635b535e51/raw/PostInstall.ps1",
        "https://gist.githubusercontent.com/Ash258/e143436e2f9183a650424d635b535e51/raw/PostInstall.User.ps1"
    ]
}
```

Now when make `autoupdate.url` an array. Result of checkver is this (it merge both urls into one):

```json
"url": [
    "https://gist.githubusercontent.com/Ash258/e143436e2f9183a650424d635b535e51/raw/PostInstall.ps1 https://gist.githubusercontent.com/Ash258/e143436e2f9183a650424d635b535e51/raw/PostInstall.User.ps1",
    "https://gist.githubusercontent.com/Ash258/e143436e2f9183a650424d635b535e51/raw/PostInstall.User.ps1"
],
"hash": [
    "75269e9878d1b7238720b29285e0370df46afc59f14a7c51e10031613f9ffb1d",
    "75269e9878d1b7238720b29285e0370df46afc59f14a7c51e10031613f9ffb1d"
],
```

Now I handle it with custom [checkhashes binary](https://github.com/lukesampson/scoop/blob/afb8622262f205e3958ed64ea944c2a526aec172/bin/checkhashes.ps1) (https://github.com/lukesampson/scoop/pull/2766)

Ash258:
Duplicate of #2176

PR #2803, opened 2018-11-23
Add multiple URL support for autoupdate

rasa:
Fixes (?) #2757. See also https://github.com/lukesampson/scoop-extras/pull/1459

### DIFF
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -332,7 +332,7 @@ function lets_do_autoupdates([String] $app, $json, [String] $version, $substitut
 
         if($valid) {
             # create hash
-            if ($json.autoupdate.hash.length > $i) {
+            if ($json.autoupdate.hash.length -gt $i) {
                 $h = $json.autoupdate.hash[$i]
             } else {
                 $h = $null
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -368,7 +368,7 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
 
     if ($json.url) {
         if ($json.autoupdate.url -is [System.Array]) {
-            $return = lets_do_autoupdates $app $json $version $substitutions
+            $has_changes, $has_errors = lets_do_autoupdates $app $json $version $substitutions
             $has_changes = $return[0]
             $has_errors = $return[1]
         } else {
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -250,11 +250,11 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
     return $hash
 }
 
-function update_manifest_with_new_version($json, [String] $version, [String] $url, [String] $hash, $architecture = $null) {
+function update_manifest_with_new_version($json, [String] $version, $url, $hash, $architecture = $null) {
     $json.version = $version
 
     if ($null -eq $architecture) {
-        if ($json.url -is [System.Array]) {
+        if ($json.url -is [System.Array] -and $url -isnot [System.Array]) {
             $json.url[0] = $url
             $json.hash[0] = $hash
         } else {
@@ -263,7 +263,7 @@ function update_manifest_with_new_version($json, [String] $version, [String] $ur
         }
     } else {
         # If there are multiple urls we replace the first one
-        if ($json.architecture.$architecture.url -is [System.Array]) {
+        if ($json.architecture.$architecture.url -is [System.Array] -and $url -isnot [System.Array]) {
             $json.architecture.$architecture.url[0] = $url
             $json.architecture.$architecture.hash[0] = $hash
         } else {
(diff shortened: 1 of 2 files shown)


lib/autoupdate.ps1
```diff
@@ -297,34 +297,81 @@ function get_version_substitutions([String] $version, [Hashtable] $customMatches
...
         # create new url
-        $url   = substitute $json.autoupdate.url $substitutions
-        $valid = $true
+        $url   = substitute $json.autoupdate.url[$i] $substitutions
+        $urls += $url
 
         if($valid) {
             # create hash
-            $hash = get_hash_for_app $app $json.autoupdate.hash $version $url $substitutions
+            if ($json.autoupdate.hash.length > $i) {
```
Ash258: This would never work. `>` is not comparator.

rasa: You're right. Duh.

rasa: Fixed

rasa:
Ready for testing/marging

Ash258:
I will look into it in today and tommorow and give some feedback

niheaven:
A `hashExtractionOrArrayOfHashExtrasctions` should be added in d`shema.json` that looks like:

```json
"hashExtractionOrArrayOfHashExtrasctions": {
    "anyOf": [
        {
            "$ref": "#/definitions/hashExtraction"
        },
        {
            "items": {
                "$ref": "#/definitions/hashExtraction"
            },
            "minItems": 1,
            "type": "array",
            "uniqueItems": true
        }
    ]
},
```

and change `hash`'s `$ref` to `hashExtractionOrArrayOfHashExtrasctions` in `autoupdate` so as to enable multiple hash extraction in au.

There might be some tweaks in hash extraction handling for multiple URL autoupdate, e.g. using same hash extraction formula for multi-files, or some other variations. Let's think about it~

rasa:
@niheaven Good catch! I'll add it now.

rasa:
It's working, but the order of the hashes is wrong.

rasa:
To get all the *other* .jsons to validate, we need to change
```json
        "uriOrArrayOfUris": {
            "anyOf": [
                {
                    "format": "uri",
                    "not": {
                        "pattern": "(\\$)"
                    },
```
to
```json
        "uriOrArrayOfUris": {
            "anyOf": [
                {
                    "format": "uri",
```
or
```json
        "uriOrArrayOfUris": {
            "anyOf": [
                {
                    "format": "uri",
                    "not": {
                        "pattern": "(^\\$)"
                    },
```
Is this pattern here to avoid empty URLs? If so, then doesn't `^\\$` make more sense?

r15ch13:
No that pattern disallows URLs containing $, so the autoupdater can't slip an unresolved variable in :smile:

rasa:
@r15ch13 Two points:
1. Autoupdate URLs have dollar signs
2. Dollar signs are allowed in URLs, per [here](https://perishablepress.com/stop-using-unsafe-characters-in-urls/)
I've addressed 1. in the schema. Let's ignore 2. for now.

lib/autoupdate.ps1
```diff
@@ -318,34 +318,81 @@ function get_version_substitutions([String] $version, [Hashtable] $customMatches
...
+    [Bool]$valid = $true
+    $substitutions = get_version_substitutions $version $matches
 
-        # write changes to the json object
-        if ($valid) {
-            $has_changes = $true
-            update_manifest_with_new_version $json $version $url $hash
+    if ($json.url) {
+        if ($json.autoupdate.url -is [System.Array]) {
+            $return = lets_do_autoupdates $app $json $version $substitutions
```
Ash258: This could be only:

```suggestion
            $has_changes, $has_errors = lets_do_autoupdates $app $json $version $substitutions
```

niheaven:
Now the codes are applied to noarch, and they should be copied to arch-32 and arch-64 too.

BTW, the hash handling part is insufficiency, I thought, and a specialized function to handle multiple hash extracting methods may help. I'm trying to make one, but maybe later.

lib/autoupdate.ps1
```diff
@@ -318,34 +318,81 @@ function get_version_substitutions([String] $version, [Hashtable] $customMatches
...
-        $valid = $true
+        $url   = substitute $json.autoupdate.url[$i] $substitutions
+        $urls += $url
 
         if($valid) {
             # create hash
-            $hash = get_hash_for_app $app $json.autoupdate.hash $version $url $substitutions
+            if ($json.autoupdate.hash.length -gt $i) {
+                $h = $json.autoupdate.hash[$i]
+            } else {
```
niheaven:
```suggestion
            } elseif ($json.autoupdate.hash.length -ne 0) {
                $h = $json.autoupdate.hash[-1]
            } else {
```

If `hash`'s length is less than `url`, the last `hash` expression is reused for remaining `url`s.

So we could use one `hash` expression for all these `url`s if hash extraction url and regex are the same, e.g., `$sha256.*?$basename`, different hashes for different `$basename`s.

lib/autoupdate.ps1
```diff
@@ -318,34 +318,81 @@ function get_version_substitutions([String] $version, [Hashtable] $customMatches
...
+    $has_errors = $false
+    [Bool]$valid = $true
+    $substitutions = get_version_substitutions $version $matches
 
-        # write changes to the json object
-        if ($valid) {
-            $has_changes = $true
-            update_manifest_with_new_version $json $version $url $hash
+    if ($json.url) {
+        if ($json.autoupdate.url -is [System.Array]) {
```
niheaven: This if...else... should be applied to `$json.architecture`'s `ForEach-Object` script too.

niheaven:
@rasa Could I keep on working with multi-urls and hashes autoupdate support based on this PR?

rasa:
> @rasa Could I keep on working with multi-urls and hashes autoupdate support based on this PR?

Sure! Sorry I dropped the ball on implementing your suggestions.

rasa:
Moved to #3518

PR #2806, opened 2018-11-25
Add rcedit version 1.1.1

Calinou:
[rcedit](https://github.com/electron/rcedit) is a command-line tool that can be used to edit resources embedded in PE executables (such as icons).

### DIFF
--- a/bucket/rcedit.json
+++ b/bucket/rcedit.json
@@ -1,0 +1,28 @@
+{
+    "homepage": "https://github.com/electron/rcedit",
+    "description": "Command-line tool for editing resources of PE executables",
+    "license": "MIT",
+    "version": "1.1.1",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


bucket/rcedit.json
```diff
@@ -0,0 +1,28 @@
...
+    },
+    "bin": "rcedit.exe",
+    "checkver": "github",
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/electron/rcedit/releases/download/$version/rcedit-x64.exe"
+            },
+            "32bit": {
+                "url": "https://github.com/electron/rcedit/releases/download/$version/rcedit-x86.exe"
```
Ash258: Wrong format in both URLs. `download/v$version` instead of `download/$version`

Ash258: Also missing rewrite `#/rcedit.exe`

PR #2842, opened 2018-12-04
fix SHA256 sums for 32-bit and 64-bit arch

klauern:
I'm not sure how it happened, but the checksums for this version weren't matching.


rasa:
Sorry, I didn’t see this PR when I did 
https://github.com/lukesampson/scoop/commit/bdf694de5ae0cdd1379dc537487c9ebabed36fbf

PR #2852, opened 2018-12-05
[autoupdate] base64 hash extraction length

Ash258:
sha512 encoded hash length for [Franz](https://github.com/meetfranz/franz/releases/download/v5.0.0-beta.19/latest.yml) is 88 characters length and autoupdate just dont extract it's value.

If anyone have better fix, feel free to edit.

Relate lukesampson/scoop-extras#1487

### DIFF
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -58,7 +58,7 @@ function find_hash_in_textfile([String] $url, [String] $basename, [String] $rege
     # convert base64 encoded hash values
     if ($hash -match '^(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=|[A-Za-z0-9+\/]{4})$') {
         $base64 = $matches[0]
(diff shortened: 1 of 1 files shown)


Ash258:
No. It's already implemeted. (lukesampson/scoop@04c9ddeb6d3b99496c39543ad468d34f4f1adeff)

r15ch13:
Actually, it would be enough to fix the condition.
from: `$hash.length -in @(32, 40, 64, 128)`
to: `$hash.length -notin @(32, 40, 64, 128)`

This check prevents overwriting `$hash` with garbage if the result in `$base64` contains valid base64 characters but not a real hash. It's just a safety net, I messed up 😁

Ash258:
Ok. Tested with `-notin` variant with franz and rambox pro (from extras) and it's working 👍
Do you know some others manifests, which have encoded hashes?

r15ch13:
@Ash258 
https://github.com/lukesampson/scoop-extras/blob/master/blink1control2.json
https://github.com/lukesampson/scoop-extras/blob/master/inboxer.json
https://github.com/lukesampson/scoop-extras/blob/master/signal.json
https://github.com/lukesampson/scoop-extras/blob/master/cacher.json

Some of them require changes to work because the `.yml` has changed.

PR #2863, opened 2018-12-06
Add recycle-bin

Ash258:


iquiw:
@Ash258 @rasa This shouldn't be put under `bucket` directory?

Ash258:
@iquiw You're right. Misclicked folder.

PR #2870, opened 2018-12-09
Fix winpython's checkver and switch to innosetup

benplm:
I had to change the checkver url to get it updated.
Also, since Python 3.7, Winpython switched to innosetup, so I updated the manifest accordingly.

To remove all optional stuff I wanted to use the "extract_dir" as before, but with innosetup I had to write a pre_install script which, I'm afraid, it may not be the best in terms of efficiency... so please be welcome to edit it if you think there's a better way to accomplish that.

### DIFF
--- a/bucket/winpython.json
+++ b/bucket/winpython.json
@@ -17,7 +17,7 @@
     "installer": {
         "script": [
             "Get-ChildItem \"$dir\" -Exclude 'python-*' | Remove-Item -Force -Recurse",
-            "$fold = (Get-ChildItem \"$dir\")[0] | Select-Object -ExpandProperty FullName",
+            "$fold = \"$dir\\\" + (Get-ChildItem \"$dir\")[0]",
             "Move-Item \"$fold\\*\" \"$dir\" -Force",
             "Remove-Item $fold -Recurse"
         ]
--- a/bucket/winpython.json
+++ b/bucket/winpython.json
@@ -17,7 +17,7 @@
     "installer": {
         "script": [
             "Get-ChildItem \"$dir\" -Exclude 'python-*' | Remove-Item -Force -Recurse",
-            "$fold = \"$dir\\\" + (Get-ChildItem \"$dir\")[0]",
+            "$fold = \"$dir\\\" + ((Get-ChildItem \"$dir\")[0])",
             "Move-Item \"$fold\\*\" \"$dir\" -Force",
             "Remove-Item $fold -Recurse"
         ]
--- a/bucket/winpython.json
+++ b/bucket/winpython.json
@@ -17,9 +17,8 @@
     "installer": {
         "script": [
             "Get-ChildItem \"$dir\" -Exclude 'python-*' | Remove-Item -Force -Recurse",
-            "$fold = \"$dir\\\" + ((Get-ChildItem \"$dir\")[0])",
-            "Move-Item \"$fold\\*\" \"$dir\" -Force",
-            "Remove-Item $fold -Recurse"
+            "Move-Item \"$dir\\python-*\\*\" \"$dir\" -Force",
+            "Remove-Item \"$dir\\python-*\""
         ]
     },
     "bin": [
--- a/bucket/winpython.json
+++ b/bucket/winpython.json
@@ -1,19 +1,20 @@
 {
     "description": "Free, open-source and portable Python distribution for Windows",
-    "version": "3.6.7.0",
+    "version": "3.7.1.0",
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://downloads.sourceforge.net/project/winpython/WinPython_3.6/3.6.7.0/WinPython64-3.6.7.0Zero.exe#/dl.7z",
-            "hash": "e52e2606f3a2f7d79fd9eac0701b144efe0832ecdcf7c5a2f3bf3aebba4e70c6",
-            "extract_dir": "python-3.6.7.amd64"
+            "url": "https://downloads.sourceforge.net/project/winpython/WinPython_3.7/3.7.1.0/WinPython64-3.7.1.0Zero.exe",
+            "hash": "4e81abd373c089cd9b6687393bb3d179e6792a97d8f088f77d35b9067f118c0f",
+            "pre_install": "Get-ChildItem -Path \"$dir\" -Recurse -Exclude \"python-3.7.1.amd64\" | Select -ExpandProperty FullName | Where {$_ -notlike \"$dir\\python-3.7.1.amd64\\*\"} | sort length -Descending | Remove-Item -Force; Move-Item \"$dir\\python-3.7.1.amd64\\*\" \"$dir\"; Remove-Item \"$dir\\python-3.7.1.amd64\""
         },
         "32bit": {
-            "url": "https://downloads.sourceforge.net/project/winpython/WinPython_3.6/3.6.7.0/WinPython32-3.6.7.0Zero.exe#/dl.7z",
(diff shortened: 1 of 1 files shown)


benplm:
@Ash258 it only keeps the folder containing the basic Python installation, while removing all the rest. I know it's not quite elegant, but it makes sure that we do not delete any file we want to keep.

Do you say it's a _mess_ because it's one line, or because you think it's over-complicated? Can you suggest a simpler way to accomplish the same?

Ash258:
What are this nonbasic installtion files?? List them, there has to be better way to delete them.

benplm:
@Ash258 all the files except those inside the directory "python-3.7.1.amd64". The script just removes all the other files and then moves the content of that directory to the parent directory.

Basically:
Remove `$dir\*` except `$dir\python-3.7.1.amd64\*`

The best way would be to only extract the content of this folder in the first place, but it's not possible anymore with 7zip and the "extract_dir" option since they switched to innosetup. Any thoughts?

Ash258:
Multiplatform no need for update.

```powershell
Get-ChildItem \"$dir\" -Exclude 'python-*' | Remove-Item -Force -Recurse
$fold = (Get-ChildItem \"$dir\")[0] # You could assume it would be first and only one
Move-Item "$fold\*" \"$dir\" -Force
Remove-Item $fold -Recurse
```

bucket/winpython.json
```diff
@@ -23,22 +24,21 @@
...
         ]
     ],
-    "env_add_path": [
-        "scripts"
-    ],
+    "env_add_path": "scripts",
     "checkver": {
-        "url": "https://winpython.github.io/md5_sha1.txt",
-        "re": "-([\\d.]+)Zero.exe"
+        "url": "https://github.com/winpython/winpython/releases",
```
Ash258: Switch to `"github": "https://github.com/winpython/winpython/"`

Ash258:
Why don't use github download urls?

Ash258:
Linted manifest

```json
{
    "description": "Free, open-source and portable Python distribution for Windows",
    "version": "3.6.7.0",
    "license": "MIT",
    "homepage": "https://winpython.github.io/",
    "architecture": {
        "64bit": {
            "url": "https://downloads.sourceforge.net/project/winpython/WinPython_3.6/3.6.7.0/WinPython64-3.6.7.0Zero.exe",
            "hash": "e52e2606f3a2f7d79fd9eac0701b144efe0832ecdcf7c5a2f3bf3aebba4e70c6"
        },
        "32bit": {
            "url": "https://downloads.sourceforge.net/project/winpython/WinPython_3.6/3.6.7.0/WinPython32-3.6.7.0Zero.exe",
            "hash": "8b6b22f36aecb80ce8b13212deb4bbba0e9a902ddd081d7c6288cf0f44e419be"
        }
    },
    "innosetup": true,
    "installer": {
        "script": [
            "Get-ChildItem \"$dir\" -Exclude 'python-*' | Remove-Item -Force -Recurse",
            "$fold = (Get-ChildItem \"$dir\")[0] # You could assume it would be first and only one",
            "Move-Item \"$fold\\*\" \"$dir\" -Force",
            "Remove-Item $fold -Recurse"
        ]
    },
    "bin": [
        "python.exe",
        "pythonw.exe",
        [
            "python.exe",
            "python3"
        ]
    ],
    "env_add_path": "scripts",
    "checkver": {
        "github": "https://github.com/winpython/winpython/",
        "regex": "-([\\d\\.]+)Zero\\.exe"
    },
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": "https://downloads.sourceforge.net/project/winpython/WinPython_$majorVersion.$minorVersion/$version/WinPython64-$versionZero.exe"
            },
            "32bit": {
                "url": "https://downloads.sourceforge.net/project/winpython/WinPython_$majorVersion.$minorVersion/$version/WinPython32-$versionZero.exe"
            }
        },
        "hash": {
            "url": "https://winpython.github.io/md5_sha1.txt",
            "find": "([a-fA-F\\d]{64})\\s\\|\\s$basename"
        }
    }
}
```

Proof of work

benplm:
Hi @Ash258 ,

Thanks so much for the installer script!

However, `(Get-ChildItem \"$dir\")[0]` will just return the folder name, not the full path. So, when it tries to run the Move-Item command, it returns that the path `[current-directory]\python-x.x.x.amd64` doesn't exists.
You can test by running `ls (Get-ChildItem \"$dir\")[0]` from a different directory, it will search the folder inside your current dir.

The solution would be to get the full path:
`"$fold = (Get-ChildItem \"$dir\")[0] | Select-Object -ExpandProperty FullName"`

In this way it gets the full path `C:\blah-blah\python-x.x.x.amd64`
And it works fine.

I'll push the commit asap, thanks again.

bucket/winpython.json
```diff
@@ -1,20 +1,27 @@
...
+            "url": "https://downloads.sourceforge.net/project/winpython/WinPython_3.7/3.7.1.0/WinPython32-3.7.1.0Zero.exe",
+            "hash": "a09c79c7630278f831d257c03bab99f76d3a0e23f8b4362a3e1e2495a3aa088f"
         }
     },
-    "homepage": "https://winpython.github.io/",
+    "innosetup": true,
+    "installer": {
+        "script": [
+            "Get-ChildItem \"$dir\" -Exclude 'python-*' | Remove-Item -Force -Recurse",
+            "$fold = \"$dir\\\" + (Get-ChildItem \"$dir\")[0]",
```
Ash258: This would be evaluated as `someDir" + (Get-ChildItem "$dir")[0])`

Ash258: Just changed it to: `$fold = \"$dir\\\" + ((Get-ChildItem \"$dir\")[0])`

Ash258: Sorry. Misslooked one quotation marks.

benplm: Great, thanks for the advice!

benplm: Ok, it should be fine now

Ash258: Still seeing url instead of github

`"github": "https://github.com/winpython/winpython/"`

benplm: As you said, if I use github it fails in getting the latest version, since their tags do not match the software's version.
i.e. it matches 3.6.7.0, while the latest version is 3.7.1.0.
They're both maintained, but I guess we should stick to the latest stable version, right?

benplm: Oh, I see, the order of the assets is lexicographic in the tag's url (https://github.com/winpython/winpython/releases/tag/1.11.20181031), while it's not in the release url (https://github.com/winpython/winpython/releases) so the first occurrence is different.

Is there a way to match the occurrence with the greater number (numerical sorted)?

Ash258: What?? No.

Just change `"url": "github.com/.../releases/latest/"` to `"github": "https://github.com/winpython/winpython/"` It's same, but if github change their URLs it will be easier to change one file, instead of bunch of manifests.

benplm: I see that would be the best practice, so we stick to the previous version (3.6.7.0)?
cattura

Ash258: Weird. But easily fixable.

Add reverse: true

proof

benplm: Cool! Let's hope it'll remain consinstent with the future releases 😆

r15ch13: Can be condensed to:

```json
    "installer": {
        "script": [
            "Get-ChildItem \"$dir\" -Exclude 'python-*' | Remove-Item -Force -Recurse",
            "Move-Item \"$dir\\python-*\\*\" \"$dir\"",
            "Remove-Item 'python-*'"
        ]
    },
```

benplm: Good, I changed it accordingly

PR #2889, opened 2018-12-13
Add Alacritty

notlmn:
Alacritty from https://github.com/jwilm/alacritty

This is my first scoop script so if I'm doing anything wrong please do point out.


Ash258:
Śhould be moved to extras since it's GUI application. Also why to use working directory parameter when it's by default?? Based on their documentation: 

> On Windows the config file is located at:
> `%UserProfile%\alacritty.yml`

Ash258:
But maybe these applications have some exlusions, since cmder is in main too.

notlmn:
Opened in scoop-extras: https://github.com/lukesampson/scoop-extras/pull/1514

PR #2890, opened 2018-12-13
scoop-uninstall: prevent read-only junction from breaking uninstall

h404bi:
**The work is done, now needs code review and somebody to test, don't merge currently**.
Follow up #2882


h404bi:
Closed it, I'll open a brand new pull-request which combines #2891 with #2890 #2882 , totally rewrited logic of data persisting.

PR #2891, opened 2018-12-13
install: totally rewritten on data persistting

h404bi:
**still WIP don't merge, needs code review and somebody to test**.
Follow up #2890 

To participate in the code review and tests, go visit https://github.com/h404bi/scoop-persist-test for test cases.

Issue #2881, opened 2018-12-12
scoop-cleanup: Remove-Item : Index (zero based) must be greater than or equal to zero...


h404bi:
```
$ scoop cleanup *
Removing cmder: 1.3.8Remove-Item : Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
At C:\Users\<omitted>\scoop\apps\scoop\current\libexec\scoop-cleanup.ps1:51 char:9
+         Remove-Item $dir -Recurse -Force
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : NotSpecified: (:) [Remove-Item], FormatException
+ FullyQualifiedErrorId : System.FormatException,Microsoft.PowerShell.Commands.RemoveItemCommand

 1.3.9Remove-Item : Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
At C:\Users\<omitted>\scoop\apps\scoop\current\libexec\scoop-cleanup.ps1:51 char:9
+         Remove-Item $dir -Recurse -Force
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : NotSpecified: (:) [Remove-Item], FormatException
+ FullyQualifiedErrorId : System.FormatException,Microsoft.PowerShell.Commands.RemoveItemCommand


Removing php: 7.2.12Remove-Item : Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
At C:\Users\<omitted>\scoop\apps\scoop\current\libexec\scoop-cleanup.ps1:51 char:9
+         Remove-Item $dir -Recurse -Force
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : NotSpecified: (:) [Remove-Item], FormatException
+ FullyQualifiedErrorId : System.FormatException,Microsoft.PowerShell.Commands.RemoveItemCommand


Everything is shiny now!
```

```
$ $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.1.1
PSEdition                      Core
GitCommitId                    6.1.1
OS                             Microsoft Windows 6.1.7601 S
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

[...]

Issue #2832, opened 2018-12-03
scoop-cleanup is not friendly to restricted user


h404bi:
https://github.com/lukesampson/scoop/blob/master/libexec/scoop-cleanup.ps1#L45

```powershell
> fsutil.exe
The FSUTIL utility requires that you have administrative privileges.
```

PR #2897, opened 2018-12-14
rewrite data persisting

### LINKED ISSUE #2724: Persisting bug with already existing folder

### LINKED ISSUE #2900: Cleanup is not $dir aware
Regression after e4e76d3a32fcc33ca6a0fe6b4d3b99988e6863be 

proof

/cc @h404bi

h404bi:
**Data matters, please review**

This is a rework of #2891 #2890 and #2882 . This will also fix #2724 and fix #2900 , close #2779 

**Introduction**

I add a function called `unlink_persist_data($dir)`, which recursively unlink all junction/hard link in the given directory. This affects some sub-commands which have interaction with junction/hard link:

- `scoop-install`: the persisting logic has been improved as follow:
  1. if there are data in the `persist` folder, also in app(`source`) folder, rename that one located in app folder (e.g. app's default setting files) with a `original` suffix, then create link from `persist` to  app
  2. if there are data in the `persist` folder, but no in app(`source`) folder, just create link from `persist` to  app
  3. if there is no data in the `persist` folder (e.g. fresh install), but there are data in app(`source`) folder (e.g. app's default setting files), we will just use that default setting files as the first-time persisting. So move that files from app folder to `persist` folder, then create link from `persist` to  app
  4. But what if if there is neither data in the `persist` folder (e.g. fresh install), nor in the app(`source`) folder (e.g. setting files will be created after first startup, like `Everthing.db`). We need to create empty persisting target in the `persist` folder. But by default we can't know if a persisting target is a file or a directory (e.g. `conf.d`). So we create a directory by default, and to avoid this, manifest maintainers should use `pre_install` to manually create the source file before persisting.
- `scoop-reset`: `reset` command uses the logic of `install`, but there is a problem: before re-persisting data, there have been `junction/hard link` in the app(`source`) folder. It will throw an exception of file exists #2724 . To fix this, we should unlink all old link before re-persisting, using `unlink_persist_data`.
- `scoop-uninstall`: `Remove-Item` can not remove `NTFS junction`, we need to unlink all persistting data, before uninstalling/deleting an app, but keeping persisting data.
- `scoop-cleanup`: like `uninstall`, `Remove-Item` can not remove `NTFS junction`, we need to unlink all persistting data, before deleting old versions of an app. Before PR #2882 , it use `fsutil.exe` to unlink junction, which is not friendly to restricted users (#2832 and #2881 ).

Beyond the logic improvement, there is a new feature now: it supports sub-folder data persisting, like:

```json
{
    "homepage": "https://scoop.sh",
    "description": "A dummy manifest for scoop tests.",
    "license": "Freeware",
    "version": "1.1.0",
    "url": "https://get.scoop.sh",
    "pre_install": [
        "if (!(test-path \"$dir\\dummy.txt\")) { new-item -force \"$dir\\dummy.txt\" -itemtype file | out-null }",
        "if (!(test-path \"$dir\\dummydir\")) { new-item \"$dir\\dummydir\" -itemtype directory | out-null }",
        "if (!(test-path \"$dir\\subdir\")) { new-item \"$dir\\subdir\" -itemtype directory | out-null }",
        "if (!(test-path \"$dir\\subdir\\subdummydir\")) { new-item \"$dir\\subdir\\subdummydir\" -itemtype directory | out-null }",
        "if (!(test-path \"$dir\\subdir\\subdummy.txt\")) { new-item -force \"$dir\\subdir\\subdummy.txt\" -itemtype file | out-null }",
    ],
    "persist": [
        "dummy.txt",
        "dummydir",
        "subdir\\subdummydir",
        "subdir\\subdummy.txt"
    ]
}
```
So no need to strip directories of source for target anymore.

To participate in the code review and tests, go visit https://github.com/h404bi/scoop-persist-test for test cases.

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -1262,11 +1262,11 @@ function unlink_persist_data($dir) {
     Get-ChildItem -Recurse $dir | ForEach-Object {
         $file = $_
         if ($null -ne $file.LinkType) {
-            $filepath = Resolve-Path $file
+            $filepath = $file.FullName
             # directory (junction)
             if ($file -is [System.IO.DirectoryInfo]) {
                 # remove read-only attribute on the link
-                attrib -R /L $file
+                attrib -R /L $filepath
                 # remove the junction
                 & "$env:COMSPEC" /c "rmdir /s /q $filepath"
             } else {
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -1202,13 +1202,16 @@ function persist_def($persist) {
     }
 
     if (!$target) {
-        $target = fname($source)
+        $target = $source
     }
 
     return $source, $target
 }
 
 function persist_data($manifest, $original_dir, $persist_dir) {
+    # unlink all potential old link before doing data persisting
+    unlink_persist_data $original_dir
+
     $persist = $manifest.persist
     if($persist) {
         $persist_dir = ensure $persist_dir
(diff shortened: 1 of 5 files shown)


r15ch13:
About those hard-links:
I found that not all of them work. Like [Witch Blast](https://github.com/Calinou/scoop-games/pull/10) for example. `config.dat` and `game.sav` get created on the first start and first new game.
The `pre_install` creates an empty `game.sav` file, but the game removes it on the first start without unlinking it (probably because the format is wrong). Saving the game will write a new `game.sav` but the file in `$persist_dir` is still 0kb.
For these cases, my idea was, that all hard-linked files get copied over to `$persist_dir` if the directory is modified in some way (like uninstall/update/reset).
This could also solve the problem with files that are not created on install like the aforementioned `Everything.db`.

h404bi:
Added the trigger for closing issue #2900  , there is no further changes I can do for this pull-request.

Issue #2906, opened 2018-12-16
Scoop does not update itself automatically when user has never done `scoop update` before


nimzo6689:
When running `scoop install` or `scoop update`, Scoop should update itself if last update timestamp is earlier than 3 hours ago.
But, currently it does not work for the first time.

This problem seems regression after f3f559c

Also, after the installation of Scoop, last update timestamp should be set like `scoop config lastupdate [System.DateTime]::Now.ToString('o')`.

PR #2907, opened 2018-12-16
Fix. Scoop does not update itself automatically for the first time.

nimzo6689:
Fixed. #2906

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -595,15 +595,12 @@ function show_app($app, $bucket, $version) {
 
 function last_scoop_update() {
     $last_update = (scoop config lastupdate)
-    if(!$last_update) {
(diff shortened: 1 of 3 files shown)


lib/core.ps1
```diff
@@ -595,15 +595,12 @@ function show_app($app, $bucket, $version) {
 
 function last_scoop_update() {
     $last_update = (scoop config lastupdate)
-    if(!$last_update) {
-        $last_update = [System.DateTime]::Now
-    }
 
-    if($last_update.GetType() -eq [System.String]) {
+    if ($null -ne $last_update -and $last_update.GetType() -eq [System.String]) {
```
excitoon: Without L598-L600 we don't need testing for string.

Issue #2693, opened 2018-10-19
Path 'pollution' is not working


vsDizzy:
According to the docs, scoop should not pollute environment Path, but it does so aggressively.
![image](https://user-images.githubusercontent.com/5798340/47238701-adb69700-d3eb-11e8-889f-04e0944a3299.png)

Ash258:
That depends on manifest. Some apps need to use own registry variables and path extensions to work.

https://github.com/lukesampson/scoop/blob/dd12ec482c8dfa9d2b617e46ff9ceb485471be84/bucket/nodejs.json#L21-L24

Most of manifest use only `\shims`.

fujii:
I don't want to put installed apps in my PATH.
I'd like to suggest a new command 'scoop shell' to start a PowerShell with PATHs of all installed apps, or 'scoop env' to dump all PATHs.

bcatubig:
That idea sounds familiar to how `Pipenv` works. I'm not sure what type of refactoring that would include, but does seem interesting.

MunsterPlop:
That's actually becoming an issue for me. I installed quite a few softwares with scoop and lately I keep hitting the user PATH variable characters limit (2047), which just ends up truncating itself whenever I try to install something new, sometimes even messing up the shims path.

kbytesys:
I've the same problem, when the PATH variable go over the 2047 limit everything is screwed up and I've to remove items manually. I see that the issue type is "enhancement", but is a real bug for me because this breaks my system.

rashil2000:
#1813

PR #2917, opened 2018-12-20
Private environment variables in shims

excitoon:
Hi.

I'd like to suggest a feature - possibility to specify environment variables right in shim.
It will allow to not pollute system variables with application-specific data and make ideally isolated application packages.

For example, I wish to install Python 2.7 and allow user to use his own Python installation as well (many packages like Anaconda, MSVC, Jupyter install Python as part of itself). In this case, I make following shims:

`pip2.shim`:
```
path = C:\Users\Chebotarev_V\scoop\apps\python27\current\python.exe
args = -m pip
env::PYTHONHOME = C:\Users\Chebotarev_V\scoop\apps\python27\current\
```

`python2.shim`:
```
path = C:\Users\Chebotarev_V\scoop\apps\python27\current\python.exe
env::PYTHONHOME = C:\Users\Chebotarev_V\scoop\apps\python27\current\
```

And depending on executable I call, my packages will be kept completely intact by foreign Python. I can install two versions of Python to the same user and be sure everything is fine.

I hope you get the idea, some applications can have non-compatible requirements for environment variables and this patch enables possibility to use them.

Suggestions are welcome.


excitoon:
One more example:
```
path = python3
args = -c "import os; print(os.environ)"
env::Path = my_custom_path;%Path%
```

excitoon:
Also, I want to note, that shim is a critical piece of Scoop. I guess we should have tests somewhere in separate PR.

excitoon:
Also my patch will allow to fix this #2693 .

supporting/shimexe/shim.cs
```diff
@@ -164,5 +183,31 @@ internal struct PROCESS_INFORMATION {
             }
             return config;
         }
+
+        private static string ExpandEnvironmentVariables(IDictionary environment_dictionary, string name)
```
r15ch13: [`Environment.ExpandEnvironmentVariables()`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.expandenvironmentvariables) already exists in .NET
How is this implementation different?

supporting/shimexe/shim.cs
```diff
@@ -91,10 +95,25 @@ internal struct PROCESS_INFORMATION {
             if(!string.IsNullOrEmpty(cmd_args)) cmd_args = " " + cmd_args;
             var cmd = "\"" + path + "\"" + cmd_args;
 
+            var environment_dictionary = Environment.GetEnvironmentVariables();
+            foreach (KeyValuePair<string, string> entry in config) {
+                if (entry.Key.StartsWith("env::")) {
+                    environment_dictionary[entry.Key.Substring(5)] = ExpandEnvironmentVariables(environment_dictionary, entry.Value);
```
r15ch13: Wouldn't [`Environment.SetEnvironmentVariable()`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.setenvironmentvariable) work in this case?

supporting/shimexe/shim.cs
excitoon: Existing implementation works with current environment in place and mine expands name using local copy of environment from dictionary. Theoretically, I could go another way, change current environment because we run only one program and then exit, but current approach seems more clear to me. Should I change it?

excitoon:
@r15ch13 Please, let's make JSON in separate PR if needed.

supporting/shimexe/shim.cs
excitoon: I guess it would. Should I move to it?

supporting/shimexe/shim.cs
excitoon: Well, I change it. Actually it is the only case where we spoil current environment. Funny, but I never did it earlier.

excitoon:
By the way, currently `perl` package pollutes `PATH` in very bad way, overriding default `gcc`.

excitoon:
Hi @r15ch13. Is there anything that prevents us from merging it? I can see that people react in good way about this change.

supporting/shimexe/shim.cs
```diff
@@ -91,6 +91,12 @@ internal struct PROCESS_INFORMATION {
             if(!string.IsNullOrEmpty(cmd_args)) cmd_args = " " + cmd_args;
             var cmd = "\"" + path + "\"" + cmd_args;
 
+            foreach (KeyValuePair<string, string> entry in config) {
+                if (entry.Key.StartsWith("env::")) {
```
r15ch13: Use similar syntax to PowerShell `$env:PATH` > `env:PATH`. Easier to copy and paste :smile:
```suggestion
                if (entry.Key.StartsWith("env:")) {
```

supporting/shimexe/shim.cs
```diff
@@ -91,6 +91,12 @@ internal struct PROCESS_INFORMATION {
...
+                    Environment.SetEnvironmentVariable(entry.Key.Substring(5), Environment.ExpandEnvironmentVariables(entry.Value));
```
r15ch13: Change substring accordingly
```suggestion
                    Environment.SetEnvironmentVariable(entry.Key.Substring(4), Environment.ExpandEnvironmentVariables(entry.Value));
```

r15ch13:
This requires changes to [`shim()`](https://github.com/lukesampson/scoop/blob/7bedd449b5bb4ef6d4a53a9f4aee48db3fb847bf/lib/core.ps1#L389) to function.

supporting/shimexe/shim.cs
excitoon: Okay!

rashil2000:
The shim code in this PR is no longer used

melMass:
Probably not ideal but I overcome my needs by making a ["bin wrapper"](https://github.com/melMass/scoop-mtb/blob/master/bin/generate_bin_wrapper.ps1) to use in [manifests](https://github.com/melMass/scoop-mtb/blob/81a5fe14f9491df5efc4cc1d8c1246c37aa4949e/bucket/gaffer.json#L24-L26)

PR #2918, opened 2018-12-20
Shims for relative paths

excitoon:
Hi.

Can we have a possibility to specify relative paths in shims?
Like this:
```
path = ../apps/<app>/current/app.exe
```

That will allow users to simply move 'scoop' folder from machine to machine or from user to user.
Now we have absolute paths, and this is not much useful.


excitoon:
I've remembered another reason. Making paths relative will allow users to prepare software bundles in order to run in some isolated environments. For example, I want to run some tools in another computer without Internet, and I simply copy `scoop` folder using a flash drive.

excitoon:
Hi. What do you think about this change?

r15ch13:
Sorry for the late review.

There is no need to call `GetDirectoryName()` and `GetExecutingAssembly()` because the results are already available from line 67-69.

https://github.com/lukesampson/scoop/blob/0215c7bc0b3ce542162f078b4cf303238ae719b7/supporting/shimexe/shim.cs#L67-L69

Therefore it's enough to call the following. In addition it would be nice to check if the file actually exists.
```c#
            if (!System.IO.Path.IsPathRooted(path)) {
                // allow relative path to exe
                path = Path.Combine(dir, path);
            }

            if(!File.Exists(path)) {
                Console.Error.WriteLine("Couldn't find " + Path.GetFileName(path) + " in " + Path.GetDirectoryName(path));
                return 1;
            }
```

puxu-msft:
I think it's better to introduce a environment variable substitution here instead of fixed relative paths.

Issue #1606, opened 2017-07-18
Support shims to both console and GUI apps


prasannavl:
When using any tools, like `sysinternals` for example, an empty command window (the shim) is open, and has to be manually closed. 

The ideal way would be that the shim itself be a "win32" app. Currently all shims are console programs. i.e, not linked as "win32" apps for gui mode.

EDIT:
Suggestion, allow for properties in the manifest to configure which binaries are gui shims, and which should be commandline shims, and use shims accordingly.

SimplyKnownAsG:
would it be possible to add a flag in the json (then in the .shim) indicating that shim.exe should not wait indefinitely? Something like

```C#
bool fork = bool.Parse(Get("fork", config) ?? "false");
if(fork) {
    return 0;
}
else {
    WaitForSingleObject(pi.hProcess, INFINITE);

    uint exit_code = 0;
    GetExitCodeProcess(pi.hProcess, out exit_code);

    // Close process and thread handles.
    CloseHandle(pi.hProcess);
    CloseHandle(pi.hThread);

    return (int)exit_code;
}
```

I'm not sure how one might specify this in the `app.json`, and currently it looks like `function shim` within `core.ps1` would need to accept another argument. It might make sense (for future concerns) to allow `$arg` to be a dict of key/vals that can be dumped to the `.shim` file for loading. Not to get ahead of this current problem.

SimplyKnownAsG:
@r15ch13 is there a way you think this should be accomplished? i'd be happy to close #2006 if there were another way around it. I suppose we could also change GUIs to just add the path rather than creating a shim. is there a preference of adding to path vs creating a shim?

Congee:
A symbolic link would suffice. BTW, why do we need a `shim.exe`?

rasa:
A symbolic link doesn't work, as the exe that is the symlink's target cannot find its related .dlls, as the "exe's directory" is the directory the shim is in, not where the target lives.

SimplyKnownAsG:
Windows isn't very friendly about symbolic links, but hard links are more accessible for those who do not have admin rights. However, I think that @rasa is correct about `.dll`s.

@Congee, I'm not sure I understand the question:
> why do we need a `shim.exe`?

I think the intent of `scoop/shims/*.exe` is used to help prevent pollution of the `PATH`. `shim.exe` is a fancy/funky symlink

fritzmg:
> > why do we need a `shim.exe`?
> 
[...]

PR #2921, opened 2018-12-20
Wait for console applications only

excitoon:
Follow up for #1606 .
Possibly as an example and possibly as a real solution. This code should be pretty fast because at the moment of execution application is already read and mapped into virtual memory.
But we might want it to be run at install time and pass wait flag as an argument, I don't know.


r15ch13:
Neat! 😮😍

excitoon:
I guess we should not merge it. I've just understood that some people open programs from Win+R. In this case they would have flickering black window of shim if they run GUI application. I guess we should rewrite this to PowerShell, and run while shimming, setting a flag in `.shim` like `gui = 0/1`...

rasa:
Would it help if we provided a “GUI shim” to call GUI exes, by setting `/target:winexe` in [build.ps1](
https://github.com/lukesampson/scoop/blob/11a37f44ac995484ac28fc60893c27718c46df6a/supporting/shimexe/build.ps1#L9), per [here](
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/target-compiler-option)?

excitoon:
Finally, I decided **not to skip waiting in GUI mode** because some programs may rely on exit time. For example `git rebase -i` relies on editor (which can be GUI).

rasa:
This issue should be fixed in #3998, which was merged into master. To use, type:
```
scoop config shim kiennq
scoop reset *
```
or
```
scoop config shim 71
scoop reset *
```
Eventually, one of those shims (probably kiennq) we be made the default. 

If this issue is still a problem after trying both solutions above, please reopen this issue. Thanks!

excitoon:
@rasa Does it really need 2 years to answer that it is already solved? By the way, alt shims do not allow to automatically determite application subsystem, so you are wrong in that comment.

rasa:
@excitoon My apologies for closing this. I erroneously thought this issue was fixed by the new shims. Thank you for pointing out that I was mistaken, and for developing this solution.

rasa:
I resolved the conflicts, but more work needs to be done before this can be merged, so I am going to convert to a draft.

rasa:
@excitoon Since it was my screw up, I will take on this PR, and move it forward, if that's OK with you. Thanks again for your help.

excitoon:
@rasa okay, no problem, I'm actually not good in PowerShell, anyway

excitoon:
@rasa I just thought we can add a shim for running some applications (16-bit MZ) in DosBox. That is the thing Chocolatey definitely miss.

excitoon:
Hi @rasa . Is there any news? I noticed modern shims can't be killed with Ctrl+C, this is quite inconvenient if you run desktop apps from console.

rashil2000:
Merged in #5559

PR #2922, opened 2018-12-21
Adjust checkver for rust and rust-msvc autoupdate

hsiktas:
My first little commit here. Please check that I did not miss anything important. ;-)


r15ch13:
Thanks 👍

PR #2925, opened 2018-12-24
let scoop update root path in settings.txt

Retia-Adolf:
Added post_install script, auto updates settings.txt like:
from `root: C:\ProgramData\scoop\persist\nvm\nodejs`
to `root: C:\Users\Retia\scoop\persist\nvm\nodejs`
(only edits this one, try to keep others the same)
else if persist dir is moved, it keeps using old settings which is correct, but while old path is used too.

### DIFF
--- a/bucket/nvm.json
+++ b/bucket/nvm.json
@@ -1,12 +1,27 @@
 {
-    "homepage": "https://github.com/coreybutler/nvm-windows",
     "version": "1.1.7",
+    "description": "A node.js version management utility for Windows",
+    "homepage": "https://github.com/coreybutler/nvm-windows",
+    "license": "MIT",
+    "notes": "You'll need to restart powershell/cmd to have it reload Environment Variables so nvm will work correctly",
     "url": "https://github.com/coreybutler/nvm-windows/releases/download/1.1.7/nvm-noinstall.zip",
-    "bin": [
-        "nvm.exe"
+    "hash": "md5:00acf86e40c5f038cca8c383b9d2c207",
+    "pre_install": [
+        "# $architecture variable from lib/install.ps1 scope",
+        "$valueArch = $architecture -replace '(\\d+)bit', '$1'",
+        "if(Test-Path \"$persist_dir\\settings.txt\") {",
+        "    $valueProxy = (Get-Content \"$persist_dir\\settings.txt\") -match('proxy:\\s+') -replace 'proxy:\\s+', ''",
+        "} else {",
+        "    $valueProxy = 'none'",
+        "}",
+        "$content = \"root: $persist_dir\\nodejs`r`n\"",
+        "$content += \"arch: $valueArch`r`n\"",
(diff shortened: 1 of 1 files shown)


Ash258:
In pre_install there is already `root` set to `$persist_dir\\nodejs`. You will edit that settings file again and replace it with same value. 

pre_install:  ``\"root: $persist_dir\\nodejs`r`n``
your post_install: ``\"root: $persist_dir\\nodejs`r`n\"``

What happened in your case (if I could guess) is that you installed nvm as global app and then move that globally persisted file into local persist_dir. This use case is pretty single user targetted. This change should be done by you manually, because when you did this change in nvm, you should change it everywhere.

Here is linted manifest with no need for architecture specific pre_install: 

```json
{
    "version": "1.1.7",
    "description": "A node.js version management utility for Windows",
    "homepage": "https://github.com/coreybutler/nvm-windows",
    "license": "MIT",
    "notes": "You'll need to restart powershell/cmd to have it reload Environment Variables so nvm will work correctly",
    "url": "https://github.com/coreybutler/nvm-windows/releases/download/1.1.7/nvm-noinstall.zip",
    "hash": "md5:00acf86e40c5f038cca8c383b9d2c207",
    "pre_install": [
        "    # $architecture variable from lib/install.ps1 context",
        "$valueArch = $architecture -replace '(\\d+)bit', '$1'",
        "if (!(Test-Path \"$dir\\settings.txt\")) {",
        "    $content = \"root: $persist_dir\\nodejs`r`n\"",
        "    $content += \"arch: $valueArch`r`n\"",
        "    $content += 'proxy: none'",
        "    Set-Content \"$dir\\settings.txt\" $content -Encoding Ascii",
        "}"
    ],
    "persist": [
        "nodejs",
        "settings.txt",
        [
            "elevate.cmd",
            "nodejs\\elevate.cmd"
        ],
        [
            "elevate.vbs",
            "nodejs\\elevate.vbs"
        ]
    ],
    "env_add_path": "nodejs\\nodejs",
    "env_set": {
        "NVM_HOME": "$dir",
        "NVM_SYMLINK": "$persist_dir\\nodejs\\nodejs"
    },
    "bin": "nvm.exe",
    "checkver": "github",
    "autoupdate": {
        "url": "https://github.com/coreybutler/nvm-windows/releases/download/$version/nvm-noinstall.zip",
        "hash": {
            "url": "$url.checksum.txt",
            "find": "([a-fA-F0-9]{32})"
        }
    }
}
```

Retia-Adolf:
Yes, but I think that is same for other cases like moving persisted file from one user/computer to another, so... is this ok?

Ash258:
When you move settings like this, you should be be aware of them and how to change them. There is no precedence / way how to determine if files have some hardcoded paths, which need to be changed.
Some applications append these settings after first run / settings change in GUI and you should replace all of them, not just one.

rasa:
Stale

PR #2930, opened 2018-12-27
Allowed underscores in package names.

excitoon:
For example: `gcc-x86_64` (this is not host architecture, which can be expressed by `-a`, but target one).

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -610,7 +610,7 @@ function applist($apps, $global) {
 }
 
 function parse_app([string] $app) {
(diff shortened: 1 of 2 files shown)


r15ch13:
Please add a test case to https://github.com/lukesampson/scoop/blob/master/test/Scoop-Core.Tests.ps1#L258

excitoon:
Okay.

excitoon:
@r15ch13 I'm finally finished it. Please check if it is fine.

PR #2934, opened 2018-12-27
Fixed OpenSSH infinite link. #2927

### LINKED ISSUE #2927: OpenSSH links into infinite loop
OpenSSH as installed via scoop has a link directory to the users home dir. Since scoop apps are in the home dir, this becomes an infinite loop.

I noticed it while trying to Windows-Search a file. It was unable to proceed the home directory. So Windows-Search is unusable.

scoop

excitoon:
Fixes #2927

### DIFF
--- a/bucket/openssh.json
+++ b/bucket/openssh.json
@@ -61,10 +61,10 @@
         "usr/bin/sshd.exe"
     ],
     "pre_install": [
-        "mkdir $dir/tmp",
-        "mkdir $dir/home",
-        "mkdir $dir/home/$env:UserName",
-        "mkdir -Force $env:UserProfile/.ssh",
-        "cmd /c mklink /j $dir\\home\\%UserName%\\.ssh %UserProfile%\\.ssh"
+        "mkdir \"$dir/tmp\"",
+        "mkdir \"$dir/home\"",
+        "mkdir \"$dir/home/$env:UserName\"",
+        "mkdir -Force \"$env:UserProfile/.ssh\"",
+        "cmd /c mklink /j \"$dir\\home\\%UserName%\\.ssh\" \"%UserProfile%\\.ssh\""
     ]
 }
--- a/bucket/openssh.json
+++ b/bucket/openssh.json
@@ -63,6 +63,8 @@
     "pre_install": [
         "mkdir $dir/tmp",
         "mkdir $dir/home",
-        "cmd /c mklink /j $dir\\home\\%USERNAME% %USERPROFILE%"
+        "mkdir $dir/home/$env:UserName",
+        "mkdir -Force $env:UserProfile/.ssh",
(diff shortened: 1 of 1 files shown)


bucket/openssh.json
```diff
@@ -63,6 +63,8 @@
     "pre_install": [
         "mkdir $dir/tmp",
         "mkdir $dir/home",
-        "cmd /c mklink /j $dir\\home\\%USERNAME% %USERPROFILE%"
+        "mkdir $dir/home/$env:USERNAME",
```
Ash258: Use quotes around path with $dir to prevent spaces.

bucket/openssh.json
```diff
@@ -63,6 +63,8 @@
     "pre_install": [
         "mkdir $dir/tmp",
         "mkdir $dir/home",
-        "cmd /c mklink /j $dir\\home\\%USERNAME% %USERPROFILE%"
+        "mkdir $dir/home/$env:UserName",
```
excitoon: Where have you been when @rasa made this PR? :)))) @Ash258

Ash258:
Should not there be some `uninstaller.script` property, which unlink that junction?

bucket/openssh.json
excitoon: I guess I should make it around all paths.

excitoon:
@Ash258 I have no idea. I guess simple delete shall understand that directory junction is something to be unlinked.

excitoon:
Personally, I think it is ready to merge. One can work on 32-bit versions for OpenSSH.
I can't properly test it because for some reason I have fresh OpenSSH in my system:
```
C:\Users\vladi\Documents\src\scoop-user>where ssh
C:\Windows\System32\OpenSSH\ssh.exe

C:\Users\vladi\Documents\src\scoop-user>ssh -V
OpenSSH_for_Windows_7.7p1, LibreSSL 2.6.5
```

I guess it is Windows Update.

bucket/openssh.json
```diff
@@ -61,8 +61,10 @@
         "usr/bin/sshd.exe"
     ],
     "pre_install": [
-        "mkdir $dir/tmp",
-        "mkdir $dir/home",
-        "cmd /c mklink /j $dir\\home\\%USERNAME% %USERPROFILE%"
+        "mkdir \"$dir/tmp\"",
+        "mkdir \"$dir/home\"",
+        "mkdir \"$dir/home/$env:UserName\"",
+        "mkdir -Force \"$env:UserProfile/.ssh\"",
```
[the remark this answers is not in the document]
excitoon: Because first entries do not exist and last one may already exist.

rasa:
Should we consider instead of using msys2's ssh, we use the ssh built into git-for-windows?  See https://github.com/rasa/scoop/blob/01b832f0d9948481d7125de49191b336a1c4e6f1/bucket/openssh.json

excitoon:
@r15ch13

PR #2937, opened 2018-12-28
Add Microsoft R Open v3.5.1

niheaven:
Add MRO v3.5.1, an enhanced distribution of R from Microsoft Corporation.

- Name: Microsoft R Open
- Description: Microsoft R Open is the enhanced distribution of R from Microsoft Corporation. The current release, Microsoft R Open 3.5.1, is based the statistical language R-3.5.1 and includes additional capabilities for improved performance, reproducibility and platform support.
- Version: 3.5.1
- Homepage: https://mran.microsoft.com

### DIFF
--- a/bucket/mro.json
+++ b/bucket/mro.json
@@ -1,0 +1,46 @@
+{
+    "version": "3.5.1",
+    "description": "Microsoft R Open: The enhanced distribution of R from Microsoft Corporation",
+    "homepage": "https://mran.microsoft.com/",
+    "license": "GPL-2.0-only",
+    "depends": [
+        "wixtoolset",
+        "extras/vcredist2015"
+    ],
+    "url": "https://mran.blob.core.windows.net/install/mro/3.5.1/microsoft-r-open-3.5.1.exe#/mro.exe",
+    "hash": "30de2b2fefed091452695e03c805fcefcef00ab30b3df04363b72d9d44bac494",
+    "installer": {
+        "script":[
+            "&(scoop which dark) -nologo -x \"$dir\\tmp\" \"$dir\\mro.exe\" | Out-Null",
+            "extract_msi \"$dir\\tmp\\AttachedContainer\\ROpen.msi\" \"$dir\\tmp\" | Out-Null",
+            "Move-Item \"$dir\\tmp\\Microsoft\\MRO-$version.0\\*\" \"$dir\" -Force",
+            "extract_7zip \"$dir\\Setup\\*.cab\" \"$dir\\Setup\"",
+            "Remove-Item \"$dir\\bin\\x64\\Rblas.dll\", \"$dir\\bin\\x64\\Rlapack.dll\" -Force",
+            "Move-Item \"$dir\\Setup\\bin\\x64\\*\" \"$dir\\bin\\x64\" -Force",
+            "Move-Item \"$dir\\Setup\\library\\*\" \"$dir\\library\" -Force",
+            "Remove-Item \"$dir\\tmp\", \"$dir\\Setup\", \"$dir\\mro.exe\" -Force -Recurse"
+        ]
+    },
+    "bin": [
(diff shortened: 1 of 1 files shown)


Ash258:
- It's 64bit only. Wrap all installation steps into `architecture.64bit` property (url, hash, installer, bin) 
- Add hash extraction from: https://mran.microsoft.com/download
    - It looks consistent between releases, so something like `find: "SHA256: ([A-Fa-f\\d]{64})"` should be enough

niheaven:
- Wrap every installation steps into arch.64bit.
- Since https://mran.microsoft.com use ArgularJS which I'm not familar with to build the main body, the simple way that find: "SHA 256: ([A-Fa-f\\d]{64})" didn't work on download page. And that is also why I try to checkver on GitHub instead of homepage of MRO.
- The SHA256 must be on the Azure storage, and I'll check the js to find them. But now, using 'extract' method as an alternative.

niheaven:
Sorry for the reverse mistake. The newest one is okay.

Since https://github.com/Microsoft/microsoft-r-open/releases use MRO-x.x.x as tags, the checkver.github doesn't work for this type, so checkver.url and a regex are used.

Ash258:
regex and github could be combined.

niheaven:
The autoupdate.hash is extract from download [api](https://mranapi.azurewebsites.net/api/download), and checkver is moved to [app.config](https://mran.microsoft.com/assets/configurations/app.config.json). Now the manifest is ready to merge.

There is a question that `$.platforms[0].versions[(@.length-1)].sha256` cannot work for sha256 extraction, but `$.platforms[0].versions[21].sha256` do, i.e., `(@.length-1)` got a error. Regex works perfectly here, but why jsonpath can't?

Ash258:
`$...platforms[?(@.name = 'windows')].versions[?(@.versionInfo == '3.5.1' && @.downloadText == 'microsoft-r-open-3.5.1.exe')].sha256` should do the trick.

bucket/mro.json
```diff
@@ -0,0 +1,59 @@
+{
+    "version": "3.5.1",
+    "description": "Microsoft R Open: The enhanced distribution of R from Microsoft Corporation",
+    "homepage": "https://mran.microsoft.com/",
+    "license": "GPL-2.0-only",
+    "depends": [
+        "wixtoolset",
+        "extras/vcredist2015"
```
Ash258: Remove vcredist from depends and move it to suggest.

niheaven:
`$..versions[?(@.downloadText == 'microsoft-r-open-3.5.1.exe')].sha256` work at http://jsonpath.com/, but doesn't work in `chechver -u`. The code is

```json
"hash": {
    "mode": "json",
    "jsonpath": "$..versions[?(@.downloadText == '$basename.exe')].sha256",
    "url": "https://mranapi.azurewebsites.net/api/download"
}
```

As mentioned before, if jsonpath is set to `"$.platforms[0].versions[21].sha256"`, the hash is extracted correctly. That's why regex is more useful here.

Ash258:
I tested it, but could not debug it with some good result. Keep it as it and i will look into it after.

bucket/mro.json
niheaven: VC Runtime 2015 is a must-install component during MRO installation (the checkbox is gray), so vcredist2015 is put to `depends` instead of `suggest`.
mro

niheaven: vcredist2015 is under `suggest` now.

niheaven:
regex is just okay and let's forget jsonpath :tada:

Ash258:
It's deeper problem. But it should be done in core and then change it here after fixing.

niheaven:
JSONPath error is solved by #3011 and issue about `@.length` is mentioned in JamesNK/Newtonsoft.Json#1318.

PR #2957, opened 2019-01-06
Add mach2 0.3.0.0

patrick330602:
This is a feature control multi-tool for Windows 10.


Ash258:
There is no sha256 files. Delete hash extraction property. (Same for all your PRs, except doctl)

PR #2969, opened 2019-01-09
Redirect "global" yarn directory

milang:
Current installer for `yarn` redirects `bin`, `cache` and `mirror` directories to scoop's persistent storage. However, the `global` directory (used by packages that get installed globally via `yarn global add`) is still using the default `$env:LOCALAPPDATA\Yarn\Data\global`.

This pull request adds a persistent directory `global` and then updates `.yarnrc` to use it.

Unfortunately `yarn` does not currently recognize `global-folder` configuration setting, only `--global-folder` ([source](https://github.com/yarnpkg/yarn/issues/5746#issuecomment-386295014)), so I added an extra post-install PowerShell fragment that adds the `--` prefix to `global-folder` in `.yarnrc` file.

The following screenshot illustrates the difference in `yarn global dir` output when using installing the current `yarn` bundle and when installing updated bundle (as per this PR):

yarn global dir

### DIFF
--- a/bucket/yarn.json
+++ b/bucket/yarn.json
@@ -15,9 +15,12 @@
     "persist": [
         "cache",
         "bin",
-        "mirror"
+        "mirror",
+        "global"
     ],
     "post_install": [
+        "yarn config set global-folder \"$dir\\global\"",
+        "Set-Content $env:USERPROFILE\\.yarnrc (Get-Content $env:USERPROFILE\\.yarnrc | %{ $_ -replace \"^global-folder\", \"--global-folder\" })",
         "yarn config set cache-folder \"$dir\\cache\"",
         "yarn config set yarn-offline-mirror \"$dir\\mirror\"",
         "yarn config set prefix \"$dir\""


milang:
@r15ch13 any comments?

eteeselink:
@milang I'm just a random guy from the internet. I like Scoop and I noticed there's many PRs so I figured I could spend a few minutes helping with code reviews. I have no Scoop authority whatsoever.

I think the code behind this change is good, but I'm not sure what the practical problem is with Yarn's global directory remaining in `$env:LOCALAPPDATA\Yarn\Data\global`. Isn't the whole point of it that it remains in place across updates? 

I understand that it would be neater if it's all managed in scoop's persistent storage, but it also adds complexity. Did you write this just for neatness, or is there a practical problem with Yarn managing its own global storage behind Scoop's back?

h404bi:
That's a matter of choice, it's neat that redirecting it to scoop's persistent storage, but as you said, it adds complexity.

I don't care about the problem that some apps do have behavior outside Scoop's folder, so just keeping yarn's default `global` location is always a better idea to me.
https://github.com/h404bi/dorado/blob/master/yarn.json#L17-L22

I can't force every Scoop user to do that, so making a variant manifest for yourself is a better choice, that's a highlight feature of Scoop. Make Scoop fit your need!

rasa:
This change seems reasonable, but I don’t use yarn. Are there any downsides to merging this in?

eteeselink:
I see one downside and one risk:
- downside: it adds complexity
- risk: will this work correctly across upgrades? people who upgrade yarn will already have a yarn global folder outside the scoop tree. i feel like it'd need a lot of careful testing.

Finally, while I can see that this PR has been a lot of work to make, and the code itself looks fine, I still don't know what problem it solves. If it's just the aesthetics of not having a Yarn folder outside the Scoop tree then I'm against merging this in for the above-mentioned downside and risk. But maybe there's a different problem that I fail to see. @milang can you chip in? 

(note: i am just a guy from the internet, my opinion should not weigh particularly heavily)

niheaven:
I've made a PR in upstream (https://github.com/yarnpkg/yarn/pull/7056) that add `global-folder` support in `.yarnrc`, and since yarn global binshim has some problems and is not fixed recently, I suggest use `global-folder\node_modules\.bin` instead.

When the upstream PR is not merged, maybe a hardcoded fix to yarn is better then hardcoded fix to install script in scoop manifest. I'll try to make one (a simple -replace doesn't work).

milang:
Thank you @niheaven, your fix is much better.

PR #2982, opened 2019-01-13
pshazz: Remove askpass.exe from bin

niheaven:
`pshazz` remove askpass.exe from https://github.com/lukesampson/pshazz/commit/f40bfabb011e38215f66610e5dce432fa67ba393


niheaven:
#2978 has fixed it.

PR #2983, opened 2019-01-14
Add sops 3.2.0

goostleek:
https://github.com/mozilla/sops

### DIFF
--- a/bucket/sops.json
+++ b/bucket/sops.json
@@ -1,9 +1,9 @@
 {
     "homepage": "https://github.com/mozilla/sops",
-    "description": "SOPS: Secrets OPerationS",
+    "description": "Editor of encrypted files that supports YAML, JSON, ENV, INI and BINARY formats and encrypts with AWS KMS, GCP KMS, Azure Key Vault and PGP",
     "license": "MPL-2.0",
     "version": "3.2.0",
-    "depends": "gnupg",
+    "depends": "gpg",
     "architecture": {
         "64bit": {
             "url": "https://github.com/mozilla/sops/releases/download/3.2.0/sops-3.2.0.exe#/sops.exe",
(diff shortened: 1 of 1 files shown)


bucket/sops.json
```diff
@@ -0,0 +1,22 @@
+{
+    "homepage": "https://github.com/mozilla/sops",
+    "description": "SOPS: Secrets OPerationS",
+    "license": "MPL-2.0",
+    "version": "3.2.0",
+    "depends": "gnupg",
```
Ash258: Where is mentioned you need to have gnupg installed??

goostleek: It's not mentioned directly but it won't work w/o `gpg` binary. If you look at the documentation [§2.9 Specify a different GPG executable](https://github.com/mozilla/sops#29specify-a-different-gpg-executable) you can see
> `sops `checks for the `SOPS_GPG_EXEC` environment variable. If specified, it will attempt to use the executable set there instead of the default of **`gpg`.**

I have changed the dependency from `gnupg` to `gpg` to be more straightforward because [`gnupg`](https://github.com/lukesampson/scoop/blob/eef925fe6e7102cbc8679fb69f0a9a3ef2d4ea6c/bucket/gnupg.json) and [`gpg`](https://github.com/lukesampson/scoop/blob/eef925fe6e7102cbc8679fb69f0a9a3ef2d4ea6c/bucket/gpg.json) manifests are identical.

bucket/sops.json
```diff
@@ -0,0 +1,22 @@
+{
+    "homepage": "https://github.com/mozilla/sops",
+    "description": "SOPS: Secrets OPerationS",
```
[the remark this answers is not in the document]
goostleek: Fixed.

PR #2988, opened 2019-01-15
Add smimesign 0.0.10

goostleek:
https://github.com/github/smimesign


Ash258:
1. Remove hardcoded hash from autoupdate.
1. You can't rename `.zip` to `.exe`.
    1. Why would you do it, when that zip only contains single exe binary?
    1. Just keep it zip. It will be automatically unzipped.

goostleek:
All the suggested fixes applied @Ash258.

PR #2995, opened 2019-01-16
Add dotcover 2018.3.1

snowypowers:


r15ch13:
Please use the [jetbrains bucket](https://github.com/Ash258/Scoop-JetBrains) by running `scoop bucket add jetbrains`

snowypowers:
Awesome! Thanks for the work! I have verified that the bucket works, the install went as planned and I was able to use the tool.

PR #2996, opened 2019-01-16
Add version variables to autoupdate.hash.regex

niheaven:
Add [version variables](https://github.com/lukesampson/scoop/wiki/App-Manifest-Autoupdate#version-variables) to `autoupdate.hash.regex` for some hash extract situations.

The idea is from Atom's nupkg hash extraction. Atom's [releases page](https://github.com/atom/atom/releases/latest) provides nupkg named `atom-1.34.0-full.nupkg` (for i686) and `atom-x64-1.34.0-full.nupkg` (for x64_86), and it also provides [`RELEASES`](https://github.com/atom/atom/releases/download/v1.34.0/RELEASES) and [`RELEASES-x64`](https://github.com/atom/atom/releases/download/v1.34.0/RELEASES-x64) files that give sha1 hash of nupkg files. Unfortunately in `RELEASES-x64` the nupkg file name is `atom-1.34.0-full.nupkg` without `-x64`. 

I haven't found some method to write a regex to extract the hash without `$version`, but [`get_hash_for_app()`](../blob/master/lib/autoupdate.ps1#L133) only support `$basename$ in regex. So I add the code, and it works for the following hash extract:

```json
"url": "https://github.com/atom/atom/releases/download/v$version/atom-x64-$version-full.nupkg#dl.7z",
"hash": {
    "url": "$baseurl/RELEASES-x64",
    "regex": "([a-fA-F0-9]+)\\s+?(?:atom-$version-full.nupkg)"
}
```

### DIFF
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -9,7 +9,7 @@ TODO
 . "$psscriptroot/core.ps1"
 . "$psscriptroot/json.ps1"
 
-function find_hash_in_rdf([String] $url, [String] $filename) {
+function find_hash_in_rdf([String] $url, [String] $basename) {
     $data = $null
     try {
         # Download and parse RDF XML file
@@ -24,7 +24,7 @@ function find_hash_in_rdf([String] $url, [String] $filename) {
     }
 
     # Find file content
-    $digest = $data.RDF.Content | Where-Object { [String]$_.about -eq $filename }
+    $digest = $data.RDF.Content | Where-Object { [String]$_.about -eq $basename }
 
     return format_hash $digest.sha256
 }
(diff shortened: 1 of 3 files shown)


r15ch13:
How about adding a `$regexEscape` parameter to substitute:
```powershell
function substitute($entity, [Hashtable] $params, [Bool]$regexEscape = $false) {
    if ($entity -is [Array]) {
        return $entity | ForEach-Object { substitute $_ $params $regexEscape}
    } elseif ($entity -is [String]) {
        $params.GetEnumerator() | ForEach-Object {
            if($regexEscape -eq $false) {
                $entity = $entity.Replace($_.Name, $_.Value)
            } else {
                $entity = $entity.Replace($_.Name, [Regex]::Escape($_.Value))
            }
        }
        return $entity
    }
}
```

And instead of using `$basename`-Parameter for `find_hash_in_textfile()` we pass down the `$substitutions`. (requires adding the `$basename` to the hashtable)

niheaven:
Thanks @r15ch13, I've just made a minimal change to achieve my goal, and your implement are more systemic and readable. I've adopted your advise and extended it to JSONPath (though may not needed now).

I've tested the code on atom's nupkg version (for regex), nuget (for jsonpath checkver) and openssl (for jsonpath hash extraction), everything is okay, the new logic works well and the old logic is not affected.

It's good for me now, and if merged, I'll submit a pr about new hash extract method of atom.

PR #3009, opened 2019-01-18
mongodb: Add persist and fix bin

niheaven:
- Persist `data`, `log` and `bin\mongod.cfg`
- Modify `bin\mongod.cfg` to use persisted `data` and `log` dir
- Modigy `bin\mongod.exe` shim to use `bin\mongod.cfg` by default
- Add notes about prerequisties and config file useage


niheaven:
It's been modified. @Ash258

PR #3022, opened 2019-01-23
core.ps1: Fix url_remote_filename (edited)

niheaven:
EDITED: No WebRequest, Local URL Treatment

Current `url_remote_filename()` cannot handle redirected urls like "https://www.apache.org/dyn/closer.cgi?action=download&filename=tomcat/tomcat-9/v9.0.14/bin/apache-tomcat-9.0.14-windows-x64.zip", this modified one works in most cases.

- Get filename from PathAndQuery instead of AbsolutePath
- Throw away leading `?=`
- If got `vxx.xx.xx`, fallback to AbsolutePath
- Use fragment name if all above failed

### DIFF
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -162,7 +162,7 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
     debug $hashfile_url
     if ($hashfile_url) {
         write-host -f DarkYellow 'Searching hash for ' -NoNewline
-        write-host -f Green $(url_remote_filename $url) -NoNewline
+        write-host -f Green $basename -NoNewline
         write-host -f DarkYellow ' in ' -NoNewline
         write-host -f Green $hashfile_url
     }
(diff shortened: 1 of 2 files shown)


niheaven:
"AppVeyor build failed" since there is not a "foo.txt" file in example.org...

r15ch13:
The intent of this function is to get the filename from the URL alone, not create a new web request.
The filename for the Tomcat URL is `closer.cgi`, which is technically correct. (`$basename` will be `closer.cgi`)

One way to fix this would be adding a fragment to the URL and make it available via a `$fragment`-variable

`https://www.apache.org/dyn/closer.cgi?action=download&filename=tomcat/tomcat-$majorVersion/v$version/bin/apache-tomcat-$version-windows-x64.zip#/apache-tomcat-$version-windows-x64.zip`

niheaven:
Yes, but the REAL filename of the download url and what is wanted is `apache-tomcat-$version-windows-x64.zip`. 

If no webrequest, it can be extracted by

```powershell
Split-Path (New-Object URI $url).PathAndQuery -Leaf
```

That is, use `URI.PathAndQuery` instead of `AbsolutePath` to extract last filename. And the result is the same while there is no query.

Above method could handle most of cases that use a query, and for some special cases a special treatment could be set., e.g.

https://go.microsoft.com/fwlink/?linkid=843454#/dl.7z （from dotnet.json)
https://drive.google.com/uc?id=0B3X9GlR6EmbnbnBsTXlfS1J5UjQ&export=download#/gdrive.exe (from gdrive.json)
https://chocolatey.org/api/v2/package/ScriptCs/0.17.1?fn=/dl.zip (from scriptcs.json)

A new commit has submit that only use local url treatment, no webrequest.

niheaven:
@r15ch13 Please review the edited function, I've change it to only local treatment.

PR #3026, opened 2019-01-23
[auto-pr] Use regex instead of Starts|EndsWith

Ash258:
On my end result of `hub status --porcelain -uno` is prefixed with space on left for some reason, which failed test for `startsWith('M  ')` since it's strict match

Auto-PR

Alternatively $status could be Trimed before condition

Trimed


r15ch13:
Please use `\x20` to only allow spaces. We don't want to match other whitespaces.

r15ch13:
This was actually intended ... Don't ask me why I merged it.
See https://git-scm.com/docs/git-status#_short_format

PR #3037, opened 2019-01-26
watchexec: remove extract_dir

sh4869221b:


Ash258:
Remove it from autoupdate too.

Issue #3003, opened 2019-01-17
Extract supportings source code into custom repositories


Ash258:
To keep this repository cleaner and more relate, it should be shrinked from not needed files (not talking about bucket folder, that's tracked somewhere else). Supporting folder can contains only binaries of each utility, not whole source code, with all project files. (Better management of changes, release page for better updating of them, ...)

Extract validator and shimexe into `scoopinstaller` github organization. (SideNote: Please rename it to use PascalCase or at least first upper character; it looks better)

Some kind of script for updating them will be required, but it should be only scoop like "installation" into this supporting folder, which would be executed manually. It could be done on user side, but I can't think about best way how to do it. Best solution I can think of is just keep it inside repository to reduce / eliminate error domain.

I would really appreciate this extraction / script to handle getting yaml library without some git tree / submodules way.

PR #3066, opened 2019-02-02
[Proof of Concept] Supportings source code extraction

Ash258:
#3003

There is not really need to have 3 projects inside 1 repository for obvious reasons (history, size, project managment, issues / prs / board for each project, code quality of each project ...).

This is just example of pretty easy script which get all needed dependencies for project using `scoop` like way. This could be possible using gitmodules / trees, but not sure, how would this affect normal users.

Since current supportings are projects, which are maintained by scoop's developers, it's prety easy to extract them, create correct flow of develping, pipelining and automating release process (using appveyor on tag to test, build, publish github release, ...)

Since checkver does not return any data, updating would be done after some git status (if manifest was updated) or just assumption, that captured output would match some regex.

Feel free to suggest different way how to get, update dependencies, since this was my initial idea how to handle this.

Manifests (except yaml) are only illustrative, there is need to extract code first and create releases.


Ash258:
Similar way is pretty common. Nice example is vscode's syntaxes. [CSS extension](https://github.com/Microsoft/vscode/tree/master/extensions/css) - They just have cgmanifest, which continas github url (atom's syntaxes, textmate bundles), commit hash and run updates when needed

niheaven:
If #3060 be merged, scoop should have version number, and this type of update (not single DLL, as a whole) should bump its number, maybe separate json is not needed😁

Ash258:
/cc @r15ch13

Could we discuss this change? I would like to start on YAML manifests support again from Wednesday and this would make it more easier.

PR #3068, opened 2019-02-02
[Autoupdate] Substitute $hash variables for easier maintaining

Ash258:
There is no need to always write regex for each hash type, when they could be automatically substituted.

Now it's possible to use something like:

```json
"autoupdate": {
    "hash": "MD5:\\s*($hashMD5)",
    "hash": "SHA1:\\s*($hashSHA1)",
    "hash": "SHA256:\\s*($hashSHA256)",
    "hash": "SHA512:\\s*($hashSHA512)",
    "hash": "base64encodedHash:\\s*($hashBASE64)"
}
```

Best example i found are base64 hashes, because it's regex is pretty overkill. Rambox pro for example:

```json
{
    "autoupdate": {
        "url": "https://github.com/ramboxapp/download/releases/download/v$version/RamboxPro-$version-win.exe#/cosi.7z",
        "hash": {
            "url": "https://github.com/ramboxapp/download/releases/download/v$version/latest.yml",
            "find": "sha512:\\s+($hashBASE64)"
        }
    }
}
```

Proof

### DIFF
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -1,13 +1,10 @@
 <#
 TODO
- - add a github release autoupdate type
  - tests (single arch, without hashes etc.)
  - clean up
 #>
-. "$psscriptroot\..\lib\json.ps1"
-
-. "$psscriptroot/core.ps1"
-. "$psscriptroot/json.ps1"
+. "$PSScriptRoot\core.ps1"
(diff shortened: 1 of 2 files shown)


niheaven:
Should BASE64 hash type be integrated in common sha512 or sha256 as current function do? That is, test BASE64 in every hash type or just needed ones. Just for convenience.

PR #3069, opened 2019-02-02
Update innounp.json

YenForYang:
Fix autoupdate; use hash extracted from url.


Ash258:
Let this on contributors to release it into binary mirror:

https://github.com/lukesampson/scoop/issues/3047#issuecomment-459610077

SideNote: Sourceforge has automatic hash extraction mode, no need to specify regex.

PR #3098, opened 2019-02-08
Update haxe-dev and support release candidate in checkver Updated hash for hashlink

mundusnine:


bucket/haxe-dev.json
```diff
@@ -30,7 +30,7 @@
     ],
     "checkver": {
         "url": "https://github.com/HaxeFoundation/haxe/releases",
-        "re": "([\\d.]+-preview[\\d.]+)"
+        "re": "(\\d\\.\\d\\.\\d-preview\\.\\d|\\d\\.\\d\\.\\d\\-rc\\.\\d)"
```
Ash258: Why don't go with simple regex?

`([\d\.]+)-(rc|preview)[\d\.]+`

mundusnine: I just started doing regex and I am going with what I can find and I didn't know we could do this so simply. Will change it.

mundusnine: I tried your command and it doesn't work

Ash258: Proof

mundusnine: The one in the pictures isn't the same as the one in your comment. But I will change it to what is in your picture. 
image

Ash258: It's same. It's only escaped to confort json.

PR #3108, opened 2019-02-11
rclone: Add desc and au.hash, persist config, and use download url in main site

niheaven:
- Add `description`
- Use download urls in `downloads.rclone.org` (main site)
- Add `env_set` to persist config in persist_dir of scoop
- Use version file in download site to `checkver`
- Add `autoupdate.hash`

### DIFF
--- a/bucket/rclone.json
+++ b/bucket/rclone.json
@@ -16,9 +16,6 @@
         }
     },
     "bin": "rclone.exe",
-    "env_set": {
-        "XDG_CONFIG_HOME": "$persist_dir\\.."
-    },
     "checkver": {
         "url": "https://downloads.rclone.org/version.txt",
         "regex": "rclone v(.+)"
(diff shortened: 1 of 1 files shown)


r15ch13:
The default value `%XDG_CONFIG_HOME%` is `%HOME%\.config` and should be changed to `persist_dir` of rclone for the whole system. This would probably break other Linux based programs too. (See https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)

niheaven:
Use `$persist_dir\..` to make sure config file is `persist\rclone\rclone.conf` and not `persist\rclone\rclone\rclone.conf`.

But maybe it's better to not set the env var, and let rclone place its config to `%HOME%\.config\rclone`, isn't it?

r15ch13:
Yes, please remove the `env_set` property.

niheaven:
It's okay, and the one in rclone-beta(https://github.com/scoopinstaller/versions/pull/102) is also removed.

PR #3121, opened 2019-02-13
feat(core): support loading basedirs from config

h404bi:
- [x] BLOCKED BY #3242 (MERGED)

Reopen #3116 . ~~This is a breaking change.~~

- ~~revert #3115 , import `config.ps1` in `core.ps1` instead~~
- add support loading env from `~/.config/scoop/config.json` config file, preparing for the new scoop installer (see [scoopinstaller/install#2](https://github.com/scoopinstaller/install/issues/2) and [scoopinstaller/install#4](https://github.com/ScoopInstaller/Install/pull/4))
- drop `$oldscoopdir = "$env:LOCALAPPDATA\scoop"` support, I think nobody use old scoop dir now, it has been 3 years.

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -838,18 +838,7 @@ function get_magic_bytes_pretty($file, $glue = ' ') {
 #       for all communication with api.github.com
 Optimize-SecurityProtocol
 
-# Note: The default directory changed from ~/AppData/Local/scoop to ~/scoop
-#       on 1 Nov, 2016 to work around long paths used by NodeJS.
-#       Old installations should continue to work using the old path.
-#       There is currently no automatic migration path to deal
-#       with updating old installations to the new path.
 $scoopdir = $env:SCOOP, (get_config 'rootPath'), "$env:USERPROFILE\scoop" | Select-Object -first 1
-
-$oldscoopdir = "$env:LOCALAPPDATA\scoop"
-if((test-path $oldscoopdir) -and !$env:SCOOP) {
-    $scoopdir = $oldscoopdir
-}
-
 $globaldir = $env:SCOOP_GLOBAL, (get_config 'globalPath'), "$env:ProgramData\scoop" | Select-Object -first 1
 
 # Note: Setting the SCOOP_CACHE environment variable to use a shared directory
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -1,34 +1,3 @@
-# Note: The default directory changed from ~/AppData/Local/scoop to ~/scoop
-#       on 1 Nov, 2016 to work around long paths used by NodeJS.
-#       Old installations should continue to work using the old path.
-#       There is currently no automatic migration path to deal
-#       with updating old installations to the new path.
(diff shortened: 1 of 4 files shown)


h404bi:
please also add `please-review` label.

Ash258:
I am not against this change in overall, but i like that idea, when you can adress scoop location in other scripts / programs using `$env:SCOOP(_HOME)`. I usually use it a lot and it's more convenient to guide user (for example) using `cd $env:SCOOP` rather than `cd (scoop config rootPath)`.

Should be global path set in `.scoop` file too?? Each users will have to set it on user creation / .. to match other users. Global path as machine environment variable is more suitable I think.

h404bi:
@Ash258 It's ok to keep `$env:SCOOP` for users to adress scoop location. I prefer to get it from `scoop config rootPath` and set it to `$env:SCOOP` in the `scoop runtime` for users, instead of saving it to and getting it from environment variable. Will consider to tweak the implementation.

r15ch13:
Using `$env:` and `config` values is a good solution, but which one has priority over the other?
Keep in mind that this might break [scoop-extras/checkver.ps1](https://github.com/lukesampson/scoop-extras/blob/master/bin/checkver.ps1) and other `bin` scripts.

h404bi:
@r15ch13 
1. I think `config` values should have higher priority, and `$env:SCOOP` is an 'alias' to `scoop config rootPath`.(Just, don't write environment variable if you can, please..) An user can get scoop path from `$env:SCOOP` but can only change the path from `scoop config` (but hey, is it really someone will change the scoop path after installation?)

2. It's ok.
    ```powershell
    f(!$env:SCOOP_HOME) { $env:SCOOP_HOME = resolve-path (split-path (split-path (scoop which scoop))) }
    ```

@Ash258 
> Global path as machine environment variable is more suitable I think.

Agree.

h404bi:
Now the changes is non-invasive, just added support of loading env from `~/.scoop` config file, to reflect the future standalone installer.

h404bi:
Anyone else wants to give a review?

lib/core.ps1
```diff
@@ -1,22 +1,14 @@
-# Note: The default directory changed from ~/AppData/Local/scoop to ~/scoop
-#       on 1 Nov, 2016 to work around long paths used by NodeJS.
-#       Old installations should continue to work using the old path.
-#       There is currently no automatic migration path to deal
-#       with updating old installations to the new path.
-$scoopdir = $env:SCOOP, "$env:USERPROFILE\scoop" | Select-Object -first 1
+. "$PSScriptRoot/config.ps1"
```
r15ch13: This will break the current installer

h404bi: Good catch, didn't considering that the old installer just downloading a single `core.ps1`. It's hard to move on now..

niheaven:
Reviewed and tested without reinstallation of Scoop. LGTM.

If #3242 merged, I will test with reinstallation again.

h404bi:
@niheaven Thanks for respone and test, will rebase the codes after getting #3242 adopted.

PR #3126, opened 2019-02-14
Fix fosshub.com url handler and add fosshub hash mode

niheaven:
- Fix broken fosshub.com url handle function. It can handle urls like `https://www.fosshub.com/Calibre.html/calibre-64bit-3.39.1.msi` or `https://www.fosshub.com/Audacity.html?dwl=audacity-win-2.3.0.zip` and get the right download link.
- Add `fosshub` hash extraction mode.


niheaven:
This acts like clicking download link on the webpage, and during code testing, about 10 calls per sec hasn't meed the limit, so it should be sufficiant for daily use.

PR #3127, opened 2019-02-14
remove python-exp

### LINKED ISSUE #2962: What is python-exp for?
https://github.com/lukesampson/scoop/blob/15e3c25ea35fd771fac3c738ac5aacaf4a96ef38/bucket/python-exp.json

`python experimental version`? Can't find any related information. Should the manifest be removed?

h404bi:
resolve conflict, reopen #2999 , close #2962


niheaven:
For beta version of python, one can use `versions/python-beta`

PR #3131, opened 2019-02-15
haskell: add 32bit

h404bi:
close #1676

### DIFF
--- a/bucket/haskell.json
+++ b/bucket/haskell.json
@@ -5,6 +5,10 @@
         "64bit": {
             "url": "https://haskell.org/platform/download/8.6.3/HaskellPlatform-8.6.3-core-x86_64-setup.exe",
             "hash": "d94df2008a96a12957d27e4a99bbee38bd84f096a9474ede20126c73c89c1304"
+        },
+        "32bit": {
+            "url": "https://haskell.org/platform/download/8.6.3/HaskellPlatform-8.6.3-core-i386-setup.exe",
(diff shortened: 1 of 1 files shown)


Ash258:
Hash extraction is possible from <https://www.haskell.org/platform/windows.html>

Something like `64 bit Core:<br><textarea.+>($sha256)`

h404bi:
https://regex101.com/r/Nr6InZ/1/  you will get two matches.

Ash258:
😒 They render windows first on windows end, but mac os hash is first in html ... 😒
Someone would expect, that `/windows.html` will contains only windows realease 🤷‍♂️

Ash258:
Then this should do the trick:

32 bit is OK. And make 64 bit complex to specify boundaries between two windows specific text guards.

```json
"architecture": {
            "64bit": {
                "url": "https://haskell.org/platform/download/$version/HaskellPlatform-$version-core-x86_64-setup.exe",
                "hash": {
                    "url": "https://www.haskell.org/platform/windows.html",
                    "regex": "(?sm)Start Menu.*64 bit Core:<br><textarea.+>($sha256).*32 bit Core"
                }
            },
            "32bit": {
                "url": "https://haskell.org/platform/download/$version/HaskellPlatform-$version-core-i386-setup.exe",
                "hash": {
                    "url": "https://www.haskell.org/platform/windows.html",
                    "regex": "32 bit Core:<br><textarea.+>($sha256)"
                }
            }
        }
```

image

PR #3148, opened 2019-02-21
[Workaround] Download and verify before update

### LINKED ISSUE #1605: update-flow: uninstall after download
Currently, scoop update flow works by uninstall the app, downloading, and then installing again. It would be smoother to recover from network errors, if does the downloading first, and then the uninstall followed by install. 

Obviously, not a priority issue, but thought it would be a good design flow to track.

### LINKED ISSUE #2220: When updating, download & verify new package before uninstalling
Currently, performing an upgrade of a package runs an uninstallation before downloading the updated files. It is possible for the download to fail or for the checksum to not match, leaving no copy of the package installed and, potentially, no way to reinstall it. Ideally, the update package should be downloaded and verified before the uninstallation process begins.

Ash258:
This is temporary solution for #2220 until whole install.ps1 refactor is done to make this tweak easily, in the best way one-liner. ~~Today i will follow with issue how this refactor should looks like, because there is some inconsistency and could be much more easier to maintain.~~ #3149 

- Closes #2220
- Closes #2343
- Closes #1605 

## Aria2

ARia OK

Aria Fail

## Normal

OK

Fail

Co-authored-by: Chawye Hsu <h404bi@outlook.com>

### DIFF
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -153,6 +153,41 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
 
     write-host "Updating '$app' ($old_version -> $version)"
 
+    # region Workaround
+    # Workaround for https://github.com/lukesampson/scoop/issues/2220 until install is refactored
+    # Remove whole region after proper fix
+    Write-Host "Downloading new version to verify integrity of files"
+    if (aria2_enabled) {
+        dl_with_cache_aria2 $app $version $manifest $architecture $cachedir $manifest.cookies $true $check_hash
+    } else {
+        # dl_with_cache $app $version $
+        $urls = url $manifest $architecture
+
+        foreach ($url in $urls) {
+            dl_with_cache $app $version $url $null $manifest.cookies $true
+
+            $manifest_hash = hash_for_url $manifest $url $architecture
+            $source = fullpath (cache_path $app $version $url)
+            $ok, $err = check_hash $source $manifest_hash $(show_app $app $bucket)
+
+            if(!$ok) {
+                error $err
+                if(test-path $source) {
+                    # rm cached file
+                    Remove-Item -force $source
+                }
+                if($url.Contains('sourceforge.net')) {
+                    Write-Host -f yellow 'SourceForge.net is known for causing hash validation fails. Please try again before opening a ticket.'
+                }
+                abort $(new_issue_msg $app $bucket "hash check failed")
+            }
+        }
+    }
+    # There is no need to check hash again
+    $check_hash = $false
+    Write-Host "Verifing OK. Updating continue" -ForegroundColor Green
+    # endregion Workaround
+
     $dir = versiondir $app $old_version $global
 
     write-host "Uninstalling '$app' ($old_version)"
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -156,17 +156,16 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
     # region Workaround
     # Workaround for https://github.com/lukesampson/scoop/issues/2220 until install is refactored
     # Remove and replace whole region after proper fix
-    if ($check_hash) {
-        Write-Host "Downloading new version to verify integrity of files"
-        if (aria2_enabled) {
-            dl_with_cache_aria2 $app $version $manifest $architecture $cachedir $manifest.cookies $true $check_hash
-        } else {
-            # dl_with_cache $app $version $
-            $urls = url $manifest $architecture
+    Write-Host "Downloading new version"
+    if (aria2_enabled) {
+        dl_with_cache_aria2 $app $version $manifest $architecture $cachedir $manifest.cookies $true $check_hash
+    } else {
+        $urls = url $manifest $architecture
 
-            foreach ($url in $urls) {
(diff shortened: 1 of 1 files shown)


libexec/scoop-update.ps1
```diff
@@ -153,6 +153,41 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
...
+    } else {
+        # dl_with_cache $app $version $
+        $urls = url $manifest $architecture
+
+        foreach ($url in $urls) {
+            dl_with_cache $app $version $url $null $manifest.cookies $true
+
+            $manifest_hash = hash_for_url $manifest $url $architecture
+            $source = fullpath (cache_path $app $version $url)
+            $ok, $err = check_hash $source $manifest_hash $(show_app $app $bucket)
```
h404bi: The workaround always check hash even if the user specifies `--skip` or install a [`nightly` version of app](https://github.com/lukesampson/scoop/blob/0156d5d10be1827de900637de454531a143bf43a/libexec/scoop-update.ps1#L135-L138). You should respect users' behavior if they want to do that, [the option of just downloading packages](https://github.com/lukesampson/scoop/blob/9ce2fda8543913bc229203feffcbac5232ff266d/lib/install.ps1#L319) before uninstalling, but without hash checking.

h404bi:
Shouldn't be like this? Now it won't download packages before uninstalling if the `$check_hash` is `false` after you wrapping the whole fix with `if ($check_hash) {}`.
```diff
diff --git a/fix.ps1 b/fix.ps1
index cdd96cc..ed0d983 100644
--- a/fix.ps1
+++ b/fix.ps1
@@ -1,7 +1,7 @@
 # region Workaround
 # Workaround for https://github.com/lukesampson/scoop/issues/2220 until install is refactored
-# Remove whole region after proper fix
-Write-Host "Downloading new version to verify integrity of files"
+# Remove and replace whole region after proper fix
+Write-Host "Downloading new version of '$app'"
 if (aria2_enabled) {
     dl_with_cache_aria2 $app $version $manifest $architecture $cachedir $manifest.cookies $true $check_hash
 } else {
@@ -11,24 +11,25 @@ if (aria2_enabled) {
     foreach ($url in $urls) {
         dl_with_cache $app $version $url $null $manifest.cookies $true
 
-        $manifest_hash = hash_for_url $manifest $url $architecture
-        $source = fullpath (cache_path $app $version $url)
-        $ok, $err = check_hash $source $manifest_hash $(show_app $app $bucket)
+        if ($check_hash) {
+            $manifest_hash = hash_for_url $manifest $url $architecture
+            $source = fullpath (cache_path $app $version $url)
+            $ok, $err = check_hash $source $manifest_hash $(show_app $app $bucket)
 
-        if(!$ok) {
-            error $err
-            if(test-path $source) {
-                # rm cached file
-                Remove-Item -force $source
+            if (!$ok) {
+                error $err
+                if (test-path $source) {
+                    # rm cached file
+                    Remove-Item -force $source
+                }
+                if ($url.Contains('sourceforge.net')) {
+                    Write-Host -f yellow 'SourceForge.net is known for causing hash validation fails. Please try again before opening a ticket.'
+                }
+                abort $(new_issue_msg $app $bucket "hash check failed")
             }
-            if($url.Contains('sourceforge.net')) {
-                Write-Host -f yellow 'SourceForge.net is known for causing hash validation fails. Please try again before opening a ticket.'
-            }
-            abort $(new_issue_msg $app $bucket "hash check failed")
         }
     }
 }
 # There is no need to check hash again
 $check_hash = $false
-Write-Host "Verifing OK. Updating continue" -ForegroundColor Green
 # endregion Workaround
```

Ash258:
Why? When `check_hash` is false, then this workaround is not neeeded at all. It will just behave with old flow.

1. Uninstall
1. Download
1. Install

h404bi:
Because you should also care about the network interruption.

h404bi:
Awesome, that's it.
```
$ scoop update vim --skip
vim: 8.1.0982 -> 8.1.0991
Updating one outdated app:
Updating 'vim' (8.1.0982 -> 8.1.0991)
Downloading new version
gvim_8.1.0991_x64.zip (17.6 MB) [=====================>             ]  62%
 Received an unexpected EOF or 0 bytes from the transport stream.
At C:\Users\<omitted>\scoop\apps\scoop\current\lib\install.ps1:126 char:9
+         throw $e
+         ~~~~~~~~
+ CategoryInfo          : OperationStopped: (:) [], IOException
+ FullyQualifiedErrorId :  Received an unexpected EOF or 0 bytes from the transport stream.
$ 
```

r15ch13:
@h404bi is this working and merge ready? Can't test, because my Windows machine is dead.
/cc @rasa

h404bi:
@r15ch13 Looks Good To Me. 
@Ash258 I still think `Write-Host "Verifing OK. Updating continue" -ForegroundColor Green` is not necessary, but whatever.

Ash258:
Suggest alternative or should be completely removed?

It would be better to notify user on failing, that actual installation wasn't touched, but it's not possible without major core editing.

h404bi:
@Ash258 suggest that line of ouput should be removed, hash check result has been shown before.

h404bi:
@r15ch13 it's ready to merge.

PR #3160, opened 2019-02-22
Update install.ps1

Tiberriver256:
```pwsh
(Get-ExecutionPolicy) -gt 'RemoteSigned'
```

Will always evaluate to false no matter what the policy is set to. It should be set to:

```pwsh
(Get-ExecutionPolicy) -eq 'RemoteSigned'
```

### DIFF
--- a/bin/install.ps1
+++ b/bin/install.ps1
@@ -12,7 +12,7 @@ if(($PSVersionTable.PSVersion.Major) -lt 3) {
 }
 
 # show notification to change execution policy:
-if((Get-ExecutionPolicy) -eq 'RemoteSigned' -or (Get-ExecutionPolicy) -eq 'ByPass') {
+if((Get-ExecutionPolicy) -ge 'RemoteSigned' -or (Get-ExecutionPolicy) -eq 'ByPass') {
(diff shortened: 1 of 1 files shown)


Ash258:
You are wrong:

image

Tiberriver256:
My bad 😞😞😞 

Should it be set to ```-ge``` instead then in order to execute on a machine whose policy is set to 'RemoteSigned'?

image

I was unable to install it on my work PC which has a policy set to RemoteSigned from the URL but when I removed the check out of the code everything ran just fine.

r15ch13:
If this condition is `true` then the script will be stopped because of the `break`
`Unrestricted` has a value of `0` and `RemoteSigned` of `1`.
This block should only be called if the Policy is `AllSigned` (`2`), `Restricted` (`3`) or `Undefined` (`5`).


https://github.com/PowerShell/PowerShell/blob/5d54f1aa3871a826409496437e25856dc263ccc4/src/System.Management.Automation/security/SecuritySupport.cs#L31-L65

Tiberriver256:
Got it 👍

PR #3170, opened 2019-02-25
[Schema] Add metalink mode to hashmode

Ash258:
Schema is missing property for metalink

See: https://github.com/lukesampson/scoop/blob/57b0b0e1eb2c946ca74541b2464d3e61e195e979/lib/autoupdate.ps1#L213-L218

Meld example


Ash258:
Also added sourceforge and fosshub, and matched their code to follow others.

PR #3171, opened 2019-02-26
pshazz, concfg, psutils: Fix checkver error

niheaven:
- Using `/commits/master.atom` file in `checkver` to avoid github page load error (get wrong commit date 2016.03.23)
- Modify `checkver` of tools in `psutils` that return last modify date of single `ps1` file.


h404bi:
Thanks bro, it really sucks that `concfg` is continually getting wrong updates.

PR #3180, opened 2019-03-01
Add Xidel

gaeulbyul:
> Xidel is a command line tool to download and extract data from HTML/XML pages using CSS selectors, XPath/XQuery 3.0, as well as querying JSON files or APIs (e.g. REST) using JSONiq.
> https://github.com/benibela/xidel

### DIFF
--- a/bucket/xidel.json
+++ b/bucket/xidel.json
@@ -2,7 +2,7 @@
     "homepage": "http://www.videlibri.de/xidel.html",
     "description": "command line tool to download and extract data from HTML/XML pages as well as JSON APIs.",
     "version": "0.9.8",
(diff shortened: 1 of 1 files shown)


bucket/xidel.json
```diff
@@ -0,0 +1,15 @@
+{
+    "homepage": "http://www.videlibri.de/xidel.html",
+    "description": "command line tool to download and extract data from HTML/XML pages as well as JSON APIs.",
+    "version": "0.9.8",
+    "license": "GPL-3.0",
```
r15ch13:
```suggestion
   "license": "GPL-3.0-or-later",
```

PR #3195, opened 2019-03-07
download progress without newline

star2000:
There are too many download notifications. This is not necessary. We only need one line to download the progress.


Ash258:
Post before and after screenshots.

star2000:
before:

after:

PR #3197, opened 2019-03-07
Python: Use install script instead of original installer

niheaven:
Use `installer.script` instead of original installer to scoop install python, so as to not interact with system registry.

Use `dark.exe` in `wixtoolset` to extract the installer. Since there are many software that could be extracted by `dark`, a single package that contains `dark` only may be useful. (wixtoolset is 32.7MB, while dark with dlls is ~700KB (zipped).)


Ash258:
+1 For single dark executable manifest.

niheaven:
I've submitted a PR in scoopinstaller/binary-mirror (https://github.com/scoopinstaller/binary-mirror/pull/4), and a seperated dark manifest will be made when the PR is merged.

niheaven:
The CI check error is a false positive.

PR #3205, opened 2019-03-10
Git-Annex: Fix hash, add hash autoupdate

Retia-Adolf:


Ash258:
Why `$url.info` and `$sha256` variable were not used in autoupdate??

r15ch13:
Missed that

PR #3207, opened 2019-03-11
Add wp-cli version 2.1.0

edgardmessias:
https://wp-cli.org/


edgardmessias:
This project is like composer, but for wordpress developers

PR #3210, opened 2019-03-12
nvm: Test for persisting file

niheaven:
For testing with #3209 

Please totally cleanup previous installed nvm for testing, esp. persisted `setting.txt`.


niheaven:
Move testing manifests to #3209.

PR #3212, opened 2019-03-12
[WIP] Persistence rework with backward compatibility (persist)

r15ch13:
Complete rework of the `persist` property including backward compatibility.

- [ ] Introduces new `persist`-objects to JSON Schema <sup>subject to change</sup>
  ```json
    "persist": [
        {
            "type": "file",
            "name": "config.dat",
            "content": "[Config]`nlanguage=eng",
            "encoding": "UTF-8",
            "method": "copy"
        },
        {
            "type": "directory",
            "name": "data",
            "method": "link"
        },
        {
            "type": "directory",
            "name": "plugins",
            "method": "merge"
        }
    ]
  ```
- [x] Introduces new functions for Junction and Hardlink handling including Tests
  - Drops `$env:COMSPEC` and `attrib` calls
- [x] Extract old `persist_data` behaviour to `persist_data_old` <sup>subject to change</sup>

- [ ] More tests (more mocking)
- [ ] Implement actual feature :grin:

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -285,7 +285,47 @@ function isFileLocked([string]$path) {
 }
 
 function is_directory([String] $path) {
-    return (Test-Path $path) -and (Get-Item $path) -is [System.IO.DirectoryInfo]
+    return (Test-Path $path) -and (Get-Item -Path $path) -is [System.IO.DirectoryInfo]
+}
+
+function is_junction([String] $path) {
+    return (Test-Path $path) -and (Get-Item -Path $path).LinkType -eq 'Junction'
+}
+
+function create_junction([String] $link, [String] $target) {
+    if (!(Test-Path $link) -and (is_directory $target)) {
+        New-Item -ItemType Junction -Path (Split-Path -Path $link) -Name (Split-Path -Leaf $link) -Target $target | Out-Null
+        $dirInfo = New-Object System.IO.DirectoryInfo($link)
+        $dirInfo.Attributes = $dirInfo.Attributes -bor [System.IO.FileAttributes]::ReadOnly
+        return $true
+    }
+    return $false
+}
+
+function remove_junction([String] $link) {
+    if(is_junction $link) {
+        $dirInfo = New-Object System.IO.DirectoryInfo($link)
+        $dirInfo.Attributes = $dirInfo.Attributes -band (-bnot [System.IO.FileAttributes]::ReadOnly)
+        $dirInfo.Delete()
+        return $true
+    }
+    return $false
+}
+
+function create_hardlink([String] $link, [String] $target) {
+    if (!(Test-Path $link) -and (Test-Path $target) -and !(is_directory $target)) {
+        New-Item -ItemType HardLink -Path (Split-Path -Path $link) -Name (Split-Path -Leaf $link) -Target $target | Out-Null
+        return $true
+    }
+    return $false
+}
+
+function remove_hardlink([String] $link) {
+    if ((Test-Path $link) -and !(is_directory $link)) {
+        Remove-Item -Path $link | Out-Null
+        return $true
+    }
+    return $false
 }
 
 function movedir($from, $to) {
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -927,14 +927,9 @@ function link_current($versiondir) {
         abort "Error: Version 'current' is not allowed!"
     }
 
-    if(test-path $currentdir) {
-        # remove the junction
-        attrib -R /L $currentdir
-        & "$env:COMSPEC" /c rmdir $currentdir
-    }
-
-    & "$env:COMSPEC" /c mklink /j $currentdir $versiondir | out-null
-    attrib $currentdir +R /L
+    # recreate new junction
+    remove_junction $currentdir | Out-Null
+    create_junction $currentdir $versiondir | Out-Null
     return $currentdir
 }
 
@@ -950,11 +945,7 @@ function unlink_current($versiondir) {
     if(test-path $currentdir) {
         write-host "Unlinking $(friendly_path $currentdir)"
 
-        # remove read-only attribute on link
-        attrib $currentdir -R /L
-
-        # remove the junction
-        & "$env:COMSPEC" /c "rmdir $currentdir"
+        remove_junction $currentdir | Out-Null
         return $currentdir
     }
     return $versiondir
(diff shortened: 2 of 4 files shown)


schema.json
```diff
@@ -202,6 +202,80 @@
...
+                                "type": "string"
+                            },
+                            "target": {
+                                "type": "string"
+                            },
+                            "type": {
+                                "enum": [
+                                    "file",
+                                    "directory",
+                                    "folder"
```
Ash258: Why both folder and directory? Is it just alias or some actual differences?

r15ch13: Yeah, might drop it. Only `directory` should be fine.

Ash258:
Small notes without code review:

Make encoding UTF8 encoding default in code (eg. When not specified, use UTF-8), same for method link.

lib/core.ps1
```diff
@@ -285,7 +285,47 @@ function isFileLocked([string]$path) {
...
+        $dirInfo = New-Object System.IO.DirectoryInfo($link)
+        $dirInfo.Attributes = $dirInfo.Attributes -bor [System.IO.FileAttributes]::ReadOnly
+        return $true
+    }
+    return $false
+}
+
+function remove_junction([String] $link) {
+    if(is_junction $link) {
+        $dirInfo = New-Object System.IO.DirectoryInfo($link)
```
niheaven: `Remove-Item xxx -Recurse -Force` could remove read-only junction. So if we want to delete junction or hardlink, a single function that `Remove-Item xxx -Recurse -Force` is enough.

schema.json
```diff
@@ -202,6 +202,80 @@
...
+                        ],
+                        "if": {
+                            "properties": {
+                                "type": {
+                                    "const": "file"
+                                }
+                            }
+                        },
+                        "then": {
+                            "required": [
```
niheaven: These two properties should not be required, they could be ommited in most instances with no contents and encoding "UTF-8" (which is default in Windows 10 in despite of encoding with Default, UTF8, ASCII, OEM in `Out-File`).

If app require different encoding, we can set through `Get-Content | Out-File -Encoding`, but it should be rare case.

schema.json
```diff
@@ -202,6 +202,80 @@
...
+                                "encoding",
+                                "content"
+                            ]
```
niheaven: These two properties should not be required, they could be ommited in most instances with no contents and encoding "UTF-8" (which is default in Windows 10 in despite of encoding with Default, UTF8, ASCII, OEM in Out-File).

If app require different encoding, we can set through Get-Content | Out-File -Encoding, but it should be rare case.

schema.json
```diff
@@ -202,6 +202,80 @@
...
+        "persist": {
+            "anyOf": [
+                {
+                    "description": "Deprecated, use object representation instead.",
+                    "type": "string"
+                },
+                {
+                    "description": "Deprecated, use object representation instead.",
+                    "items": {
+                        "$ref": "#/definitions/stringOrArrayOfStrings"
```
niheaven: We have #/definitions/stringOrArrayOfStringsOrAnArrayOfArrayOfStrings, why not using it as former?

lib/core.ps1
```diff
@@ -285,7 +285,47 @@ function isFileLocked([string]$path) {
...
+        $dirInfo.Attributes = $dirInfo.Attributes -band (-bnot [System.IO.FileAttributes]::ReadOnly)
+        $dirInfo.Delete()
```
niheaven: Remove-Item xxx -Recurse -Force could remove read-only junction. So if we want to delete junction or hardlink, a single function that Remove-Item xxx -Recurse -Force is enough.

r15ch13: Yes, but I got an error message saying powershell is not in interactive mode, so I changed it to a simple delete be the object is already there.

_Sent from my OnePlus5T using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

lib/install.ps1
```diff
@@ -1202,13 +1213,9 @@ function unlink_persist_data($dir) {
             $filepath = $file.FullName
             # directory (junction)
             if ($file -is [System.IO.DirectoryInfo]) {
-                # remove read-only attribute on the link
-                attrib -R /L $filepath
-                # remove the junction
-                & "$env:COMSPEC" /c "rmdir /s /q $filepath"
+                remove_junction | Out-Null
```
niheaven: Small typo: missing `$filename`

lib/core.ps1
niheaven: I've tested with changing two `remove_` function's contents to `Remove-Item` (after fix type in L1216), and also with replacing all `remove_junction` and `remove_hardlink` with `Remove-Item xxx -Recurse -Force | Out-Null`, and gotten no error. Maybe need more test.

PS. I've googled that without `-Force` may return the error that `not in interactive mode`. May help.

rasa:
Can most values default, so
```json
  "persist": [
      {
          "name": "config.dat" // type=file, method=link
      },
      {
          "name": "data/" // type=directory, method=link
      }
  ]
```
works?

niheaven:
Yes, it works, use a regex to judge if it contains "/" or "\", as a shorcut of "type".

Ash258:
No need to use regex. Just `"dir/".EndsWith('/')` should be enough.

schema.json
niheaven: Remark: PS 3-5.1 use "ASCII" as default, and "UTF8" give a UTF-8 with BOM. PS Core (6) use "UTF8NoBOM" as default, and "UTF8" give a UTF-8 without BOM. So PLEASE AVOID using "UTF8" as file contents encoding.

niheaven:
Remark for `encoding`: `Out-File` in PowerShell 3-5.1 use "ASCII" as default, and "UTF8" give a UTF-8 with BOM. The same function in PowerShell Core (6) use "UTF8NoBOM" as default, and "UTF8" give a UTF-8 without BOM. So PLEASE AVOID using "UTF8" as file contents encoding. "ASCII" is preferred.

h404bi:
AFAIK previously, for restricted users, read-only junctions can interupt uninstallation, and can not be remove by `Remove-Item -Force`(#2832 #2881 #2882 #2890 ). Maybe need more tests under non-admin accounts.

niheaven:
@h404bi Agree. I've tested with an admin account (with and without elevated privilege) and a non-admin test acount, and didn't gotten error with `Remove-Item -Recurse -Force`. This may need more tests.

Ash258:
> Before Update:
>   file exists
>     copy file to persist directory

This is to prevent file overwriting instead of file editing? (mainly some qT apps lukesampson/scoop-extras#2018)

If yes i would suggest this tweak to description:
```diff
-Before update
+Before uninstallation
```

PR #3222, opened 2019-03-15
plex-server: Fix depends

niheaven:


r15ch13:
Oh, I messed up the inflict resolving :smile:

Issue #2968, opened 2019-01-09
7zip URL is not valid


habamax:
I have tried to update 7zip and face the error:

```
Installing '7zip' (18.06) [64bit]
Unable to connect to the remote server
URL https://7-zip.org/a/7z1806-x64.msi is not valid
```

PS, well it looks like 7-zip.org is down.

moosad:
It was just a temporary issue, I guess, because it's up now.

habamax:
Still unable to connect...

Tried the link in Firefox -- the same, connection timed out. 
Tried with Tor browser -- it is there, no errors.

Looks like it is not accessible from Russia (or from part of it)

habamax:
nevermind, it really is blocked in Russia

http://isitblockedinrussia.com/?host=7-zip.org

PR #3230, opened 2019-03-16
7zip: Use SourceForge for download

vxna:
As mentioned in #2968 and #2764, main [7-Zip](https://7-zip.org) website is not availble across many regions of Russia so `scoop` itself is not much usable without it in the most cases. 

This PR replaces download links with offical [SourceForge 7-Zip mirror](https://sourceforge.net/projects/sevenzip/) and fixes the issue without breaking changes for anyone, also this removes little "bus-factor" for the `7zip` package itself being hosted at project website against worldwide CDN.

Please consider not to suggest VPN and/or third-party solutions, as of myself I think this PR can provide safe and opt-in experience for `scoop` end-users from Russia.

### DIFF
--- a/bucket/7zip.json
+++ b/bucket/7zip.json
@@ -8,12 +8,12 @@
     "version": "19.00",
     "architecture": {
         "64bit": {
-            "url": "https://7-zip.org/a/7z1900-x64.msi",
-            "hash": "a7803233eedb6a4b59b3024ccf9292a6fffb94507dc998aa67c5b745d197a5dc"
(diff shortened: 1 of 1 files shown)


h404bi:
But what if some other places can not access SourceForge? Changing the manifest is not a good solution IMHO.

vxna:
@h404bi Then around [45](https://github.com/lukesampson/scoop/search?l=JSON&p=1&q=sourceforge.net) current `scoop` packages will be not available for those people.

PR #3235, opened 2019-03-18
xidel: Fix checkver

niheaven:


bucket/xidel.json
```diff
@@ -7,7 +7,8 @@
     "hash": "96854c2be1e3755f56fabb8f00d1fe567108461b9fab139039219a1b7c17e382",
     "bin": "xidel.exe",
     "checkver": {
-        "github": "https://github.com/benibela/xidel"
+        "url": "https://github.com/benibela/xidel/releases/latest",
```
Ash258: Keep github instead of url.

```suggestion
        "github": "https://github.com/benibela/xidel",
```

niheaven: Thanks. Yes, it's `releases/latest`, so `github`.

PR #3242, opened 2019-03-18
refactor(config): Move config handling to core.ps1

r15ch13:
- Migrate `~\.scoop` to `~\.config\scoop\config.json`
  - `~\.config` is used by other tools too
  - Proper highlighting in text editor because of `.json` file extension
- Remove `hashtable` and `hashtable_val` functions because `ConvertFrom-Json` is enough
  - These functions are obsolete because `ConvertFrom-Json` can handle all data formats
- Add PowerShell 6 time conversion to tests
- Removes `config.ps1` imports from `bin\*.ps1` and `libexec\*.ps1`

~~[~~`scoop alias`~~](https://github.com/lukesampson/scoop/blob/master/libexec/scoop-alias.ps1) requires changes to remove the `hashtable` specific stuff.~~


lib/core.ps1
```diff
@@ -18,6 +18,14 @@ $globaldir = $env:SCOOP_GLOBAL, "$env:ProgramData\scoop" | Select-Object -first
 #       Use at your own risk.
 $cachedir = $env:SCOOP_CACHE, "$scoopdir\cache" | Select-Object -first 1
 
+$configFile = "$env:USERPROFILE\.config\scoop\config.json"
```
niheaven:
```suggestion
$configHome = $env:XDG_CONFIG_HOME, "$env:USERPROFILE\.config" | Select-Object -First 1
$configFile = "$configHome\scoop\config.json"
```

`$HOME/.config` is [`$XDG_CONFIG_HOME`](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), so check it first.

lib/core.ps1
```diff
@@ -42,6 +50,76 @@ function Get-UserAgent() {
...
+function get_config($name, $default) {
+    if($null -eq $scoopConfig.$name -and $null -ne $default) {
+        return $default
+    }
+    return $scoopConfig.$name
+}
+
+function set_config($name, $val) {
+    if(!$scoopConfig) {
+        $scoopConfig = @{ $name = $val }
```
niheaven:
```suggestion
        ensure (Split-Path -Path $configFile) | Out-Null
        $scoopConfig = @{ $name = $val }
```

If no config, ensure `$configFile`'s path is there, otherwise config will not be created.

lib/core.ps1
```diff
@@ -42,6 +50,76 @@ function Get-UserAgent() {
...
+            $val = [System.Convert]::ToBoolean($val)
+        }
+        $scoopConfig.$name = $val
+    }
+
+    if($null -eq $val) {
+        $scoopConfig.PSObject.Properties.Remove($name)
+    }
+
+    ConvertTo-Json $scoopConfig | Set-Content $configFile -Encoding utf8
```
niheaven:
```suggestion
    ConvertTo-Json $scoopConfig | Set-Content $configFile -Encoding ASCII
```

As mentioned in https://github.com/lukesampson/scoop/pull/3212#issuecomment-472328841, we should encode file in "ASCII", since "UTF8" in Powershell 5- will be "UTF8-with-BOM", while "ASCII" be "UTF8".

niheaven:
When moving config, there are two `warn()` which is defined below and give errors (`warn` not defined). Should replace with hardcode or put `#message`(L138-L189) at top of `core.ps1`.

niheaven:
L83-L87: If config value is `false`, `$scoopConfig.$name` will be false and fall to `else` section even if there is `$name` in config...
Test case:

```powershell
scoop config debug true # okay
scoop config debug false # okay
scoop config debug false # error
```

It can be fixed by replacing L83-L87 with

```powershell
$scoopConfig | Add-Member -Force -NotePropertyMembers @{ $name = $value }
```

That is, force add note member (we want to set it, so force is okay, no matter if there is $name)

niheaven:
Another implementation: pre-defined config file that contained all config option with default value, so we could only modify existing value, no need to add new one. The pre-defined one can be in scoop's root dir and copy to config home during first run. That may help user to know what configurable options are scoop having.

test/Scoop-Config.Tests.ps1
```diff
@@ -1,20 +1,45 @@
...
+
+        $obj.one | Should -BeExactly 1
+        $obj.two[0].a | Should -Be "a"
+        $obj.two[1] | Should -Be "b"
+        $obj.two[2] | Should -BeExactly 2
+        $obj.three.four | Should -BeExactly 4
+        $obj.five | Should -BeTrue
+        $obj.six | Should -BeFalse
+        [System.DateTime]::Equals($obj.seven, $(New-Object System.DateTime (2018, 06, 25, 09, 03, 15, 805))) | Should -BeTrue
+        [System.DateTime]::Equals([System.DateTime]::parse($obj.eight), $(New-Object System.DateTime (2019, 03, 18, 15, 22, 09, 393))) | Should -BeTrue
```
h404bi: the test failed.

niheaven: This is due to TimeZone.

```powershell
$(New-Object System.DateTimeOffset (2019, 03, 18, 15, 22, 09, 393, $(New-Object TimeSpan (1, 0, 0)))).LocalDateTime
```

libexec/scoop-uninstall.ps1
```diff
@@ -5,15 +5,16 @@
...
+
+. "$PSScriptRoot\..\lib\core.ps1"
+. "$PSScriptRoot\..\lib\manifest.ps1"
+. "$PSScriptRoot\..\lib\help.ps1"
+. "$PSScriptRoot\..\lib\install.ps1"
+. "$PSScriptRoot\..\lib\shortcuts.ps1"
+. "$PSScriptRoot\..\lib\psmodules.ps1"
+. "$PSScriptRoot\..\lib\versions.ps1"
+. "$PSScriptRoot\..\lib\getopt.ps1"
+. "$PSScriptRoot\..\lib\config.ps1"
```
niheaven:
```suggestion

```

Typo?

test/Scoop-Config.Tests.ps1
r15ch13: Now it will only test if it's a DateTime object. Not gonna bother about Timezones ...

libexec/scoop-uninstall.ps1
r15ch13: Will be removed later, because of merge conflict

lib/core.ps1
```diff
@@ -42,6 +51,81 @@ function Get-UserAgent() {
...
+
+function set_config($name, $value) {
+    if(!$scoopConfig) {
+        ensure (Split-Path -Path $configFile) | Out-Null
+        $scoopConfig = @{ $name = $value }
+    } else {
+        if($value -eq [bool]::TrueString -or $value -eq [bool]::FalseString) {
+            $value = [System.Convert]::ToBoolean($value)
+        }
+        if($scoopConfig.PSObject.Properties.Item($name)) {
```
niheaven:
```suggestion
        if($null -ne $scoopConfig.$name) {
```

If `$scoopConfig` don't have `$name`, it is `$null`. It's needless to check the properties.

niheaven: No `$null` value will be in config, since it is removed in L90-L92 below, so check isnull is enough.

niheaven:
There is only one thing remaining... Since `set_config` has output now, we should `Out-Null` it when we don't need, otherwise, e.g., `scoop update` will produce:

```
λ  scoop update
Updating Scoop...
Updating 'Ash258' bucket...
Updating 'dorado' bucket...
Updating 'extras' bucket...
Updating 'java' bucket...
Updating 'nerd-fonts' bucket...
Updating 'nih' bucket...
Updating 'rasa' bucket...
Updating 'versions' bucket...


SCOOP_BRANCH           : master
rootPath               : E:\Scoop
aria2-enabled          : False
msiextract_use_lessmsi : True
globalPath             : E:\ScoopG
debug                  : False
lastupdate             : 2019-03-27T11:20:17.0666758+08:00
cachePath              : E:\ScoopC
SCOOP_REPO             : https://github.com/lukesampson/scoop

```

Black lines above is added by scoop itself, that's messy, I think.

r15ch13:
Rebased onto `develop` and added missing `| Out-Null` redirections. Please test again :smile:

PR #3249, opened 2019-03-22
Add checkver and autoupdate from Chocolatey

niheaven:
TL;DR

```json
"checkver": {
    "choco": "ScriptCS"
}
```

- Add `chocolatey` (alias `choco`) to `checkver`.
- Using `([\d.]+)\s` in `https://chocolatey.org/packages/$choco` to check new version
- Using `https://packages.chocolatey.org/$choco.$version.nupkg` as `autoupdate.url`
- Using `$sha256.*?$basename` in `https://chocolatey.org/packages/$choco` to get hash

Test Case: `scriptcs.json` attached

Similar apps: Apps that distribute Chocolatey releases such as `extras/nuget-package-explorer`


Ash258:
Why are you removing autoupdate form manifest and then adding it back with some default values?? None of autoupdate type do this. You should follow other ones. In this state you can't create own autoupdate. Multiple URLs (for example when it's implemented), extract_dir, ...

bucket/scriptcs.json
```diff
@@ -7,10 +7,6 @@
     "extract_dir": "tools",
     "bin": "scriptcs.exe",
     "checkver": {
-        "url": "https://chocolatey.org/packages/ScriptCs",
-        "regex": "ScriptCs ([\\d.]+)"
-    },
-    "autoupdate": {
-        "url": "https://packages.chocolatey.org/ScriptCs.$version.nupkg"
+        "choco": "ScriptCs"
```
h404bi: I think this line should be in `checkver`.

niheaven:
> 
> 
> Why are you removing autoupdate form manifest and then adding it back with some default values?? None of autoupdate type do this. You should follow other ones. In this state you can't create own autoupdate. Multiple URLs (for example when it's implemented), extract_dir, ...

Your're right, we could use `chocolatey` mode in hash extraction instead. See the new commit for modifications, it's just like "fosshub" and "sf".

lib/autoupdate.ps1
```diff
@@ -196,6 +196,10 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
         $hashmode = 'sourceforge'
     }
 
+    if (!$hashfile_url -and $url -match "^(?:.*chocolatey.org\/)(?<choconame>.*)$version.*$") {
+        $hashmode = 'chocolatey'
```
Ash258: Update schema aswell

bucket/scriptcs.json
niheaven: Since sometimes Chocolatey page name is not the same as `$app`...

I'll add the option to use just `"checkver": "choco"`, just like `"checkver": "github"`.

niheaven:
`$choconame` used in `https://chocolatey.org/packages/$choconame` now use `($App -replace '[-]', '')` if `"checkver": "choco"`. This is for `extras/nuget-package-explorer` which name in choco is `nugetpackageexplorer`. If there is other general replacement rule, it can be add to the `-replace` regex.

bin/checkver.ps1
```diff
@@ -105,6 +107,7 @@ $Queue | ForEach-Object {
     Register-ObjectEvent $wc downloadstringcompleted -ErrorAction Stop | Out-Null
 
     $githubRegex = '\/releases\/tag\/(?:v|V)?([\d.]+)'
+    $chocoRegex = 'Chocolatey.*?([\d.]+)[\r\n]'
```
r15ch13: The regex is too ambiguous (See: https://regex101.com/r/QUv1Em/1)


Better would be `Nuspec reference:\s*&lt;version&gt;(.*)&lt;\/version&gt;`
(See: https://regex101.com/r/QUv1Em/2)

niheaven: Use `>([\d.]+)<\/h2>` as the regex.

r15ch13: Or `<h2[^>]+>([\d.]+)<\/h2>` 😄

niheaven: `>([\d.]+)<\/h2>` is enough, I thought, with only 1 found, that is the version number:)

niheaven: Okay, changed it.

brandon93s:
Do we really want scoop `autoupdate` and `checkver` dependent on another package manager? This feels like a strange dependency and allows for issues upstream to trickle into scoop. 

Could / should we limit this to trusted packages only to reduce the security implications?

niheaven:
Checkver and hash from Chocolatey is for those packages that are released on Chocolatey, e.g. [ScriptCS](http://scriptcs.net/), [NuGet Package Explorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer), [NUnit Console](https://github.com/nunit/nunit-console/), etc.

Those packages are *released* on Chocolatey that a full nupkg package is on Choco's server, and Choco provides Virus Scan Results (which includes checksum). For these ones, we could use them just likes Choco is a software release source, not another packages manager.

For most packages that are *managed* by Choco, there are only a download link and install ps1 in the nupkg, not the package, so these ones wouldn't meet the requirement of this PR, and the package's checksum couldn't be gotten by this.

brandon93s:
Ah okay, thanks for the clarification!

r15ch13:
Some changes:
- Don't remove the `-` from the app name (just use the correct choco name for the manifest, or not `choco` property)
- Match only the `.nupkg` file instead of the `h2`-Tag
- Improve `hashfile_url` regex
- Use more specific regex for extracting the hash

`checkver.ps1`
```pwsh
    # remove $chocoRegex

    if ($json.checkver -eq 'choco' -or $json.checkver -eq 'chocolatey') {
        $url = "https://chocolatey.org/packages/$App"
        $regex = "(?i:$App).([\d.]+).nupkg"
    }

    if ($json.checkver.choco) {
        $url = "https://chocolatey.org/packages/$($json.checkver.choco)"
        $regex = "(?i:$($json.checkver.choco)).([\d.]+).nupkg"
    }

    if ($json.checkver.chocolatey) {
        $url = "https://chocolatey.org/packages/$($json.checkver.chocolatey)"
        $regex = "(?i:$($json.checkver.chocolatey)).([\d.]+).nupkg"
    }
```

`autoupdate.ps1`
```pwsh
    if (!$hashfile_url -and $url -match "(?:packages.)?chocolatey.org\/(?:packages\/)?(?<choconame>[\w-]+)(\.$version\.nupkg$|\/$version$)") {
        $hashmode = 'chocolatey'
    }
# snip

        'sourceforge' {
            # change the URL because downloads.sourceforge.net doesn't have checksums
            $hashfile_url = (strip_filename (strip_fragment "https://sourceforge.net/projects/$($Matches.project)/files/$($Matches.file)")).TrimEnd('/')
            $hash = find_hash_in_textfile $hashfile_url $substitutions '"$basename":.*?"sha1":\s"([a-fA-F0-9]{40})"'
        }
        'chocolatey' {
            $hashfile_url = "https://chocolatey.org/packages/$($Matches.choconame)"
            $hash = find_hash_in_textfile $hashfile_url $substitutions 'sha256:\s+([a-fA-F0-9]{64})"[^>]+>$basename'
        }
```

niheaven:
Done.

For autoupdate.url pattern, `https://chocolatey.org/api/v2/package/ScriptCs/0.17.1` should not be allowed since this will not give a correct file name so that auto hash extraction will not work.

PR #3257, opened 2019-03-23
global: Add GNU GLOBAL v6.6.3

niheaven:


niheaven:
Add `description` and change `homepage` to GNU GLOBAL's homepage instead of win32-port's homepage.

PR #3274, opened 2019-03-28
install.ps1: Run post_install immediately after installer

niheaven:
Now `post_installer` is run after creating shims and shortcuts, that is incomprehensible. The intuitive logic should be `pre_install > run_installer > post_install` and then do some end work likes creating shims, persisting data, etc..

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -44,6 +44,7 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
     unpack_inno $fname $manifest $dir
     pre_install $manifest $architecture
     run_installer $fname $manifest $architecture $dir $global
+    post_install $manifest $architecture
     ensure_install_dir_not_in_path $dir $global
     $dir = link_current $dir
     create_shims $manifest $dir $global $architecture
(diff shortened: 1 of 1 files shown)


r15ch13:
This might break `post_install` scripts relying on the data in `$persist_dir`.

niheaven:
Move persist_data before post_install

The motive is to move shims and shortcuts creating after `post_install`, so that the script could do some tweaks that may needed in shims or shorcuts.

Ash258:
This will break mpv and mpv-git (for example) manifests

linsui:
I wonder what the difference between `installer.script` and `pre_install` is. Why not run `installer.script` after `install.file` ? A use case is calibre. Now we can't run `"Move-Item \"$dir\\Calibre Portable\\*\" \"$dir\"` after install before `bin`, `shortcuts` and `persist` .
By the way, now there is no `post_install` in `mpy` .

rashil2000:
@niheaven is this still necessary?

niheaven:
Yes, but this may be contained in install's refactoring.

PR #3278, opened 2019-03-28
 micronaut: update to version 1.0.5

cyparu:


Ash258:
What's wrong with excavator? Why this was not picked up?

r15ch13:
apt-upgrade seems to have killed the docker containers mid-update... restarting now. Should be running again.

PR #3281, opened 2019-03-29
fix(scoop-install.ps1): Auto uninstall previous failed installation

### LINKED ISSUE #4058: Cancelled installation treated as installed by scoop

niheaven:
- Close #4058
Automatically uninstall previous failed installation when detected, rather then give an error and let user do uninstallation manually.


niheaven:
Ping @r15ch13

PR #3283, opened 2019-03-29
nodejs: Update to version 11.13.0

hpsbranco:


Ash258:
No need to post these PRs. At least manually. Just run auto-pr with -request parameter or keep it on @r15ch13 when he fix Excavator

hpsbranco:
I was actually testing auto-pr when I've made these.
I just picked some random apps I use.

PR #3299, opened 2019-04-02
`autoupdate.ps1`: Fix `$checksum` to allow BASE64

niheaven:
Enable `$checksum` passthrough BASE64 encoded checksum, e.g. `sha512` of `latest.yml` in many electron released apps (joplin, netron, etc.).

md5:     32 -> 22 + 2
sha1:    40 -> 27 + 1
sha256:  64 -> 43 + 1
sha512: 128 -> 86 + 2


r15ch13:
Hm, would rather add a new one and leave `$checksum` as is.
`'$base64' = '([a-zA-Z0-9+\/=]{24,128})';`

niheaven:
Add seperate `$base64`.

In my opinion, `$checksum` stands for an universal checksum, so it should extract BASE64 checksum. But using a seperate `$base64` is also acceptable, just as `$sha256`, `$sha516`.

PR #3300, opened 2019-04-02
Added profile folder to persist for clink

edgardmessias:


c30ra:
Why the persist is no longer in the current manifest?

PR #3310, opened 2019-04-08
tesseract: Update to 4.1.0.20190314

niheaven:
tesseract: Update to 4.1.0.20190314

tesseract-languages: Add equ.traineddata (Math / equation detection)

### DIFF
--- a/bucket/tesseract.json
+++ b/bucket/tesseract.json
@@ -1,15 +1,16 @@
 {
     "homepage": "https://github.com/UB-Mannheim/tesseract/wiki",
     "license": "Apache-2.0",
-    "version": "4.0.0.20181030",
+    "version": "4.1.0.20190314",
+    "description": "Open Source OCR Engine",
     "architecture": {
         "64bit": {
-            "url": "https://digi.bib.uni-mannheim.de/tesseract/tesseract-ocr-w64-setup-v4.0.0.20181030.exe#/dl.7z",
-            "hash": "f590c1fb1a5e02781635758eada2d87084d64d1e6fb0f761ef8237e60d7ea8ea"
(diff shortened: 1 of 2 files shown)


bucket/tesseract.json
```diff
@@ -42,8 +43,8 @@
         "tesseract-languages": "tesseract-languages"
     },
     "checkver": {
-        "re": "tesseract-ocr-w32-setup-v(?<version>[\\d.]+).exe",
-        "url": "https://digi.bib.uni-mannheim.de/tesseract/"
+        "re": "tesseract-ocr-w32-setup-v(4.[\\d\\w.-]+)\\.exe",
```
Ash258: Is this hardcoded `4` needed? When you use C=M.. workaround it will always be on top

niheaven: Version 4 is the current stable version, and version 3 is previous release, it's not updated since version 4 stable released. In the near future, no hardcoded version number is needed.

bucket/tesseract.json
```diff
@@ -43,8 +43,8 @@
         "tesseract-languages": "tesseract-languages"
     },
     "checkver": {
-        "re": "tesseract-ocr-w32-setup-v(4.[\\d\\w.-]+)\\.exe",
-        "url": "https://digi.bib.uni-mannheim.de/tesseract/?C=M;O=D"
+        "url": "https://digi.bib.uni-mannheim.de/tesseract/?C=M;O=D",
+        "regex": "tesseract-ocr-w32-setup-v([\\d\\w.-]+)\\.exe"
```
Ash258: `\\d` is already contained inside `\\w`

>  \w - "word character": ASCII letter, digit or underscore

https://regex101.com/r/QYnbxR/1/

```suggestion
        "regex": "tesseract-ocr-w32-setup-v([\\w\\.-]+)\\.exe"
```

PR #3204, opened 2019-03-10
decompress.ps1: Rewrite extract_xxx functions

niheaven:
PART II of https://github.com/lukesampson/scoop/pull/3169

- Combine `lessmsi_config`, `extract_lessmsi` into `extract_msi` (unique function, switch by `MSIEXTRACT_USE_LESSMSI`)
- Rewrite `unpack_inno` to `extract_inno` and handle InnoSetup in `dl_urls` with `extract_inno`
- Use `Expand-Zipfile` function in [scoopinstaller/install](https://github.com/scoopinstaller/install/blob/master/install.ps1) as new `extract_zip`
- Use `&(file_path 7zip 7z.exe)` instead of `7z` shim to avoid error caused by `scoop install 7zip 7zip-zstd; scoop uninstall 7zip-zstd`
- `extract_xxx` now have similar params (`$path, $to, $recurse`)
- Remove unused condition in `depends.ps1`

Test Cases are as follow:
```
Describing Decompression function

  Context 7zip extractionLoading 7z1900-x64.msi from cache
Checking hash of 7z1900-x64.msi ... ok.
Extracting 7z1900-x64.msi ... done.
Linking D:\Scoop\apps\7zip\current => D:\Scoop\apps\7zip\19.00
'7zip' (19.00) was installed successfully!
    [+] extract normal compressed file 179ms
    [+] extract nested compressed file 495ms
    [+] extract nested compressed file with different inner name 235ms
    [+] works with '$recurse' param 89ms

  Context msi extraction
Loading lessmsi-v1.6.3.zip from cache
Checking hash of lessmsi-v1.6.3.zip ... ok.
Extracting lessmsi-v1.6.3.zip ... done.
Linking D:\Scoop\apps\lessmsi\current => D:\Scoop\apps\lessmsi\1.6.3
'lessmsi' (1.6.3) was installed successfully!
    [+] extract normal MSI file 1.71s
    [+] extract empty MSI file using lessmsi 317ms
    [+] works with '$recurse' param 1.11s

  Context inno extraction
Loading innounp048.rar from cache
Checking hash of innounp048.rar ... ok.
Extracting innounp048.rar ... done.
Linking D:\Scoop\apps\innounp\current => D:\Scoop\apps\innounp\0.48
'innounp' (0.48) was installed successfully!
    [+] extract Inno Setup file 246ms
    [+] works with '$recurse' param 169ms

  Context zip extraction
    [+] extract compressed file 117ms
    [+] works with '$recurse' param 112ms
```

[...]

PR #3314, opened 2019-04-09
refactor(cmd_available/Get-Command): Add Test-CommandAvailable function

niheaven:
A 5 years old function `cmd_available` is aiming to check existence of `$cmd` (used to check 7z, innounp), and can replace `Get-Command` with a `[Boolean]` output.


lib/core.ps1
```diff
@@ -152,6 +152,11 @@ function file_path($app, $file) {
     return $null
 }
 
+function cmd_available($cmd) {
+    try { Get-Command $cmd -ea stop | out-null } catch { return $false }
+    $true
```
Ash258: Why not just one liner with bool cast??

```suggestion
    return [bool] (Get-Command $cmd -ErrorAction SilentlyContinue)
```

Or at least use return
```suggestion
    return $true
```
image

niheaven: Thanks, just move this function from `install.ps1`.

niheaven:
Change function body to `return [Boolean](Get-Command $cmd -ErrorAction Ignore)`

r15ch13:
Since it gets refactored anyway, let's rename it to `Test-CommandAvailable` to make a small move towards common PowerShell naming conventions :smile:

niheaven:
Rename to `Test-CommandAvailable`

I'll rename functions in #3204 later.

test/Scoop-Core.Tests.ps1
```diff
@@ -135,12 +135,6 @@ Describe "get_app_name_from_ps1_shim" -Tag 'Scoop' {
         get_app_name_from_ps1_shim "$shim_path" | should -be "mockapp"
     }
 
-    It "returns app name if file exists and is a shim to an app cerca August 2018" -skip:$isUnix {
```
r15ch13: Why was this test removed? 🤔

niheaven: `shim` is tested in the above test, and this should be removed (maybe after 2018.08?)

niheaven: If it is not proper to remove it here, I could make a new PR to remove it. I remove it in passing while modifying the following two tests.

r15ch13:
Please resolve the merge conflict.

PR #3330, opened 2019-04-17
PowerShell: Update requirement to version 5

### LINKED ISSUE #3325: Would scoop update its minimum recommended PowerShell version to 5?
https://github.com/lukesampson/scoop/issues/2402, https://github.com/lukesampson/scoop-extras/pull/1993#issuecomment-482459556

Would scoop update to PS 5? Appveyor CI check only check 5 and 6, and 5 has several useful new cmdlet, e.g. `Expand-Archive`.

Windows 7 has PS 2, and 8 has 3, 8.1 has 4, 10 has 5, and I think the major Windows version that user has should be 10 and 7. For Windows 10, no update is needed, for Windows 7, there is no difference to update to 3 or 5.

niheaven:
Close #3325

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,12 +37,13 @@ Scoop is a command-line installer for Windows.
 ## What does Scoop do?
 
 Scoop installs programs from the command line with a minimal amount of friction. It tries to eliminate things like:
-* Permission popup windows
-* GUI wizard-style installers
-* Path pollution from installing lots of programs
-* Unexpected side-effects from installing and uninstalling programs
-* The need to find and install dependencies
-* The need to perform extra setup steps to get a working program
+
+- Permission popup windows
+- GUI wizard-style installers
+- Path pollution from installing lots of programs
+- Unexpected side-effects from installing and uninstalling programs
+- The need to find and install dependencies
+- The need to perform extra setup steps to get a working program
 
 Scoop is very scriptable, so you can run repeatable setups to get your environment just the way you like, e.g.:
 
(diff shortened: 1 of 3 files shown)


rikuben:
How about suggesting to install pwsh too as an alternative?

README.md
```diff
@@ -57,13 +58,14 @@ If you've built software that you'd like others to use, Scoop is an alternative
 
 ## Requirements
 
-* Windows 7 SP1+ / Windows Server 2008+
-* [PowerShell 3](https://www.microsoft.com/en-us/download/details.aspx?id=34595) (or later) and [.NET Framework 4.5+](https://www.microsoft.com/net/download)
-* PowerShell must be enabled for your user account e.g. `set-executionpolicy remotesigned -s currentuser`
+- Windows 7 SP1+ / Windows Server 2008+
+- [PowerShell 5](http://aka.ms/wmf5download) (or later) and [.NET Framework 4.5](https://www.microsoft.com/net/download) (or later)
```
r15ch13:
```suggestion
- [PowerShell 5](https://aka.ms/wmf5download) (or later) and [.NET Framework 4.5](https://www.microsoft.com/net/download) (or later)
```

niheaven:
Good idea! I'll add it later.

Ash258:
Suggesting pwsh is nice, but how would updates be handled?

Problem which i am thinkig of today: When you install pwsh using scoop and you need to update pwsh. How do you do this? Because uninstallation will try to remove pwsh (uninstall old) and fail due to it is use.

niheaven:
I don't suggest install pwsh using scoop, since it will not be registered in the system, and won't be found by many apps.

And in readme of scoop, pwsh is a prerequisite of installation, new user shouldn't install it by scoop, so it's ok, I think.

Issue #2715, opened 2018-10-30
support more file extensions for manifest


ekoeryanto:
It is nice if we support more extension for manifest, like

1. app.yml
2. app.toml, etc.

Ash258:
I am going to look for supporting yaml. Best option for this types files. Mainly for supporting [block chomping](https://yaml-multiline.info/) (#2609)

It should not be so hard. It would only required extrernal 
1. [module](https://github.com/cloudbase/powershell-yaml)
1. Add wraper for `parse_json` like `parse` with arguments `-Yaml` `-Json` with default fallback
    1. Since ConvertFrom-Yaml work exactly like ConvertFrom-Json only 1 Global variable, which indicates extension should be needed
1. Of couse fix tests, all depends lines.

shoogle:
It might be a good idea to replace JSON with a more readable format, but supporting more than one extension is not a good idea. It just means more documentation and unnecessary dependencies. Obviously you'd need a transition period where both JSON and the new format are supported but, keeping them both indefinitely is not a good idea.

couleurm:
> It might be a good idea to replace JSON with a more readable format, but supporting more than one extension is not a good idea. It just means more documentation and unnecessary dependencies. Obviously you'd need a transition period where both JSON and the new format are supported but, keeping them both indefinitely is not a good idea.

yaml is built on top of json, it wouldn't be an issue to mix them up

niheaven:
@rashil2000 Why close such feature request issue?

rashil2000:
Detailed discussion regarding manifests formats happened here https://github.com/ScoopInstaller/Scoop/issues/2609

niheaven:
That one is closed, too...

Please keep feature request issues open as reminders, and wait for someone (maintainers or not) to realize them.

goyalyashpal:
toml sucks, but yeah, i myself came here this time to say that yaml (more precisely, strict-yaml) is much much better format for such use-cases as manifests.

especially it's multiline support far far outshines the single line string arrays in json for `pre_install`, `pre_uninstall` etc.

goyalyashpal:
> strict yaml parser in pure powershell

powershell doesn't have FFI to libs having C-ABI?

couleurm:
does this use YamlDotNet?

goyalyashpal:
support comments (jsonc) in app manifests #5913

as comments help a lot:
* in commenting out the details for the architectures u don't use in local manifests
[...]

PR #3332, opened 2019-04-18
[manifest] Present new Out-Manifest and Get-Manifest function

Ash258:
First part of YAML manifests support.

These wrapper functions will help with integrating new manifest types and make code a bit cleaner.

#2715


niheaven:
Why not use `parse_json` as warpper function of `Get-Manifest` so `parse_json` could be deleted after no reference exist?

Ash258:
Because i want to extend that funtion get-manifest to something like this:

```powershell
function Get-M {
switch ($extensionOfFile) {
    'yaml|'yml' {
        $man parse_yml
     }
	default {
		$man = parse_json
	}
}
return $man
}
```

Ash258:
https://github.com/Ash258/scoop/blob/35af9a326aa6c542f75407f2ea7ca02e11b37634/lib/manifest.ps1#L59-L70
This is old implementation which i will port to use switch maybe

niheaven:
Maybe as follow:

```powershell
function parse_json {
    Get-Manifest ...
}
function Get-Manifest (..., [Switch]$XML) {
    if ($XML) {
        ...
    } else {
        ...
    }
}
```

Ash258:
I do not want to add switch parameter. I will add yaml support and someone could add toml support, so new parameter need to be set, which is not good.

I rather put filepath and then determine file type and execute correct parsing function.

Ash258:
Maybe move `parse_json` into `lib\json.ps1`? 🤔

niheaven:
And a new `xml.ps1`

Ash258:
Exactly.
Try to move towards strict and more divergent /independent modules.

PR #3336, opened 2019-04-19
scoop-status: Add bucket and global info to status output

jfut:
This will add bucket and global info to status output, like `scoop list`.

- Before:

```
PS C:\Users\USERNAME> scoop status
Scoop is up to date.
Updates are available for:
    explzh.ja: 7.67 -> 7.78
    go: 1.12.3 -> 1.12.4
    pleiades4.8-java-win-full: 4.8.0.v20180627 -> 4.8.0.v20180923
    jetbrains-toolbox: 1.14.5037 -> 1.14.5179
    keepassxc: 2.4.0 -> 2.4.1
```

- After:

```
PS C:\Users\USERNAME> scoop status
Scoop is up to date.
Updates are available for:
    explzh.ja: 7.67 -> 7.78 [jfut] *global*
    go: 1.12.3 -> 1.12.4 *global*
    pleiades4.8-java-win-full: 4.8.0.v20180627 -> 4.8.0.v20180923 [pleiades] *global*
    jetbrains-toolbox: 1.14.5037 -> 1.14.5179
    keepassxc: 2.4.0 -> 2.4.1 [extras]
```

### DIFF
--- a/libexec/scoop-status.ps1
+++ b/libexec/scoop-status.ps1
@@ -44,6 +44,7 @@ $true, $false | ForEach-Object { # local and global apps
     Get-ChildItem $dir | Where-Object name -ne 'scoop' | ForEach-Object {
         $app = $_.name
         $status = app_status $app $global
+        $install_info = install_info $app $status.version $global
         if($status.failed) {
             $failed += @{ $app = $status.version }
         }
@@ -51,7 +52,7 @@ $true, $false | ForEach-Object { # local and global apps
             $removed += @{ $app = $status.version }
         }
         if($status.outdated) {
-            $outdated += @{ $app = @($status.version, $status.latest_version) }
+            $outdated += @{ $app = @($status.version, $status.latest_version, $install_info.bucket, $global) }
         }
         if($status.missing_deps) {
             $missing_deps += ,(@($app) + @($status.missing_deps))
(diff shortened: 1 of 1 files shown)


Ash258:
I do not see a reason to see this in status when you see this in `list`.

rasa:
I don’t think the bucket is that helpful, but showing it’s a global install alerts the user to adding `-g` to the update command. In fact, it would be nice if `scoop update *` had a “update all user and global apps” in a single run. Say `scoop update -a *`? Or better yet, we could add a `-u` option to indicate user updates, and then `scoop update *` updates both user and global apps. 

Also, if a user specifies an app, scoop should update it without the user specifying `-g` even if the app is installed globally.

jfut:
Thanks for the reply and hint.

My understanding was incorrect. I misunderstood that it need to update them separately such as `sudo scoop update go -g` and `scoop update keepassxc`, so I have always checked `scoop list` after `scoop status` to know if the app is global.

But, it was safe to just run `sudo scoop update go keepassxc -g`. `go` has been updated with global and` keepassxc` has been updated with user-local.

Also, although `scoop update *` is very nice, I wanted to update them separately, as updating of running applications may fail, or it may include applications that take a long time to update.

I close this PR. Thank you!

PR #3338, opened 2019-04-19
Fix install.ps1

twilight-sparkle-irl:
## Issue 1: Bypass is accidentally disallowed

This is why I even poked at this script in the first place. The commit where this check was added ([a3cae34](a3cae34)) failed to realize that checking `-or (Get-ExecutionPolicy) -eq "Bypass"` meant that the script would *fail* on Bypass, since if the condition is true, it stops the install instead of continuing it. I've fixed this by replacing it with `-and (Get-ExecutionPolicy) -ne "Bypass"`, meaning if it's greater than RemoteSigned *and* is *not* Bypass, not if it's greater than RemoteSigned or is Bypass.

## Issue 2: Doesn't allow user to force install 

Even though the Bypass thing is silly and should simply be patched out, I think it's worth letting users decide on their own to let the install continue if they really, really want.    
Therefore, `$ScoopInstallDontKill` and `$ScoopInstallConfirmKill` now allow the user to respectively non-interactively and interactively ignore requirements. These probably don't have to be documented anywhere, anyone in a scenario where they make sense would be able to find them on their own or should be asking for help anyway.

## Issue 3: Install failures don't restore environment

Noticed this while working on Issue 2 -- the original script simply returned (or exited, but we'll get to that later)...but success correctly restored a variable. Meaning that the function for ending the install on failure was even more useful. Nice.

## Nitpick: Consistent decisions as to style and commands

It was quite obvious that the one part by one person, and the other wasn't. `$core_file` and `$zipfile`? Write-Host and Write-Output?   
I went ahead and decided Write-Host was correct (scripts that fail should have the human looking over why anyway) and variables should have \_s in them.    
Another decision I made is that there's no reason to check if you're in a script before exiting, as you don't have to -- we already have Error Action set to stop the script, so all we need to do is throw an error and it'll stop. Using "break" is simply hacky.

## Nitpick: remote install

`iex (iwr` works on Powershell 3. Scoop requires Powershell 3. QED. 

---

I realize that this issue comes off as a bit snarky, but it's not meant to be and I apologize if it's rude!! Feel free to point out any issues or such with what I've added.

### DIFF
--- a/bin/install.ps1
+++ b/bin/install.ps1
@@ -1,29 +1,42 @@
 # requires -v 3
 
 # remote install:
-#   iex (new-object net.webclient).downloadstring('https://get.scoop.sh')
+#   iex (iwr https://get.scoop.sh)
 $old_erroractionpreference = $erroractionpreference
 $erroractionpreference = 'stop' # quit if anything goes wrong
 
+function KillScoopInstall() { 
+    if ($ScoopInstallDontKill) { return }
+    if ($ScoopInstallConfirmKill) {
+        if ((Read-Host -prompt "Continue? [y/N]") -match "^[yY]") { # Continues on anything starting with "y" or "Y"
+            Write-Output "Continuing..."
+            return
+        }
+    }
+    $erroractionpreference = $old_erroractionpreference
+    Remove-Item -path Function:KillScoopInstall # Doesn't make sense to leave it hanging around...
+    throw "A requirement failed and the install was stopped. Please review the output."
+}
+
 if(($PSVersionTable.PSVersion.Major) -lt 3) {
-    Write-Output "PowerShell 3 or greater is required to run Scoop."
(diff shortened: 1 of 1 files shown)


Ash258:
I do not see a reason to maintain installation inside this repo, when there is already https://github.com/scoopinstaller/install/, which is not dependent on scoop core and ready for bucket folder extraction.

no-boot-device:
I didn't know about that as https://get.scoop.sh/ does not redirect there, and https://scoop.sh/ does not mention it. Perhaps they should point there instead...?

h404bi:
The new standalone installer is still WIP, the website will redirect there when the installer is ready.

rashil2000:
> I didn't know about that as https://get.scoop.sh/ does not redirect there, and https://scoop.sh/ does not mention it. Perhaps they should point there instead...?

It does now

PR #3341, opened 2019-04-19
[buckets] Optimize buckets function

Ash258:
- Use Verb-Noun naming

Keep `buckets` function for compatibility if someone use it inside manifests.

### DIFF
--- a/lib/buckets.ps1
+++ b/lib/buckets.ps1
@@ -40,12 +40,12 @@ function apps_in_bucket($dir) {
 }
 
 function Get-LocalBucket {
-	<#
+    <#
     .SYNOPSIS
         List all local buckets.
     #>
 
-	return (Get-ChildItem $bucketsdir).Name
+    return (Get-ChildItem $bucketsdir).Name
 }
 
 function buckets {
--- a/lib/buckets.ps1
+++ b/lib/buckets.ps1
@@ -1,3 +1,5 @@
+. "$PSScriptRoot\core.ps1"
+
 $bucketsdir = "$scoopdir\buckets"
 
 <#
@@ -49,6 +51,8 @@ function Get-LocalBucket {
 }
 
 function buckets {
+    Show-DeprecatedWarning $MyInvocation 'Get-LocalBucet'
+
     return Get-LocalBucket
 }
 
--- a/lib/buckets.ps1
+++ b/lib/buckets.ps1
@@ -40,11 +40,16 @@ function apps_in_bucket($dir) {
 }
 
 function buckets {
-    $buckets = @()
-    if(test-path $bucketsdir) {
-        Get-ChildItem $bucketsdir | ForEach-Object { $buckets += $_.Name }
-    }
-    return $buckets
+    return (Get-ChildItem $bucketsdir).Name
+}
+
+function Get-LocalBucket {
+	<#
+    .SYNOPSIS
+        List all local buckets.
+    #>
+
+	return buckets
 }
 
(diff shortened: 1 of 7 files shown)


lib/buckets.ps1
```diff
@@ -39,12 +39,17 @@ function apps_in_bucket($dir) {
     return Get-ChildItem $dir | Where-Object { $_.Name.endswith('.json') } | ForEach-Object { $_.Name -replace '.json$', '' }
 }
 
+function Get-LocalBucket {
```
r15ch13: The plural of `Get-LocalBuckets` would be better since it was `buckets`

Ash258: https://github.com/PowerShell/PSScriptAnalyzer/blob/development/RuleDocumentation/UseSingularNouns.md

lib/buckets.ps1
```diff
@@ -39,12 +39,17 @@ function apps_in_bucket($dir) {
...
+    return (Get-ChildItem $bucketsdir).Name
+}
+
 function buckets {
-    $buckets = @()
-    if(test-path $bucketsdir) {
-        Get-ChildItem $bucketsdir | ForEach-Object { $buckets += $_.Name }
-    }
-    return $buckets
+    return Get-LocalBucket
```
r15ch13: Would it be a good idea to add a deprecated warning by using something like this?

```pwsh
function Show-DeprecatedWarning($invocation, $new) {
    Write-Host ('"{0}" will be deprecated. Please change your code/manifest to use "{1}"' -f $invocation.MyCommand.Name,$new) -f DarkYellow -NoNewline
    Write-Host " -> $($invocation.PSCommandPath):$($invocation.ScriptLineNumber):$($invocation.OffsetInLine)" -f DarkGray
}


function buckets {
	Show-DeprecatedWarning $MyInvocation 'Get-LocalBucket'
    return Get-LocalBucket
}
```

Just for functions that _might_ get used in manifests.

Ash258: image

Done. Added it to core

Ash258: This will be nice start to proper changelogs after bucket folder extraction. There could be then normal github release of scoop with some changelog list, which could contain few parts (added, changed, deprecated)

lib/core.ps1
```diff
@@ -42,6 +42,22 @@ function Get-UserAgent() {
...
+    .PARAMETER Invocation
+        Invocation to identify location of line.
+        Just pass $MyInvocation.
+    .PARAMETER New
+        New command name.
+    #>
+    param($Invocation, [String] $New)
+
+    Write-Host ('"{0}" will be deprecated. Please change your code/manifest to use "{1}"' -f $Invocation.MyCommand.Name, $New) -ForegroundColor DarkYellow -NoNewline
+    Write-Host " -> $($invocation.PSCommandPath):$($invocation.ScriptLineNumber):$($invocation.OffsetInLine)" -ForegroundColor DarkGray
```
r15ch13: Should be `$Invocation`

Ash258: Right 😒 Changed it only in one line up

Ash258:
Maybe this look better than one line?

image

r15ch13:
Add `WARN  ` to make it the same as [`warn()`](https://github.com/lukesampson/scoop/blob/2181438ce27244e0c226fb9632a3912e30fceff1/lib/core.ps1#L61) and indent it.

image

Ash258:
Done. Used warn function directly for proper reference counter, when will warn function be refactored

image

Issue #1328, opened 2017-02-11
Separate main bucket from scoop code?


deevus:
Changes to manifests are for the most part unrelated to changes to the core scoop codebase. To reduce overlap, we could move the `bucket/` directory into its own repository. Thoughts?

lukesampson:
It would be tidier I suppose. Is it causing problems the way it is though? It might even slow `scoop update` down if we have to do two git pulls rather than two. There doesn't seem to be a big benefit to justify changing it just yet.

stkb:
Wondered about this as well. The way it is at the moment, it's cumbersome to run scoop from a working copy of your own fork, since you have to keep pulling in changes manually, just to keep the main bucket up to date.

rasa:
And we could sub-module the bucket directory, if we wanted. We could even design it to allow users to sub-module any other buckets they wanted, eliminating the need for 'scoop bucket add`

AntoineTurmel:
+1 it would be better so if you watch the scoop github repository you don't have notifications about packages, only the app itself ;)

Ash258:
Core extraction should be easy task to do and should not take too long. But there is one big question.

# How to deal with bucket being in other repository?

There are 3 main options. Submodule / Subtree and actual known bucket flow.

1. Known bucket flow would be most familiar to users
    - It would require few code tweaks.
        - Replace hardcoded bucket path
        - scoop update / checkup to reflect change and no need for users to do anything.
        - It would allow user to install from main bucket using `main/manifest` syntax, which is not possible right now
1. Submodule / Subtree
    - Easiest method, with single config file, but I don't know, how would this affect users.

# Extraction process

1. [ ] Postpone Excavator
1. [ ] Binaries
	- Remove default dir parameter
1. [ ] lib
	- All hardcoded path to core repository bucket folder:
		- <https://github.com/lukesampson/scoop/blob/589303facc5284f6f95c1305191e0558c0169691/test/Scoop-Manifest.Tests.ps1#L1>
		- <https://github.com/lukesampson/scoop/blob/589303facc5284f6f95c1305191e0558c0169691/test/Scoop-Manifest.Tests.ps1#L50>
		- <https://github.com/lukesampson/scoop/blob/0029f0b29dda9d82e72e1cf53183231cd8bf79d4/lib/buckets.ps1#L64>
[...]

PR #3343, opened 2019-04-20
Change LLVM to use shims instead of adding the bin directory to PATH

DaanDeMeyer:


DaanDeMeyer:
`git-clang-format` and `hmaptool` are python scripts and `scan-build.bat` invokes a perl script in the same directory. Should I remove those since they require other tools to be installed?

DaanDeMeyer:
@Ash258 Anything else I have to do to get this merged?

Ash258:
Just wait. There is major refactoring happening right now (#3060, #1328 ) so this will be merged inside https://github.com/scoopinstaller/scoop-main repository on Wednesday.

I will port this PR after wednesday automatically (with co author tag to give you credits)

PR #3350, opened 2019-04-21
[Regression] Fix search after buckets optimize

### LINKED ISSUE #3349: Cannot find path 'C:\Users\xxx\scoop\buckets' because it does not exist.
On a clean install scoop
```ps1
> scoop search vim

Get-ChildItem : Cannot find path 'C:\Users\xxx\scoop\buckets' because it does not exist.
At C:\Users\xxx\scoop\apps\scoop\current\lib\buckets.ps1:50 char:13
+     return (Get-ChildItem $bucketsdir).Name
+             ~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\Users\xxx\scoop\buckets:String) [Get-ChildItem]    , ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.GetChildItemCommand
 ```

Probably due to the recent change with buckets function. Is there any reason to remove ` if(test-path $bucketsdir)` ?

New Code:
https://github.com/lukesampson/scoop/blob/d0a1d4fe64e4e8194ba51b68b85438030fa2489c/lib/buckets.ps1#L44-L51

Old Code:
https://github.com/lukesampson/scoop/blob/54235efe5999328cf350e1d13ff4d70b3be51f7f/lib/buckets.ps1#L42-L48

### PR DESCRIPTION
**Pull request by Ash258**
- Closes #3349 

image

`Get-LocalBucket` will return $null if folder does not exists, or is empty. So if this function return something, add $null for main. With actual implementaion there was two runs of main bucket when empty `@($null) + @($null)`

### DIFF
--- a/lib/buckets.ps1
+++ b/lib/buckets.ps1
@@ -47,7 +47,7 @@ function Get-LocalBucket {
         List all local buckets.
     #>
 
-    return (Get-ChildItem $bucketsdir).Name
+    return (Get-ChildItem $bucketsdir -ErrorAction SilentlyContinue).Name
 }
 
 function buckets {
--- a/libexec/scoop-search.ps1
+++ b/libexec/scoop-search.ps1
@@ -97,7 +97,10 @@ function search_remotes($query) {
     }
 }
 
-@($null) + @(Get-LocalBucket) | ForEach-Object { # $null is main bucket
+$buckets = Get-LocalBucket
+if ($buckets) { $buckets += $null } # $null is main bucket
+
+$buckets | ForEach-Object {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by Ash258** [comment]
That main bucket is messing with me in each PR 😂

**Comment by Summon528** [comment]
Now it print an unwanted `Updating '' bucket...` when running `scoop update`  😂
```ps1 
> scoop update

Updating Scoop...
Updating '' bucket...
```

Ash258:
Okay, screw this. Lets make this when extracting main bucket. (#3531)

PR #3351, opened 2019-04-21
[buckets] Get bach to if

Ash258:
Unhandled if case is pretty no go, but this will work unless bucket extraction is done.


r15ch13:
Why not just `ensure $bucketsdir`? :grin:

Ash258:
Same problem as before. Then it will return $null which will "break" update, search, ...

PR #3372, opened 2019-04-25
fix(scoop-update): fix branch switching

### LINKED ISSUE #3368: [Bug] Switching branches do not work reliable
Bug

There should be latest commit with: `feature: Prepare extraction of main bucket (#3060)`

This needs to be fixed in master before merging develop with bucket extraction into master.

h404bi:
close #3368 

The hotfix should also be committed into the develop branch.

### DIFF
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -29,7 +29,7 @@
 reset_aliases
 
 $opt, $apps, $err = getopt $args 'gfiksq:' 'global', 'force', 'independent', 'no-cache', 'skip', 'quiet'
-if($err) { "scoop update: $err"; exit 1 }
+if ($err) { "scoop update: $err"; exit 1 }
 $global = $opt.g -or $opt.global
 $force = $opt.f -or $opt.force
 $check_hash = !($opt.s -or $opt.skip)
@@ -39,7 +39,7 @@ $independent = $opt.i -or $opt.independent
 
 # load config
 $repo = $(get_config SCOOP_REPO)
-if(!$repo) {
+if (!$repo) {
     $repo = "https://github.com/lukesampson/scoop"
     set_config SCOOP_REPO "$repo"
 }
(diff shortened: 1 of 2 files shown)


libexec/scoop-update.ps1
```diff
@@ -82,11 +82,18 @@ function update_scoop() {
         # Check if user configured other branch
         $branch = $(get_config SCOOP_BRANCH)
         if ((git_branch) -notlike "*$branch") {
```
Ash258: I would also change this condition into something like this:

```suggestion
        if (!((git_branch) -match "\*\s+$branch")) {
```

libexec/scoop-update.ps1
```diff
@@ -53,46 +54,52 @@ if (!$branch) {
...
-        if ((git_branch) -notlike "*$branch") {
-            git_fetch --all -q
+        # Change branch if user configured other branch
+        if (!((git_branch) -match "\*\s+$branch")) {
+            # reset git fetch refs (GH-3368)
+            git_config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
+            # fetch remote branches
+            git_fetch --force origin -q
+            # checkout and track the branch
             git_checkout -B $branch -q
```
niheaven:
```suggestion
            git_checkout -B $branch -t origin/$branch -q
```

Otherwise the new branch has no tracking information...

Ash258:
@niheaven @h404bi Did you both tested latest commit?

If i need to verify it myself.

h404bi:
Yeah, I tested the whole pr, don't know niheaven whether tested it or not.

niheaven:
Tested.

But I think it should add some error action cases, e.g.

- If user gives a wrong `SCOOP_BRANCH` name
- If user has a branch with the same name but incorrect upstreaming branch
- etc.

h404bi:
Few users would use this feature, except for scoop developer. It's just a fix for #3368 and #3060 , so no need to over design currently I guess.

PR #3373, opened 2019-04-25
gpg: remove temp files

dodorz:
remove $PLUGINSDIR dirs


niheaven:
Please submit manifest related PR in https://github.com/scoopinstaller/scoop-main, see https://github.com/scoopinstaller/scoop-main/pull/3 for reference.

PR #3375, opened 2019-04-25
buckets.ps1: Fix new_issue_msg

niheaven:
Fix error when hash check failed. `main` is a known bucket and will returned by `Find-Manifest`, so the exception is needless.


Ash258:
Nice catch. Missed this.

sschneider-ihre-pvs:
powershell don't like it

PR #3389, opened 2019-04-26
 BREAKING CHANGE: drop main bucket 

h404bi:
follow up https://github.com/lukesampson/scoop/pull/3060


h404bi:
untested

Ash258:
This could be easily done after merging develop to master.

This nested bucket will not be used by any function. So you could easily push develop in master (with actual state) and then after few hours / days remove nested bucket from master.

Do this later makes more sense if anything really bad happened. Since there will be smaller diff while reverting 😂

niheaven:
So the other one is MUST, this one is MAY...

h404bi:
So when the new main bucket exists with nested bucket, which one would be used?

niheaven:
The new one, since bucket functions find manifests in buckets dir, and with app name likes `7zip`, it looks at `main`.

Ash258:
Always that one from `Find-Manifest` function, which has default main bucket located inside `$bucketsdir`

PR #3413, opened 2019-04-29
checkver.ps1: Fix example parameters

krokofant:


r15ch13:
Comments lie every time! :grin:

Ash258:
First few comments will not work. Since `$Dir` Parameter is mandatory. So `checkver.ps1 -Update` will fail (or interactively prompt for path) because you did not specified `$Dir`.

A

Issue #3303, opened 2019-04-04
Adding shims to an installation


garytyler:
Does the scoop cli include a command for adding shims to an installation? Or creating custom shims? I would like to add version appendages to my python installations. For example: `python2` -> `python27`, `python2.7`. Is there a recommended way to go about it?

Sorry if this has been asked before! I looked and couldn't find anything.

r15ch13:
If you installed python from [main/python](https://github.com/lukesampson/scoop/blob/master/bucket/python.json) and [versions/python27](https://github.com/scoopinstaller/versions/blob/master/bucket/python27.json) then `python3` and `python2` shims should already exist.

garytyler:
@r15ch13 Both `python3` and `python2` do exist, so I'm hoping to add the other variations, most particularly `python2.7` and `python3.7, to reflect a typical python CI environment expected by the test automation software `tox`. I usually copy the existing shims, but was just wondering if there is another solution. I know there is the version switcher functionality, and adding the shims might be more involved than just entering them into the manifests, but would you consider a PR for that? If the intention here is to just keep things simple, that's completely understandable. Thanks for the reply, and for this great tool!

PR #3438, opened 2019-05-06
feature(core): Add scoop-shim helper

r15ch13:
This adds support for the extracted [Shim](https://github.com/ScoopInstaller/Shim/).exe

Related: https://github.com/lukesampson/scoop/pull/3066

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -194,7 +194,7 @@ function Get-HelperPath {
     [CmdletBinding()]
     param(
         [Parameter(Mandatory = $true, Position = 0, ValueFromPipeline = $true)]
-        [ValidateSet('7zip', 'Lessmsi', 'Innounp', 'Dark', 'Aria2')]
+        [ValidateSet('7zip', 'Lessmsi', 'Innounp', 'Dark', 'Aria2', 'Shim')]
         [String]
         $Helper
     )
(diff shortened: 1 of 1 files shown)


lib/core.ps1
```diff
@@ -427,8 +428,11 @@ function shim($path, $global, $name, $arg) {
     }
 
     if($path -match '\.(exe|com)$') {
+        if(!(Test-HelperInstalled -Helper Shim)) {
+            scoop install 'main/scoop-shim'
```
Ash258: Manifest is missing binary. Is it intentional?

r15ch13: Yes, what would be the purpose of shimming the shim? :smile:

Ash258: You can use it then from anywhere and shim anything. Nice for users who want to shim custom executables.

Small workaround for #3303

r15ch13:
The robocopy tests fail because they require a shimmed file.
I also don't know of the current position in code is a good place for the installation.
It should probably be installed on update and the first scoop installation.

lib/core.ps1
r15ch13: I would like to avoid recursions :grin:

It could just be copied from `~/scoop/apps/scoop-shim/current/shim.exe`

rashil2000:
Closing in favour of #4736

PR #3450, opened 2019-05-11
Add Expand-DarkArchive and some other dependency features

r15ch13:
- Add `Expand-DarkArchive` helper function
- Allow `dark.exe` from wixtoolset and `7zip-zstd` as valid helpers
- Add warnings to 'scoop checkup' for missing unpackers
- Auto detect dependencies from `pre_install`/`post_install` and `installer.script` by simply searching for the function names.

Should be merge into `develop` before merging `develop` into `master`.
This will allow us to remove the `dark` dependency from some manifests.
`Expand-7zipArchive` is widely used without the manifest actually depending on it. 
The installation will fail with the following message when `Get-HelperPath` returns nothing.
```
The expression after '&' in a pipeline element produced an object that was not valid. It must result in a command name, a script block, or a CommandInfo object.
```

https://github.com/lukesampson/scoop/issues/3417

### DIFF
--- a/libexec/scoop-checkup.ps1
+++ b/libexec/scoop-checkup.ps1
@@ -13,6 +13,30 @@ $issues += !(check_windows_defender $true)
 $issues += !(check_main_bucket)
 $issues += !(check_long_paths)
 
+if (!(Test-HelperInstalled -Helper 7zip)) {
+    error "'7-Zip' is not installed! It's required for unpacking most programs. Please Run 'scoop install 7zip' or 'scoop install 7zip-zstd'."
+    $issues++
+}
+
+if (!(Test-HelperInstalled -Helper Lessmsi)) {
+    error "'lessmsi' is not installed! It's required for unpacking programs with MSI packages. Please run 'scoop install lessmsi'."
+    $issues++
+}
+
+if (!(Test-HelperInstalled -Helper Innounp)) {
+    error "'Inno Setup Unpacker' is not installed! It's required for unpacking InnoSetup files. Please run 'scoop install innounp'."
+    $issues++
+}
+
+if (!(Test-HelperInstalled -Helper Dark)) {
+    error "'dark' is not installed! It's require for unpacking installers created with the WiX Toolset. Please run 'scoop install dark' or 'scoop install wixtoolset'."
+    $issues++
+}
+
+if (!(Test-HelperInstalled -Helper Aria2)) {
+    warn "'aria2' is not installed! This can improve download speeds. You may run 'scoop install aria2'."
+}
+
 $globaldir = New-Object System.IO.DriveInfo($globaldir)
 if($globaldir.DriveFormat -ne 'NTFS') {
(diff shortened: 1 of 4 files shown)


libexec/scoop-checkup.ps1
```diff
@@ -13,6 +13,30 @@ $issues += !(check_windows_defender $true)
...
+    error "'Inno Setup Unpacker' is not installed! It's required for unpacking InnoSetup files. Please run 'scoop install innounp'."
+    $issues++
+}
+
+if (!(Test-HelperInstalled -Helper Dark)) {
+    error "'dark' is not installed! It's require for unpacking installers created with the WiX Toolset. Please run 'scoop install dark' or 'scoop install wixtoolset'."
+    $issues++
+}
+
+if (!(Test-HelperInstalled -Helper Aria2)) {
```
Ash258: Remove this check for aria2.

Forcing users to install aria2 is nonsense.

libexec/scoop-checkup.ps1
```diff
@@ -13,6 +13,30 @@ $issues += !(check_windows_defender $true)
 $issues += !(check_main_bucket)
 $issues += !(check_long_paths)
 
+if (!(Test-HelperInstalled -Helper 7zip)) {
+    error "'7-Zip' is not installed! It's required for unpacking most programs. Please Run 'scoop install 7zip' or 'scoop install 7zip-zstd'."
+    $issues++
+}
+
+if (!(Test-HelperInstalled -Helper Lessmsi)) {
```
Ash258: Less msi is optional dependency. Should not be reported.

libexec/scoop-checkup.ps1
r15ch13: It's a warning and doesn't increase the issue count.

r15ch13:
```powershell
if (!(Test-HelperInstalled -Helper Lessmsi)) {
    warn "'lessmsi' is not installed. (Optional) This may resolve problems MSI packages. You may run 'scoop install lessmsi'."
}
```

libexec/scoop-checkup.ps1
r15ch13:
```powershell
if (!(Test-HelperInstalled -Helper Lessmsi)) {
    warn "'lessmsi' is not installed. (Optional) This may resolve problems MSI packages. You may run 'scoop install lessmsi'."
}
```

libexec/scoop-checkup.ps1
Ash258: I know, but it is irrelevant.

Me, as user who do not want to use aria2, now always see this warning when running checkup. Sure i can intall it and disable it, but it is completely non user friendly.

libexec/scoop-checkup.ps1
Ash258: Again. Same as for aria2.

When i do not want to use lessmsi i have no option how to disable this check.

niheaven:
`dark` needn't a seperated `Expand-` function since it is only used when we want to unpack `WIX` type installer. After using dark, we need `extract_msi`...

r15ch13:
@niheaven using this function will make manifest scripts cleaner and it's easier to use.

PR #3452, opened 2019-05-12
Neovim: Add an entry for the nightly version of the package.

bgianfo:
Add a new neovim-nightly.json to allow users to install the nightly
builds of neovim, instead of just the stable/release builds.


bgianfo:
Made the PR to the new Main repro instead.. 
https://github.com/ScoopInstaller/Main/pull/50

PR #3458, opened 2019-05-13
feat(manifest): xpath support in checkver and autoupdate

### LINKED ISSUE #3451: XPath support in checkver and autoupdate
Feature: Yes
Priority: Middle

example:
```json
    "checkver": {
        "url": "https://example.com/update.xml",
        "xpath": "/App/version"
    }
```

PowerShell natively supports XML Path, so it could be implemented without difficulty.

h404bi:
Close #3451 

TODO list:

- [x] `xpath` in checkver
- [x] `xpath` in autoupdate hash extraction
- [x] `xpath` in json schema

*to be supplemented*

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -233,6 +233,11 @@ while ($in_progress -gt 0) {
         $page = $ver
         $ver = ''
     }
+    
+    if ($xpath -and $regexp) {
+        $page = $ver
+        $ver = ''
+    }
 
     if ($regexp) {
         $regex = New-Object System.Text.RegularExpressions.Regex($regexp)
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -184,7 +184,6 @@ while ($in_progress -gt 0) {
     Remove-Event $ev.SourceIdentifier
     $in_progress--
 
-
     $state = $ev.SourceEventArgs.UserState
     $app = $state.app
     $json = $state.json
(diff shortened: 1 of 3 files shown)


bin/checkver.ps1
```diff
@@ -214,6 +220,15 @@ while ($in_progress -gt 0) {
...
+    if ($xpath) {
+        # Getting version from XML, using XPath
+        $ver = ([xml]$page).SelectSingleNode($xpath).'#text'
+        if (!$ver) {
+            next "couldn't find '$xpath' in $url"
+            continue
+        }
+    }
+
     if ($jsonpath -and $regexp) {
```
Ash258: Would be nice to implement concurent usage of `xpath` and `regex`.

(Example: https://github.com/lukesampson/scoop-extras/blob/ec81f5723b3d53bdd63b42b203bf41537607925f/bucket/eclipse-testing.json)

There you would need `xpath` to get version and `regex` to strip that `/R` from it.

h404bi: I don't quite understand the use-case you provided, do you mean to add the follow codes?

```powershell
    if ($xpath -and $regexp) {
        $page = $ver
        $ver = ''
    }
```

Ash258: Yes. Because then flow will be next: 

1. `xpath` will be used to get version.
1. `regex` (could look like `(.*)/R`) will be used
1. Version will be applied to first group of regex

But version needs to strip that `/R`

Ash258:
@h404bi Found one problematic xml file

According to multiple xpath testers and evaluators this is valid expression:
XML = https://sparkle.musescore.org/stable/3/win/appcast.xml
XPATH = `/rss/channel/item/enclosure/@sparkle:version`

PR #3465, opened 2019-05-15
fix(autoupdate): Handle xml namespace in xpath mode

h404bi:
Fix https://github.com/lukesampson/scoop/pull/3458#issuecomment-492570939

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -221,8 +221,16 @@ while ($in_progress -gt 0) {
     }
 
     if ($xpath) {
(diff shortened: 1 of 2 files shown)


r15ch13:
Ah, I was on the right track with the namespace thing in my old test branch :smile:

h404bi:
It took me 2 hours to find the way to resolve the issue of xml namespace without giving the power to users to control the namespace input.

Ash258:
Excelent 🎉😍

Nice

Issue #3080, opened 2019-02-05
Scoop doesn't extract .tar.* archives fully


Calinou:
If one writes an URL to a `.tar.*` archive (such as `.tar.gz`) in a manifest's `url` field, Scoop will download it but will only extract it to an `xa-tmp.tar` file. To fix this, 7-zip needs to be called twice on those files, so that the `.tar` is unpacked to its actual data.

asmgf:
@Calinou, it seems that extraction should work.

Double extraction logic is already there: https://github.com/lukesampson/scoop/blob/2224da82f8a786b8ae9c203c67924b4c6ef2c652/lib/decompress.ps1#L24-L28

There are also a few manifests that point to `.tar.gz` archive, for example `up`: https://github.com/lukesampson/scoop/blob/a2fed1855865d3d2d8f0eb6564c52163db650177/bucket/up.json#L7
I tried running `scoop install up` and it worked fine for me.

Ash258:
@Calinou  Is it OK or is some fix needed?

Calinou:
This is the manifest I'm trying it with:

```json
{
    "homepage": "https://sourceforge.net/projects/freecs-1-5/",
    "description": "Free software reimplementation of Counter-Strike 1.5 running on FTEQW",
    "version": "2019-01-16",
    "license": "GPL-2.0-or-later",
    "url": [
        "https://downloads.sourceforge.net/project/freecs-1-5/freecs-bin-2019-01-16.tar.gz",
        "https://downloads.sourceforge.net/project/freecs-1-5/cstrike-1.5-content.tar.gz"
    ],
    "hash": [
        "34c48ef1b300a811ebf19b975be6b43c6bb2182f780bd0718b1b1ce93ca3600b",
        "71929123b9e823ca7702f5a8c9cb53db3ac5e889785b56f53343278d14def81d"
    ],
    "notes": [
        "Optional but recommended: To avoid missing assets, place an Half-Life installation's \"valve\" folder in:",
        "",
        "\t$persist_dir\\"
    ]
}
```

Console log:

```
Installing 'freecs' (2019-01-16) [64bit]
Loading freecs-bin-2019-01-16.tar.gz from cache
Checking hash of freecs-bin-2019-01-16.tar.gz ... ok.
Loading cstrike-1.5-content.tar.gz from cache
Checking hash of cstrike-1.5-content.tar.gz ... ok.
Extracting freecs-bin-2019-01-16.tar.gz ... done.
[...]

PR #3466, opened 2019-05-16
feature(decompress): Add 'ExtractDir' to 'Expand-...' functions

niheaven:
Move `$extract_dir` and `$extract_to` to `decompress.ps1` so as to not run `robocopy` so many times.

Tested on

- 'slack'(w/ `extract_dir`, 7z/nupkg)
- 'brackets'(w/ `extract_dir`, msiexec/lessmsi)
- 'atom-portable'(w/ `extract_dir` and `extract_to`)
- 'irfanview'(w/ multi `extract_to`)
- 'unxutils'(w/ `extract_to` overwrite)

### DIFF
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -44,6 +44,8 @@ function Expand-7zipArchive {
         [Parameter(Position = 1)]
         [String]
         $DestinationPath = (Split-Path $Path),
+        [String]
+        $ExtractDir,
         [Parameter(ValueFromRemainingArguments = $true)]
         [String]
         $Switches,
@@ -53,6 +55,7 @@ function Expand-7zipArchive {
         [Switch]
         $Removal
     )
+    $DestinationPath = $DestinationPath.TrimEnd("\")
     if ((get_config 7ZIPEXTRACT_USE_EXTERNAL)) {
         try {
             $7zPath = (Get-Command '7z' -CommandType Application | Select-Object -First 1).Source
@@ -64,6 +67,9 @@ function Expand-7zipArchive {
     }
     $LogPath = "$(Split-Path $Path)\7zip.log"
     $ArgList = @('x', "`"$Path`"", "-o`"$DestinationPath`"", '-y')
+    if ($ExtractDir) {
+        $ArgList += "-ir!$ExtractDir\*"
+    }
     if ($Switches) {
         $ArgList += (-split $Switches)
     }
@@ -74,7 +80,10 @@ function Expand-7zipArchive {
     }
     $Status = Invoke-ExternalCommand $7zPath $ArgList -LogPath $LogPath
     if (!$Status) {
-        abort "Failed to extract files from $Path.`nLog file:`n  $(friendly_path $LogPath)"
+        abort "Failed to extract files from $Path.`nLog file:`n  $(friendly_path $LogPath)`n$(new_issue_msg $app $bucket 'decompress error')"
+    }
+    if ($ExtractDir) {
+        movedir "$DestinationPath\$ExtractDir" $DestinationPath | Out-Null
     }
     if (Test-Path $LogPath) {
         Remove-Item $LogPath -Force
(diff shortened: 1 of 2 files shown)


lib/decompress.ps1
```diff
@@ -149,20 +167,23 @@ function Expand-InnoArchive {
...
         [Parameter(ValueFromRemainingArguments = $true)]
         [String]
         $Switches,
         [Switch]
         $Removal
     )
+    $DestinationPath = $DestinationPath.TrimEnd("\")
     $LogPath = "$(Split-Path $Path)\innounp.log"
-    $ArgList = @('-x', "-d`"$DestinationPath`"", "-c`{app`}", "`"$Path`"", '-y')
+    $ArgList = @('-x', "-d`"$DestinationPath`"", "-c`{app`}\$ExtractDir", "`"$Path`"", '-y')
```
Ash258: db2 list database directory show detail

niheaven: It will work, but `$DestinationPath` in some functions will not work... So I add it to every functions.

Ash258:
LGTM by fast static review. Will test in few hours.

niheaven:
Tested 'TrimEnd', `msiexec` and `lessmsi` need to trim `\`, others needn't.

Also fix nested 7z extraction.

All done.

lib/decompress.ps1
```diff
@@ -74,17 +80,20 @@ function Expand-7zipArchive {
...
     }
     if (Test-Path $LogPath) {
         Remove-Item $LogPath -Force
     }
-    if ((strip_ext $Path) -match '\.tar$' -or $Path -match '\.t[abgpx]z2?$') {
+    if ($IsTar) {
         # Check for tar
         $Status = Invoke-ExternalCommand $7zPath @('l', "`"$Path`"") -LogPath $LogPath
         if ($Status) {
             $TarFile = (Get-Content -Path $LogPath)[-4] -replace '.{53}(.*)', '$1' # get inner tar file name
```
r15ch13: This `[-4]` thing doesn't work for me.
```
Y:\
λ cat bla.txt

7-Zip 19.00 (x64) : Copyright (c) 1999-2018 Igor Pavlov : 2019-02-21

Scanning the drive for archives:
1 file, 1804230 bytes (1762 KiB)

Listing archive: .\radare2-extras-3.5.0.tar.gz

--
Path = .\radare2-extras-3.5.0.tar.gz
Type = gzip
Headers Size = 10

   Date      Time    Attr         Size   Compressed  Name
------------------- ----- ------------ ------------  ------------------------
                    .....     12390400      1804230  radare2-extras-3.5.0.tar
------------------- ----- ------------ ------------  ------------------------
                              12390400      1804230  1 files
Y:\
λ (Get-Content -Path 'bla.txt')[-4] -replace '.{53}(.*)', '$1'
------------------------
Y:\
```

Most of the tar files have a double extension, so why not use `[IO.Path]::GetFileNameWithoutExtension($Path)` instead?

Ash258: > Most of the tar files have a double extension, so why not use [IO.Path]::GetFileNameWithoutExtension($Path) instead?

This is for use cases when the nested filename is not equal to archive name. #3080

r15ch13: Okay. But just using a line index isn't reliable.
In https://github.com/lukesampson/scoop/commit/079c4a57269a2a570e29d4326fcb615e98f0b0b4 it was `-3` no it's `-4` 😕

Ash258: I hoped it is static as asked in https://github.com/lukesampson/scoop/pull/3204#pullrequestreview-222929383

@niheaven Why the index changed?

r15ch13: Also, the `$IsTar` check doesn't support `.tar.gz` or `.tar.xz`.

niheaven: `(strip_ext $Path) -match '\.tar$')` will fit all `.tar.xxx`, and `.tZ` is not supported by 7zip. But you regex is perfect :)

r15ch13: Missed the `strip_ext` ... I take everything back 😄

niheaven: > Okay. But just using a line index isn't reliable.
> In [079c4a5](https://github.com/lukesampson/scoop/commit/079c4a57269a2a570e29d4326fcb615e98f0b0b4) it was `-3` now it's `-4` 😕

The `Invoke-ExternalCommand` write log to a file which has a blank end line...

So if you save the output to variable it will be the last 3 line while for file it is last 4 line.

Could you try `curl`'s installation? I've tested several tar, and [-4] is the right line number.

r15ch13: Works fine 👍

lib/decompress.ps1
r15ch13: A bit late but:
`$ExtractDir` should use `{app}` as a default value instead of hardcoding it to `{app}\$ExtractDir`.
For example, fixing `freedownloadmanager` requires files from `{app}` and `{code_CefInstallDir}` to be extracted and merged.
This hardcoded value would require running:
```pwsh
Invoke-ExternalCommand (Get-HelperPath -Helper Innounp) @('-x', "-d`"$DestinationPath`"", -c"`{app`}", "`"$Path`"", '-y')
Invoke-ExternalCommand (Get-HelperPath -Helper Innounp) @('-x', "-d`"$DestinationPath`"", -c"`{code_CefInstallDir`}", "`"$Path`"", '-y')
```
instead of simply:
```pwsh
Expand-InnoArchive -Path \"$dir\\$fname\" -ExtractDir "{app}"
Expand-InnoArchive -Path \"$dir\\$fname\" -ExtractDir "{code_CefInstallDir}"
```

See:
https://github.com/lukesampson/scoop-extras/pull/2137
https://github.com/lukesampson/scoop-extras/issues/2150

Ash258: Same for GOG Galaxy

https://github.com/Ash258/scoop-Ash258/blob/master/bucket/GOGGalaxy.json

Ash258: Same for https://github.com/Ash258/scoop-Ash258/blob/master/bucket/GOGGalaxy.json

Could be simplified to just two calls of `Expand-InnoArchive`

PR #3333, opened 2019-04-18
fix(versions): Refactor `compare_versions` -> `Compare-Version`

niheaven:
fix #3329, fix #2981, fix #2720, fix https://github.com/lukesampson/scoop-extras/issues/2106

Now force updating to manifest's version, despite installed version is greater or less than it.

PR #3469, opened 2019-05-16
Improved compare_versions function

edgardmessias:
This is a suggestion to solve some problems with version.

Example: PyCharm-Professional
Current version is `2019.1.2-191.7141.48`, but the command `scoop list` and `scoop update PyCharm-Professional` show old version installed `2019.1-191.6183.50`.

Explain: Current function, when compare, use `[2019,1,191,6,6183,50]` vs `[2019,1,2,191,7141,48]`, the `191` is greater than `2`

My idea: Split first by "-", and compare each group first (Ex.: `2019.1.2` vs `2019.1`)

It work for date and version with text, ex.: `2019-05-16`, `2.2.1-beta1`

### DIFF
--- a/lib/versions.ps1
+++ b/lib/versions.ps1
@@ -18,7 +18,7 @@ function version($ver) {
         if($num) { $num } else { $_ }
     }
 }
-function compare_versions($a, $b) {
+function compare_versions_base($a, $b) {
     $ver_a = @(version $a)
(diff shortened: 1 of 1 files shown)


Ash258:
#3333

Ash258:
Supressed by #3333

PR #3487, opened 2019-05-30
feat(compress): Allow 'Expand-InnoArchive -ExtractDir' to accept '{xxx}'

niheaven:
ref: https://github.com/lukesampson/scoop/pull/3466#discussion_r288691859

Now `freedownloadmanager`'s `installer.script` should be simplified to (see https://github.com/lukesampson/scoop-extras/pull/2220)

```json
"installer": {
    "script": [
        "Expand-InnoArchive -Path \"$dir\\$fname\"",
        "Expand-InnoArchive -Path \"$dir\\$fname\" -ExtractDir '{code_CefInstallDir}' -Removal"
    ]
}
```

### DIFF
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -180,7 +180,12 @@ function Expand-InnoArchive {
         $Removal
     )
     $LogPath = "$(Split-Path $Path)\innounp.log"
-    $ArgList = @('-x', "-d`"$DestinationPath`"", "-c`{app`}\$ExtractDir", "`"$Path`"", '-y')
+    $ArgList = @('-x', "-d`"$DestinationPath`"", "`"$Path`"", '-y')
+    switch -Regex ($ExtractDir) {
+        "^[^{].*" { $ArgList += "-c{app}\$ExtractDir" }
+        "^{.*" { $ArgList += "-c$ExtractDir" }
+        Default { $ArgList += "-c{app}" }
+    }
     if ($Switches) {
         $ArgList += (-split $Switches)
     }


lib/decompress.ps1
```diff
@@ -180,7 +180,12 @@ function Expand-InnoArchive {
         $Removal
     )
     $LogPath = "$(Split-Path $Path)\innounp.log"
-    $ArgList = @('-x', "-d`"$DestinationPath`"", "-c`{app`}\$ExtractDir", "`"$Path`"", '-y')
+    $ArgList = @('-x', "-d`"$DestinationPath`"", "`"$Path`"", '-y')
```
Ash258: I think single quotes could be used instead of escaping double quote.

```suggestion
    $ArgList = @('-x', "-d'$DestinationPath'", "'$Path'", '-y')
```

niheaven: I'm not sure if single quote will expand the variables, please wait for my testing.

niheaven:
I've tested for freedownloadmanager, and it works fine. See attached PR.

lib/decompress.ps1
Ash258: It is inside double quote.

Powershell should evaluate the variable first and then pass the string with evaluated value as argument. So it should be called as "innounp -x .. -d'actualPath\whatever'"

niheaven: I've tested single quote, and it will break the extraction... Escaped double quotes are MUST here...

And I also notice that, for `-cDIR`, if `DIR` has white spaces, it needs double quotes too, but commonly there won't be any white spaces for either `{app}\xxx` or `{xxx}`. If meet so, I'll make a hotfix.

Issue #3401, opened 2019-04-28
Deleting SCOOP_BRANCH not switch to master


Ash258:
Extracted from #3397 

> Well apparently deleting the SCOOP_BRANCH config variable did not reset the branch to master correctly. Can't seem to reproduce the issue.

Since SCOOP_BRANCH config is default and created always on installation, it is required for functioning. Since we can't forbid users to delete it we need to adjust it in code.

Should be enough to replace `get_config SCOOP_BRANCH` with `get_config SCOOP_BRANCH 'master'` to return default master, when config value do not exists.

Ash258:
We should create some Hashtable with defaults which would be used in codebase.

```powershell
$CONFIG_DEFAULT = @{
    'SCOOP_BRANCH' = 'master'
    '...LESS_MSI' = $false
}
```

So we do not need to duplicate code and could easily change it if needed.

niheaven:
I like the idea. See discussion before: https://github.com/lukesampson/scoop/pull/3242#issuecomment-474013729

PR #3491, opened 2019-05-31
Command to list all scoop config settings

crheidri:
It would be nice to have a scoop config 'list' option to easily see all config settings that have been set by the user. Something similar to `git config --list` for git. As of now, you have to cherry-pick individual scoop configs from the command line:

> PS C:\Users\crheidri\projects\scoop> scoop config debug true
'debug' has been set to 'true'
PS C:\Users\crheidri\projects\scoop> scoop config debug
True

Proposed changes:
- Create `list_config()` function in core.ps1 that iterates over `$scoopConfig` and prints 'name=value' pairs.
- Add 'list' elseif statement to scoop-config.ps1 that calls `list_config()`.
- Update help message in scoop-config.ps1 to include 'list' option.

After these changes, the user can see all configuration variables at once:
> PS C:\Users\crheidri\projects\scoop> .\bin\scoop config list
debug=True
lastupdate=2019-05-31T14:35:54.2863076-05:00
SCOOP_BRANCH=master
SCOOP_REPO=https://github.com/lukesampson/scoop

One potential conflict is the word "list" can no longer be used as a config name.

### DIFF
--- a/default/config.json
+++ b/default/config.json
@@ -1,0 +1,4 @@
+{
+    "SCOOP_BRANCH": "master",
+    "SCOOP_REPO": "https://github.com/lukesampson/scoop"
+}
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -48,8 +48,13 @@ function load_cfg($file) {
 }
 
 function get_config($name, $default) {
-    if($null -eq $scoopConfig.$name -and $null -ne $default) {
-        return $default
+    if($null -eq $scoopConfig.$name) {
+        if($null -ne $default) {
+            Show-DeprecatedWarning $MyInvocation "get_config <name> (specify defaults in default/config.json)"
+            return $default
+        } else {
+            return $CONFIG_DEFAULT.$name
+        }
     }
     return $scoopConfig.$name
 }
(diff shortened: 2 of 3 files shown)


Ash258:
You are talking about `All` settings, but you are showing only that ones, which are configured. 

See https://github.com/lukesampson/scoop/issues/3401#issuecomment-487363809 for implemetation

crheidri:
> See #3401 (comment) for implemetation

The default config is a good idea. It might be nice to have a way to list configs set by default and those set by the user. I can take a shot at it if no one is working on that idea.

crheidri:
Pushed a Proof-of-Concept loading default configs. Feedback is very much welcome.

- Created default/config.json to store default scoop configs (this can be changed).
- core.ps1 loads the default configs into `$CONFIG_DEFAULT`
- `get_config $name` looks for `$name` in the user settings (`$scoopConfig`), and otherwise returns default config e.g. `$CONFIG_DEFAULT.$name`.
- Shows deprecated warning if `$default` is passed in as `get_config $name $default`.

@Ash258 `$DEFAULT_CONFIG` could still be a hashtable like in https://github.com/lukesampson/scoop/issues/3401#issuecomment-487363809, but I wasn't sure where to store it. It seems easier versioning a separate file instead of hard-coding the default configs into core.ps1.

I'd still have to refactor all the calls to get_config that pass in `$default`, and put those in default/config.json. Waiting to see if this idea is going the right direction.

PR #3503, opened 2019-06-11
feat(install): Add arch support to `env_add_path` and `env_set`

niheaven:
- Fix https://github.com/lukesampson/scoop/issues/3609

Just like `url`, `bin`, `shorcuts`, etc.

For MiKTeX that 64bit and 32bit offer different linked bin folder (which needed to be added to `PATH`)


niheaven:
@r15ch13 Please consider merging this into `master` directly to fix MiKTeX error :smile:

PR #3518, opened 2019-06-19
feat(autoupdate): add multiple URL/hash/extract_dir... support for autoupdate

niheaven:
Inherit #2803 

If hash extraction is different for urls, use array of objects. If hash extraction array's length is less then autoupdate.url's length, latest hash item is reused for the rest.

<details>
<summary>Example Manifest (irfanview)</summary>

```json
{
    "version": "4.53",
    "license": {
        "identifier": "Freeware",
        "url": "https://www.irfanview.com/eula.htm"
    },
    "homepage": "https://www.irfanview.com/",
    "architecture": {
        "64bit": {
            "url": [
                "https://www.irfanview.info/files/iview453_x64.zip",
                "https://www.irfanview.info/files/iview453_plugins_x64.zip"
            ],
            "hash": [
                "941056da4f50df44eeef46ca875255bd3d09c3aaf1bdfffad5f65de0a515ff02",
                "d5e8ae1734264287e7b442fc062899c206c54bbd423304e125c4b1d1c58d28b1"
            ],
            "shortcuts": [
                [
                    "i_view64.exe",
                    "IrfanView"
                ]
            ],
            "bin": [
                [
                    "i_view64.exe",
                    "irfanview"
                ]
            ]
        },
        "32bit": {
            "url": [
                "https://www.irfanview.info/files/iview453.zip",
                "https://www.irfanview.info/files/iview453_plugins.zip"
            ],
            "hash": [
                "89c0d29424299f9b0d36248884059baf815f0b5d4adc9c3b426f61fe68f1444a",
                "1ec241d26a0f061aa5499c606967690de8979c82fdce4a5724b4e18daa099de3"
            ],
            "shortcuts": [
                [
                    "i_view32.exe",
                    "IrfanView"
                ]
            ],
            "bin": [
                [
                    "i_view32.exe",
                    "irfanview"
                ]
            ]
        }
    },
    "pre_install": [
        "if (!(Test-Path(\"$dir/i_view64.ini\"))) {",
        "   New-Item \"$dir/i_view64.ini\" | Out-Null",
        "}",
        "if (!(Test-Path(\"$dir/i_view32.ini\"))) {",
        "   New-Item \"$dir/i_view32.ini\" | Out-Null",
        "}"
    ],
    "persist": [
        "i_view64.ini",
        "i_view32.ini",
        "Languages"
    ],
    "extract_to": [
        "",
        "Plugins"
    ],
    "checkver": {
        "url": "https://www.irfanview.com/",
        "re": ">\\s*version\\s+([\\d.-]+)\\s*<"
    },
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": [
                    "https://www.irfanview.info/files/iview$cleanVersion_x64.zip",
                    "https://www.irfanview.info/files/iview$cleanVersion_plugins_x64.zip"
                ],
                "hash": {
                    "url": "https://www.irfanview.com/64bit.htm",
                    "regex": "(?sm)$basename.*?$sha256"
                }
            },
            "32bit": {
                "url": [
                    "https://www.irfanview.info/files/iview$cleanVersion.zip",
                    "https://www.irfanview.info/files/iview$cleanVersion_plugins.zip"
                ],
                "hash": [
                    {
                        "url": "https://www.irfanview.com/main_download_engl.htm",
                        "regex": "(?sm)$basename.*?$sha256"
                    },
                    {
                        "url": "https://www.irfanview.com/plugins.htm",
                        "regex": "(?sm)$basename.*?$sha256"
                    }
                ]
            }
        }
    }
}

```

</details>

Please review, and CI test cases will be added soon.

### DIFF
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -284,43 +284,90 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
     return $hash
 }
 
-function update_manifest_with_new_version($json, [String] $version, $url, $hash, $architecture = $null) {
-    $json.version = $version
-
-    if ($null -eq $architecture) {
-        if ($json.url -is [System.Array] -and $url -isnot [System.Array]) {
-            $json.url[0] = $url
-            $json.hash[0] = $hash
-        } else {
-            $json.url = $url
-            $json.hash = $hash
-        }
-    } else {
-        # If there are multiple urls we replace the first one
-        if ($json.architecture.$architecture.url -is [System.Array] -and $url -isnot [System.Array]) {
-            $json.architecture.$architecture.url[0] = $url
-            $json.architecture.$architecture.hash[0] = $hash
-        } else {
-            $json.architecture.$architecture.url = $url
-            $json.architecture.$architecture.hash = $hash
-        }
+function Update-ManifestProperty {
+    <#
+    .SYNOPSIS
+        Update propert(y|ies) in manifest
+    .DESCRIPTION
+        Update selected propert(y|ies) to given version in manifest.
+    .PARAMETER Manifest
+        Manifest to be updated
+    .PARAMETER Property
+        Selected propert(y|ies) to be updated
+    .PARAMETER AppName
+        Software name
+    .PARAMETER Version
+        Given software version
+    .PARAMETER Substitutions
+        Hashtable of internal substitutable variables
+    .OUTPUTS
+        System.Boolean
+            Flag that indicate if there are any changed properties
+    #>
+    [CmdletBinding(SupportsShouldProcess = $true)]
+    [OutputType([Boolean])]
+    param (
+        [Parameter(Mandatory = $true, Position = 1)]
+        [PSCustomObject]
+        $Manifest,
+        [Parameter(ValueFromPipeline = $true, Position = 2)]
+        [String[]]
+        $Property,
+        [String]
+        $AppName,
+        [String]
+        $Version,
+        [Alias('Matches')]
+        [HashTable]
+        $Substitutions
+    )
+    begin {
+        $hasManifestChanged = $false
     }
-}
-
-function update_manifest_prop([String] $prop, $json, [Hashtable] $substitutions) {
-    # first try the global property
-    if ($json.$prop -and $json.autoupdate.$prop) {
-        $json.$prop = substitute $json.autoupdate.$prop $substitutions
-    }
-
-    # check if there are architecture specific variants
-    if ($json.architecture -and $json.autoupdate.architecture) {
-        $json.architecture | Get-Member -MemberType NoteProperty | ForEach-Object {
-            $architecture = $_.Name
-            if ($json.architecture.$architecture.$prop -and $json.autoupdate.architecture.$architecture.$prop) {
-                $json.architecture.$architecture.$prop = substitute (arch_specific $prop $json.autoupdate $architecture) $substitutions
+    process {
+        foreach ($currentProperty in $Property) {
+            if ($currentProperty -eq 'hash') {
+                # Update hash
+                if ($Manifest.hash) {
+                    # Global
+                    $newURL = substitute $Manifest.autoupdate.url $Substitutions
+                    $newHash = HashHelper -AppName $AppName -Version $Version -HashExtraction $Manifest.autoupdate.hash -URL $newURL -Substitutions $Substitutions
+                    $Manifest.hash, $hasPropertyChanged = PropertyHelper -Property $Manifest.hash -Value $newHash
+                    $hasManifestChanged = $hasManifestChanged -or $hasPropertyChanged
+                } else {
+                    # Arch-spec
+                    $Manifest.architecture | Get-Member -MemberType NoteProperty | ForEach-Object {
+                        $arch = $_.Name
+                        $newURL = substitute (arch_specific 'url' $Manifest.autoupdate $arch) $Substitutions
+                        $newHash = HashHelper -AppName $AppName -Version $Version -HashExtraction (arch_specific 'hash' $Manifest.autoupdate $arch) -URL $newURL -Substitutions $Substitutions
+                        $Manifest.architecture.$arch.hash, $hasPropertyChanged = PropertyHelper -Property $Manifest.architecture.$arch.hash -Value $newHash
+                        $hasManifestChanged = $hasManifestChanged -or $hasPropertyChanged
+                    }
+                }
+            } elseif ($Manifest.$currentProperty -and $Manifest.autoupdate.$currentProperty) {
+                # Update other property (global)
+                $newValue = substitute $Manifest.autoupdate.$currentProperty $Substitutions
+                $Manifest.$currentProperty, $hasPropertyChanged = PropertyHelper -Property $Manifest.$currentProperty -Value $newValue
+                $hasManifestChanged = $hasManifestChanged -or $hasPropertyChanged
+            } elseif ($Manifest.architecture) {
+                # Update other property (arch-spec)
+                $Manifest.architecture | Get-Member -MemberType NoteProperty | ForEach-Object {
+                    $arch = $_.Name
+                    if ($Manifest.architecture.$arch.$currentProperty -and ($Manifest.autoupdate.architecture.$arch.$currentProperty -or $Manifest.autoupdate.$currentProperty)) {
+                        $newValue = substitute (arch_specific $currentProperty $Manifest.autoupdate $arch) $Substitutions
+                        $Manifest.architecture.$arch.$currentProperty, $hasPropertyChanged = PropertyHelper -Property $Manifest.architecture.$arch.$currentProperty -Value $newValue
+                        $hasManifestChanged = $hasManifestChanged -or $hasPropertyChanged
+                    }
+                }
             }
         }
+    }
+    end {
+        if ($Version -ne '' -and $Manifest.version -ne $Version) {
+            $Manifest.version = $Version
+            $hasManifestChanged = $true
+        }
+        return $hasManifestChanged
     }
 }
 
@@ -360,132 +407,168 @@ function Get-VersionSubstitution {
     return $versionVariables
 }
 
-function lets_do_autoupdates([String] $app, $json, [String] $version, $substitutions) {
-    $urls = @()
-    $hashes = @()
-    $has_changes = $false
-    $has_errors = $false
-    [Bool]$valid = $true
+function Invoke-AutoUpdate {
+    param (
+        [String]
+        $AppName,
+        [String]
+        $Path,
+        [PSObject]
+        $Manifest,
+        [String]
+        $Version,
+        [Hashtable]
+        $CustomMatches
+    )
 
-    for ($i=0; $i -lt $json.autoupdate.url.Length; $i++) {
-        # create new url
-        $url   = substitute $json.autoupdate.url[$i] $substitutions
-        $urls += $url
-
-        if($valid) {
-            # create hash
-            if ($json.autoupdate.hash.length -gt $i) {
-                $h = $json.autoupdate.hash[$i]
-            } else {
-                $h = $null
-            }
-            $hash = get_hash_for_app $app $h $version $url $substitutions
-            $hashes += $hash
-            if ($null -eq $hash) {
-                $valid = $false
-                Write-Host -f DarkRed "Could not find hash!"
-            }
-        }
-    }
-
-    # write changes to the json object
-    if ($valid) {
-        $has_changes = $true
-        update_manifest_with_new_version $json $version $urls $hashes
-    } else {
-        $has_errors = $true
-        throw "Could not update $app"
-    }
-
-    return $has_changes, $has_errors
-}
-
-
-function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $matches) {
-    Write-Host -f DarkCyan "Autoupdating $app"
-    $has_changes = $false
-    $has_errors = $false
-    [Bool]$valid = $true
-    $substitutions = Get-VersionSubstitution $version $matches
-
-    if ($json.url) {
-        if ($json.autoupdate.url -is [System.Array]) {
-            $has_changes, $has_errors = lets_do_autoupdates $app $json $version $substitutions
-            $has_changes = $return[0]
-            $has_errors = $return[1]
-        } else {
-            # create new url
-            $url   = substitute $json.autoupdate.url $substitutions
-            $valid = $true
-
-            if($valid) {
-                # create hash
-                $hash = get_hash_for_app $app $json.autoupdate.hash $version $url $substitutions
-                if ($null -eq $hash) {
-                    $valid = $false
-                    Write-Host -f DarkRed "Could not find hash!"
-                }
-            }
-
-            # write changes to the json object
-            if ($valid) {
-                $has_changes = $true
-                update_manifest_with_new_version $json $version $url $hash
-            } else {
-                $has_errors = $true
-                throw "Could not update $app"
-            }
-        }
-    } else {
-        $json.architecture | Get-Member -MemberType NoteProperty | ForEach-Object {
-            $valid = $true
-            $architecture = $_.Name
-
-            # create new url
-            $url   = substitute (arch_specific "url" $json.autoupdate $architecture) $substitutions
-            $valid = $true
-
-            if($valid) {
-                # create hash
-                $hash = get_hash_for_app $app (arch_specific "hash" $json.autoupdate $architecture) $version $url $substitutions
-                if ($null -eq $hash) {
-                    $valid = $false
-                    Write-Host -f DarkRed "Could not find hash!"
-                }
-            }
-
-            # write changes to the json object
-            if ($valid) {
-                $has_changes = $true
-                update_manifest_with_new_version $json $version $url $hash $architecture
-            } else {
-                $has_errors = $true
-                throw "Could not update $app $architecture"
-            }
-        }
-    }
+    Write-Host "Autoupdating $AppName" -ForegroundColor DarkCyan
+    $substitutions = Get-VersionSubstitution $Version $CustomMatches
 
     # update properties
-    update_manifest_prop "extract_dir" $json $substitutions
+    $updatedProperties = @(@($Manifest.autoupdate.PSObject.Properties.Name) -ne 'architecture')
+    if ($Manifest.autoupdate.architecture) {
+        $updatedProperties += $Manifest.autoupdate.architecture.PSObject.Properties | ForEach-Object { $_.Value.PSObject.Properties.Name }
+    }
+    if ($updatedProperties -contains 'url') {
+        $updatedProperties += 'hash'
+    }
+    $updatedProperties = $updatedProperties | Select-Object -Unique
+    debug [$updatedProperties]
+    $hasChanged = Update-ManifestProperty -Manifest $Manifest -Property $updatedProperties -AppName $AppName -Version $Version -Substitutions $substitutions
 
-    # update license
-    update_manifest_prop "license" $json $substitutions
-
-    if ($has_changes -and !$has_errors) {
+    if ($hasChanged) {
         # write file
-        Write-Host -f DarkGreen "Writing updated $app manifest"
-
-        $path = join-path $dir "$app.json"
-
-        $file_content = $json | ConvertToPrettyJson
-        [System.IO.File]::WriteAllLines($path, $file_content)
-
+        Write-Host "Writing updated $AppName manifest" -ForegroundColor DarkGreen
+        # Accept unusual Unicode characters
+        # 'Set-Content -Encoding ASCII' don't works in PowerShell 5
+        # Wait for 'UTF8NoBOM' Encoding in PowerShell 7
+        # $Manifest | ConvertToPrettyJson | Set-Content -Path (Join-Path $Path "$AppName.json") -Encoding UTF8NoBOM
+        [System.IO.File]::WriteAllLines((Join-Path $Path "$AppName.json"), (ConvertToPrettyJson $Manifest))
         # notes
-        if ($json.autoupdate.note) {
-            Write-Host ""
-            Write-Host -f DarkYellow $json.autoupdate.note
+        $note = "`nUpdating note:"
+        if ($Manifest.autoupdate.note) {
+            $note += "`nno-arch: $($Manifest.autoupdate.note)"
+            $hasNote = $true
+        }
+        if ($Manifest.autoupdate.architecture) {
+            '64bit', '32bit' | ForEach-Object {
+                if ($Manifest.autoupdate.architecture.$_.note) {
+                    $note += "`n$_-arch: $($Manifest.autoupdate.architecture.$_.note)"
+                    $hasNote = $true
+                }
+            }
+        }
+        if ($hasNote) {
+            Write-Host $note -ForegroundColor DarkYellow
         }
     } else {
-        Write-Host -f DarkGray "No updates for $app"
+        # This if-else branch may not be in use.
+        Write-Host "No updates for $AppName" -ForegroundColor DarkGray
     }
+}
+
+## Helper Functions
+
+function PropertyHelper {
+    <#
+    .SYNOPSIS
+        Helper of updating property
+    .DESCRIPTION
+        Update manifest property (String, Array or PSCustomObject).
+    .PARAMETER Property
+        Property to be updated
+    .PARAMETER Value
+        New property values
+        Update line by line
+    .OUTPUTS
+        System.Object[]
+            The first element is new property, the second element is change flag
+    #>
+    param (
+        [Object]$Property,
+        [Object]$Value
+    )
+    $hasChanged = $false
+    if (@($Property).Length -lt @($Value).Length) {
+        $Property = $Value
+        $hasChanged = $true
+    } else {
+        switch ($Property.GetType().Name) {
+            'String' {
+                $Value = $Value -as [String]
+                if ($null -ne $Value) {
+                    $Property = $Value
+                    $hasChanged = $true
+                }
+            }
+            'Object[]' {
+                $Value = @($Value)
+                for ($i = 0; $i -lt $Value.Length; $i++) {
+                    $Property[$i], $hasItemChanged = PropertyHelper -Property $Property[$i] -Value $Value[$i]
+                    $hasChanged = $hasChanged -or $hasItemChanged
+                }
+            }
+            'PSCustomObject' {
+                if ($Value -is [PSObject]) {
+                    foreach ($name in $Property.PSObject.Properties.Name) {
+                        if ($Value.$name) {
+                            $Property.$name, $hasItemChanged = PropertyHelper -Property $Property.$name -Value $Value.$name
+                            $hasChanged = $hasChanged -or $hasItemChanged
+                        }
+                    }
+                }
+            }
+        }
+    }
+    return $Property, $hasChanged
+}
+
+function HashHelper {
+    <#
+    .SYNOPSIS
+        Helper of getting file hash(es)
+    .DESCRIPTION
+        Extract or calculate file hash(es).
+        If hash extraction templates are less then URLs, the last template will be reused for the rest URLs.
+    .PARAMETER AppName
+        Software name
+    .PARAMETER Version
+        Given software version
+    .PARAMETER HashExtraction
+        Hash extraction template(s)
+    .PARAMETER URL
+        New download URL(s), used to calculate hash locally (fallback)
+    .PARAMETER Substitutions
+        Hashtable of internal substitutable variables
+    .OUTPUTS
+        System.String
+            Hash value (single URL)
+        System.String[]
+            Hash values (multi URLs)
+    #>
+    param (
+        [String]
+        $AppName,
+        [String]
+        $Version,
+        [PSObject[]]
+        $HashExtraction,
+        [String[]]
+        $URL,
+        [HashTable]
+        $Substitutions
+    )
+    $hash = @()
+    for ($i = 0; $i -lt $URL.Length; $i++) {
+        if ($null -eq $HashExtraction) {
+            $currentHashExtraction = $null
+        } else {
+            $currentHashExtraction = $HashExtraction[$i], $HashExtraction[-1] | Select-Object -First 1
+        }
+        $hash += get_hash_for_app $AppName $currentHashExtraction $Version $URL[$i] $Substitutions
+        if ($null -eq $hash[$i]) {
+            throw "Could not update $AppName, hash for $(url_remote_filename $URL[$i]) failed!"
+        }
+    }
+    return $hash
 }
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -284,11 +284,11 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
     return $hash
 }
 
-function update_manifest_with_new_version($json, [String] $version, [String] $url, [String] $hash, $architecture = $null) {
+function update_manifest_with_new_version($json, [String] $version, $url, $hash, $architecture = $null) {
     $json.version = $version
 
     if ($null -eq $architecture) {
-        if ($json.url -is [System.Array]) {
+        if ($json.url -is [System.Array] -and $url -isnot [System.Array]) {
             $json.url[0] = $url
             $json.hash[0] = $hash
         } else {
@@ -297,7 +297,7 @@ function update_manifest_with_new_version($json, [String] $version, [String] $ur
         }
     } else {
         # If there are multiple urls we replace the first one
-        if ($json.architecture.$architecture.url -is [System.Array]) {
+        if ($json.architecture.$architecture.url -is [System.Array] -and $url -isnot [System.Array]) {
             $json.architecture.$architecture.url[0] = $url
             $json.architecture.$architecture.hash[0] = $hash
         } else {
@@ -352,6 +352,47 @@ function get_version_substitutions([String] $version, [Hashtable] $customMatches
     return $versionVariables
 }
 
+function lets_do_autoupdates([String] $app, $json, [String] $version, $substitutions) {
+    $urls = @()
+    $hashes = @()
+    $has_changes = $false
+    $has_errors = $false
+    [Bool]$valid = $true
+
+    for ($i=0; $i -lt $json.autoupdate.url.Length; $i++) {
+        # create new url
+        $url   = substitute $json.autoupdate.url[$i] $substitutions
+        $urls += $url
+
+        if($valid) {
+            # create hash
+            if ($json.autoupdate.hash.length -gt $i) {
+                $h = $json.autoupdate.hash[$i]
+            } else {
+                $h = $null
+            }
+            $hash = get_hash_for_app $app $h $version $url $substitutions
+            $hashes += $hash
+            if ($null -eq $hash) {
+                $valid = $false
+                Write-Host -f DarkRed "Could not find hash!"
+            }
+        }
+    }
+
+    # write changes to the json object
+    if ($valid) {
+        $has_changes = $true
+        update_manifest_with_new_version $json $version $urls $hashes
+    } else {
+        $has_errors = $true
+        throw "Could not update $app"
+    }
+
+    return $has_changes, $has_errors
+}
+
+
 function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $matches) {
     Write-Host -f DarkCyan "Autoupdating $app"
     $has_changes = $false
@@ -360,26 +401,32 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
     $substitutions = get_version_substitutions $version $matches
 
     if ($json.url) {
-        # create new url
-        $url   = substitute $json.autoupdate.url $substitutions
-        $valid = $true
-
-        if($valid) {
-            # create hash
-            $hash = get_hash_for_app $app $json.autoupdate.hash $version $url $substitutions
-            if ($null -eq $hash) {
-                $valid = $false
-                Write-Host -f DarkRed "Could not find hash!"
-            }
-        }
-
-        # write changes to the json object
-        if ($valid) {
-            $has_changes = $true
-            update_manifest_with_new_version $json $version $url $hash
+        if ($json.autoupdate.url -is [System.Array]) {
+            $has_changes, $has_errors = lets_do_autoupdates $app $json $version $substitutions
+            $has_changes = $return[0]
+            $has_errors = $return[1]
         } else {
-            $has_errors = $true
-            throw "Could not update $app"
+            # create new url
+            $url   = substitute $json.autoupdate.url $substitutions
+            $valid = $true
+
+            if($valid) {
+                # create hash
+                $hash = get_hash_for_app $app $json.autoupdate.hash $version $url $substitutions
+                if ($null -eq $hash) {
+                    $valid = $false
+                    Write-Host -f DarkRed "Could not find hash!"
+                }
+            }
+
+            # write changes to the json object
+            if ($valid) {
+                $has_changes = $true
+                update_manifest_with_new_version $json $version $url $hash
+            } else {
+                $has_errors = $true
+                throw "Could not update $app"
+            }
         }
     } else {
         $json.architecture | Get-Member -MemberType NoteProperty | ForEach-Object {
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -324,26 +324,34 @@ function update_manifest_prop([String] $prop, $json, [Hashtable] $substitutions)
     }
 }
 
-function get_version_substitutions([String] $version, [Hashtable] $customMatches) {
-    $firstPart = $version.Split('-') | Select-Object -first 1
-    $lastPart = $version.Split('-') | Select-Object -last 1
+function Get-VersionSubstitution {
+    param (
+        [String]
+        $Version,
+        [Hashtable]
+        $CustomMatches
+    )
+
+    $firstPart = $Version.Split('-') | Select-Object -First 1
+    $lastPart = $Version.Split('-') | Select-Object -Last 1
     $versionVariables = @{
-        '$version' = $version;
-        '$underscoreVersion' = ($version -replace "\.", "_");
-        '$dashVersion' = ($version -replace "\.", "-");
-        '$cleanVersion' = ($version -replace "\.", "");
-        '$majorVersion' = $firstPart.Split('.') | Select-Object -first 1;
-        '$minorVersion' = $firstPart.Split('.') | Select-Object -skip 1 -first 1;
-        '$patchVersion' = $firstPart.Split('.') | Select-Object -skip 2 -first 1;
-        '$buildVersion' = $firstPart.Split('.') | Select-Object -skip 3 -first 1;
+        '$version' = $Version;
+        '$dotVersion' = ($Version -replace '[._-]', '.');
+        '$underscoreVersion' = ($Version -replace '[._-]', '_');
+        '$dashVersion' = ($Version -replace '[._-]', '-');
+        '$cleanVersion' = ($Version -replace '[._-]', '');
+        '$majorVersion' = $firstPart.Split('.') | Select-Object -First 1;
+        '$minorVersion' = $firstPart.Split('.') | Select-Object -Skip 1 -First 1;
+        '$patchVersion' = $firstPart.Split('.') | Select-Object -Skip 2 -First 1;
+        '$buildVersion' = $firstPart.Split('.') | Select-Object -Skip 3 -First 1;
         '$preReleaseVersion' = $lastPart;
     }
-    if($version -match "(?<head>\d+\.\d+(?:\.\d+)?)(?<tail>.*)") {
-        $versionVariables.Set_Item('$matchHead', $matches['head'])
-        $versionVariables.Set_Item('$matchTail', $matches['tail'])
+    if($Version -match "(?<head>\d+\.\d+(?:\.\d+)?)(?<tail>.*)") {
+        $versionVariables.Set_Item('$matchHead', $Matches['head'])
+        $versionVariables.Set_Item('$matchTail', $Matches['tail'])
     }
-    if($customMatches) {
-        $customMatches.GetEnumerator() | ForEach-Object {
+    if($CustomMatches) {
+        $CustomMatches.GetEnumerator() | ForEach-Object {
             if($_.Name -ne "0") {
                 $versionVariables.Set_Item('$match' + (Get-Culture).TextInfo.ToTitleCase($_.Name), $_.Value)
             }
@@ -398,7 +406,7 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
     $has_changes = $false
     $has_errors = $false
     [Bool]$valid = $true
-    $substitutions = get_version_substitutions $version $matches
+    $substitutions = Get-VersionSubstitution $version $matches
 
     if ($json.url) {
         if ($json.autoupdate.url -is [System.Array]) {
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -839,18 +839,27 @@ function is_scoop_outdated() {
 }
 
 function substitute($entity, [Hashtable] $params, [Bool]$regexEscape = $false) {
-    if ($entity -is [Array]) {
-        return $entity | ForEach-Object { substitute $_ $params $regexEscape}
-    } elseif ($entity -is [String]) {
-        $params.GetEnumerator() | ForEach-Object {
-            if($regexEscape -eq $false -or $null -eq $_.Value) {
-                $entity = $entity.Replace($_.Name, $_.Value)
-            } else {
-                $entity = $entity.Replace($_.Name, [Regex]::Escape($_.Value))
+    $newentity = $entity
+    if ($null -ne $newentity) {
+        switch ($entity.GetType().Name) {
+            'String' {
+                $params.GetEnumerator() | ForEach-Object {
+                    if ($regexEscape -eq $false -or $null -eq $_.Value) {
+                        $newentity = $newentity.Replace($_.Name, $_.Value)
+                    } else {
+                        $newentity = $newentity.Replace($_.Name, [Regex]::Escape($_.Value))
+                    }
+                }
+            }
+            'Object[]' {
+                $newentity = $entity | ForEach-Object { substitute $_ $params $regexEscape }
+            }
+            'PSCustomObject' {
+                $newentity.PSObject.Properties | ForEach-Object { $_.Value = substitute $_.Value $params $regexEscape }
             }
         }
-        return $entity
     }
+    return $newentity
 }
 
 function format_hash([String] $hash) {
--- a/schema.json
+++ b/schema.json
@@ -170,58 +170,88 @@
             "minItems": 1,
             "type": "array"
         },
-        "autoupdate": {
+        "autoupdateArch": {
             "additionalProperties": false,
             "properties": {
-                "architecture": {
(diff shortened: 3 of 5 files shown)


rasa:
Works with the above manifest:
```
C:\scoop\buckets\extras\bucket >POWERSHELL -file ..\bin\checkver.ps1 irfanview -f
irfanview: 4.53 (scoop version is 4.52) autoupdate available
Forcing autoupdate!
Autoupdating irfanview
Searching hash for iview453.zip in https://www.irfanview.com/main_download_engl.htm
Found: 89c0d29424299f9b0d36248884059baf815f0b5d4adc9c3b426f61fe68f1444a using Extract Mode
Searching hash for iview453_plugins.zip in https://www.irfanview.com/plugins.htm
Found: 1ec241d26a0f061aa5499c606967690de8979c82fdce4a5724b4e18daa099de3 using Extract Mode
Searching hash for iview453_x64.zip in https://www.irfanview.com/64bit.htm
Found: 941056da4f50df44eeef46ca875255bd3d09c3aaf1bdfffad5f65de0a515ff02 using Extract Mode
Searching hash for iview453_plugins_x64.zip in https://www.irfanview.com/64bit.htm
Found: d5e8ae1734264287e7b442fc062899c206c54bbd423304e125c4b1d1c58d28b1 using Extract Mode
Writing updated irfanview manifest
```
It also works with a hacked [kitty.json](https://gist.github.com/rasa/a79ff4aa48031148a7c387dcf0fb0c79):
```
kitty: 0.71.0.4 (scoop version is 0.71.0.2) autoupdate available
Forcing autoupdate!
Autoupdating kitty
Downloading genpass.exe to compute hashes!
genpass.exe (42.9 KB) [==========================================================================================================================================================================================================================================================================================] 100%
Computed hash: 13673f67164c1d0431ffdb8da93c97e64e18f0458eb7d30ebd52ee18f72ef94e
Downloading kageant.exe to compute hashes!
kageant.exe (364.0 KB) [=========================================================================================================================================================================================================================================================================================] 100%
Computed hash: db058d9c25c9b36d9a1f66b7b76d3f7c591e468dd289812835c74413c0562da3
Downloading kitty.exe to compute hashes!
kitty.exe (648.5 KB) [===========================================================================================================================================================================================================================================================================================] 100%
Computed hash: 3b4e8da7b879b2cf33c4e43b795dc4aa2a9df8b17e366a58276a0916b1d41800
Downloading klink.exe to compute hashes!
klink.exe (281.0 KB) [===========================================================================================================================================================================================================================================================================================] 100%
Computed hash: f1470d2143e66325b355eedc61a0a8d0b92238c3c6d99e570f21c63a856af3ff
Downloading kscp.exe to compute hashes!
kscp.exe (780.5 KB) [============================================================================================================================================================================================================================================================================================] 100%
Computed hash: d0c0911c75fa37311a5ee600e511026798825e3174f5f06875f34cd81b6c8d52
Downloading ksftp.exe to compute hashes!
ksftp.exe (791.0 KB) [===========================================================================================================================================================================================================================================================================================] 100%
Computed hash: a5c635efc825ab797b0464b821f0a2228a1f77d1c7505ed2723a9e5c2300db51
Writing updated kitty manifest

**Comment by niheaven** [comment]
- Add [PSCustomObject] support to `core.ps1/substitute`
- Add [PSCustomObject] support to `Update-ManifestProperty`
- Find to-be-updated properties automatically
- Add 'almost' every properties in schema's autoupdate section

File: schema.json [hunk 1]
```diff
@@ -158,50 +173,164 @@
...
                                 "url": {
-                                    "format": "uri",
-                                    "type": "string"
+                                    "$ref": "#/definitions/autoupdateUriOrArrayOfAutoupdateUris"
                                 },
                                 "hash": {
-                                    "$ref": "#/definitions/hashExtraction"
+                                    "$ref": "#/definitions/hashExtractionOrArrayOfHashExtractions"
+                                },
+                                "extract_dir": {
```
r15ch13: All those properties could be moved to a single definition, I guess. 🤔

lib/autoupdate.ps1
```diff
@@ -281,44 +281,141 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
...
+        [String]
+        $AppName,
+        [String]
+        $Version,
+        [Alias("Matches")]
+        [HashTable]
+        $Substitutions
+    )
+    begin {
+        function PropertyHelper {
```
r15ch13: Is it useful to redeclare these every time when `Update-ManifestProperty` is called?

schema.json
niheaven: Move properties under architecture to single definition, but those under autoupdate, I've no idea...

lib/autoupdate.ps1
niheaven: Wow, if used manually, there is no difference, but for batch operation, moving them outside should be better.

lib/autoupdate.ps1
```diff
@@ -351,69 +379,20 @@ function get_version_substitutions([String] $version, [Hashtable] $customMatches
...
-            }
-        }
-    }
-
     # update properties
-    update_manifest_prop "extract_dir" $json $substitutions
-
-    # update license
-    update_manifest_prop "license" $json $substitutions
+    $properties_updated = @(coalesce ($json.autoupdate.PSObject.Properties.Name -ne 'architecture') @())
```
r15ch13: Doesn't work when there is only one autoupdate property.

Kitty example:
```json
    "autoupdate": {
        "url": [
            "https://github.com/cyd01/KiTTY/releases/download/v$version/genpass.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/kageant.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/kitty.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/klink.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/kscp.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/ksftp.exe"
        ]
    }
```

image

r15ch13:
`autoupdateArch` and `autoupdate` could be reduced to:

Removed `persist` and `psmodule` from `autoupdateArch` because they make no sense here.
`env_add_path` and `env_set` can also be removed from `architecture`, I guess.

```json
		"autoupdateArch": {
            "type": "object",
            "properties": {
                "url": {
                    "$ref": "#/definitions/autoupdateUriOrArrayOfAutoupdateUris"
                },
                "hash": {
                    "$ref": "#/definitions/hashExtractionOrArrayOfHashExtractions"
                },
                "extract_dir": {
                    "$ref": "#/definitions/stringOrArrayOfStrings"
                },
                "extract_to": {
                    "$ref": "#/definitions/stringOrArrayOfStrings"
                },
                "env_add_path": {
                    "$ref": "#/definitions/stringOrArrayOfStrings"
                },
                "env_set": {
                    "type": "object"
                },
                "bin": {
                    "$ref": "#/definitions/stringOrArrayOfStringsOrAnArrayOfArrayOfStrings"
                },
                "shortcuts": {
                    "$ref": "#/definitions/shortcutsArray"
                },
                "installer": {
                    "$ref": "#/definitions/installer"
                },
                "post_install": {
                    "$ref": "#/definitions/stringOrArrayOfStrings"
                },
                "pre_install": {
                    "$ref": "#/definitions/stringOrArrayOfStrings"
                },
                "uninstaller": {
                    "$ref": "#/definitions/uninstaller"
                }
            }
        },
        "autoupdate": {
            "type": "object",
            "allOf": [
                {
                    "$ref": "#/definitions/autoupdateArch"
                },
                {
                    "properties": {
                        "notes": {
                            "$ref": "#/definitions/stringOrArrayOfStrings"
                        },
                        "architecture": {
                            "type": "object",
                            "additionalProperties": false,
                            "properties": {
                                "32bit": {
                                    "$ref": "#/definitions/autoupdateArch"
                                },
                                "64bit": {
                                    "$ref": "#/definitions/autoupdateArch"
                                }
                            }
                        }
                    }
                }
            ]
        },
```

niheaven:
I've fix it, but I found something else may be broken by some commits... I need to do a more deeper review...

Ash258:
Are you sure you want to have installer scripts in autoupdate? It will introduce more messy manifests. I really do not see valid use case for them.

lib/autoupdate.ps1
niheaven: Force `$json.autoupdate.PSObject.Properties.Name` be `[Array]` solve the prob, and it could be simplified. Now everything works.

rasa:
I updated  [kitty.json](https://gist.github.com/rasa/a79ff4aa48031148a7c387dcf0fb0c79) and updated it, and installed without issue:
```batch
>powershell -file checkver.ps1 -Dir . kitty -u
kitty: 0.71.0.6 (scoop version is 0.71.0.5) autoupdate available
Autoupdating kitty
Downloading genpass.exe to compute hashes!
genpass.exe (42.9 KB) [===========================================================================================================================================================================================================================] 100%
Computed hash: 13673f67164c1d0431ffdb8da93c97e64e18f0458eb7d30ebd52ee18f72ef94e
Downloading kittygen.exe to compute hashes!
kittygen.exe (454.5 KB) [=========================================================================================================================================================================================================================] 100%
Computed hash: 4e3ccf07081cf3118669fdce720b6a2766b784dec23c3e8ee902d898d6b1c951
Downloading kageant.exe to compute hashes!
kageant.exe (364.0 KB) [==========================================================================================================================================================================================================================] 100%
Computed hash: 83019314f8f925990ac2f7ff041683acdc29a23bf370c0dd40e40b2ea6514f8b
Downloading kitty.exe to compute hashes!
kitty.exe (649.0 KB) [============================================================================================================================================================================================================================] 100%
Computed hash: 770fb13def6fab82fcaacede07181c9307bd790978a3286c256d6b1c8f80ccc7
Downloading klink.exe to compute hashes!
klink.exe (281.0 KB) [============================================================================================================================================================================================================================] 100%
Computed hash: 4c2260ea2347346ada6c120f9371f1870610dda30feabe56dd92378f8f02a0c4
Downloading kscp.exe to compute hashes!
kscp.exe (781.0 KB) [=============================================================================================================================================================================================================================] 100%
Computed hash: 49a6de18c88f9864a97bc08b4072531c693c73beda1f7f9243b10fc845c1e8a1
Downloading ksftp.exe to compute hashes!
ksftp.exe (791.5 KB) [============================================================================================================================================================================================================================] 100%
Computed hash: b0e0602d308a135a7cd626172f05ce5bf8b1f9e05b792317f4bcb05ae3af3afe
Writing updated kitty manifest

>call scoop.cmd install kitty.json
Installing 'kitty' (0.71.0.6) [64bit]
Loading genpass.exe from cache
Checking hash of genpass.exe ... ok.
Loading kittygen.exe from cache
Checking hash of kittygen.exe ... ok.
Loading kageant.exe from cache
Checking hash of kageant.exe ... ok.
Loading kitty.exe from cache
Checking hash of kitty.exe ... ok.
Loading klink.exe from cache
Checking hash of klink.exe ... ok.
Loading kscp.exe from cache
Checking hash of kscp.exe ... ok.
Loading ksftp.exe from cache
Checking hash of ksftp.exe ... ok.
Running pre-install script...
Linking C:\scoop\apps\kitty\current => C:\scoop\apps\kitty\0.71.0.6
Creating shim for 'genpass'.
Creating shim for 'kittygen'.
Creating shim for 'kageant'.
Creating shim for 'kitty'.
Creating shim for 'klink'.
Creating shim for 'kscp'.
Creating shim for 'ksftp'.
Creating shortcut for KiTTY (kitty.exe)
Persisting Sessions
Persisting kitty.ini
'kitty' (0.71.0.6) was installed successfully!

**Comment by niheaven** [comment]
@r15ch13 @Ash258 Current `schema.json` just provides probability for `autoupdate` items, i.e., manifest *COULD* have these properties autoupdated. For some properties we may *SHOULDN'T* or *NEEDN'T* do so, but we *COULD* do.

`persist` should be autoupdated with renaming tech, e.g., `lua 5.3` use `config-5.3.lua` as config, while `lua 5.2` use `config-5.2.lua`, we could persist it with renaming persist such as

```json
"persist": [
    [
        "config-5.3.lua",
        "config.lua"
    ]
],
"autoupdate": {
    "persist": [
        [
            "config-$majorVersion.$minorVersion.lua",
            "config.lua"
        ]
    ]
}
```

And you could use all [Internal Substitutable Variables](https://github.com/lukesampson/scoop/wiki/App-Manifest-Autoupdate#internal-substitutable-variables) in `installer`, not only `$version`. I haven't found valid use case now, but maybe it's useful.

linsui:
Maybe this is a use case? https://github.com/lukesampson/scoop-extras/pull/2480#discussion_r305199565

Ash258:
Noone talked about persist. I talked about scripts.

I do not agree. This will result really junk manifest when you have tons of duplicated code for just one line of variable substitution. `Installer.script`, `post_intall` and `pre_install` should not be in autoupdate.

@linsui No it is not. Completely easy use case for gci. There is no need to use substitution of internal variables. Because then you need to have all that installer script in autoupdate. Which is completely nonsense.

Ash258:
Look at this mess https://github.com/lukesampson/scoop/pull/2870/commits/a3ca2c3d3d81852bc92c9e7d5f348442b20b14b6. 

Completely non intuitive, ugly, non extendable manifest. Because user cannot write powershell scripts.

niheaven:
You could just substitute only some line, e.g. first one or two lines. `autoupdate` will do line by line replacement, that is, if you have two lines of `installer.script` in `au`, then only first two lines of `installer.script` will be replaced.

We also delete `installer`, `pre_install`, `post_install` and `pre_uninstall` now, and add them again when really needed. For other properties, there is some use cases now, I think.

Ash258:
Remove all scripts from autoupdate. It is really not good idea to have them.

For installer there could be only `file` inside autoupdate.

niheaven:
`file` is from `$fname`, so always up to date. Should we leave, e.g., `post_install` here, just in case?

Ash258:
You are missing use case when `installer.file` needs to be some file from zip archive.

```yml
url: https://path.com/to/archive.zip # This will be extracted
installer:
  file: installerWithVersion1.3.5.exe # $fname is archive.zip, not installer....exe
```

niheaven:
So let `autoupdateArch` likes

```json
        "autoupdateArch": {
            "additionalProperties": false,
            "properties": {
                "url": {
                    "$ref": "#/definitions/autoupdateUriOrArrayOfAutoupdateUris"
                },
                "hash": {
                    "$ref": "#/definitions/hashExtractionOrArrayOfHashExtractions"
                },
                "extract_dir": {
                    "$ref": "#/definitions/stringOrArrayOfStrings"
                },
                "extract_to": {
                    "$ref": "#/definitions/stringOrArrayOfStrings"
                },
                "env_add_path": {
                    "$ref": "#/definitions/stringOrArrayOfStrings"
                },
                "env_set": {
                    "type": "object"
                },
                "bin": {
                    "$ref": "#/definitions/stringOrArrayOfStringsOrAnArrayOfArrayOfStrings"
                },
                "shortcuts": {
                    "$ref": "#/definitions/shortcutsArray"
                },
                "installer": {
                    "additionalProperties": false,
                    "properties": {
                        "file": {
                            "type": "string"
                        }
                    },
                    "type": "object"
                },
                "post_install": {
                    "$ref": "#/definitions/stringOrArrayOfStrings"
                },
                "psmodule": {
                    "additionalProperties": false,
                    "properties": {
                        "name": {
                            "type": "string"
                        }
                    },
                    "type": "object"
                },
                "persist": {
                    "$ref": "#/definitions/stringOrArrayOfStringsOrAnArrayOfArrayOfStrings"
                }
            },
            "type": "object"
        },
```

Is that okay?

Ash258:
I would still remove `post_install` but if it is suitable for more collaborators, then i am fine with keeping it.

niheaven:
Done, please test it again @r15ch13 @rasa @Ash258

lib/core.ps1
```diff
@@ -823,17 +823,21 @@ function is_scoop_outdated() {
...
             } else {
                 $entity = $entity.Replace($_.Name, [Regex]::Escape($_.Value))
             }
         }
         return $entity
+    } elseif ($entity -is [Array]) {
+        return $entity | ForEach-Object { substitute $_ $params $regexEscape }
+    } elseif ($entity -is [PSObject]) {
+        $newentity = $entity.PSObject.Copy()
+        $newentity.PSObject.Properties.ForEach( { $_.Value = substitute $_.Value $params $regexEscape })
```
Ash258: Why `.ForEach` instead of `| ForEach-Object`

A

B

Ash258: Why `.ForEach` instead of `| ForEach-Object`

A

B

lib/autoupdate.ps1
```diff
@@ -431,3 +410,71 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+    $Hash = @()
+    for ($i = 0; $i -lt $URL.Length; $i++) {
+        if ($null -eq $HashExtraction) {
+            $aHashExtraction = $null
+        } else {
+            $aHashExtraction = $HashExtraction[$i], $HashExtraction[-1] | Select-Object -First 1
+        }
+        $Hash += get_hash_for_app $AppName $aHashExtraction $Version $URL[$i] $Substitutions
+        if ($null -eq $Hash[$i]) {
+            abort "Could not update $AppName, hash for $URL[$i] failed!"
```
Ash258: I do no think that powershell is expanding array access in strings

```suggestion
            abort "Could not update $AppName, hash for $($URL[$i]) failed!"
```

Ash258:
Please do not use `aWhatever` variables. It is completely unreadable in code, as it looks the same as parent collection.

lib/autoupdate.ps1
```diff
@@ -431,3 +410,71 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+        }
+    }
+    if ($Hash.Length -eq 1) {
```
Ash258: Why do you return string and array? Is there specific reason to return simple string hash?

linsui:
Could checkver be autoupdated? This may fix https://github.com/lukesampson/scoop-extras/issues/2340 .

lib/core.ps1
niheaven: I've misunderstood `ForEach-Object`'s usage. Thanks.

lib/autoupdate.ps1
niheaven: Forgotten... Delete and no error.

lib/autoupdate.ps1
```diff
@@ -431,3 +424,101 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+        if (($null -eq $Value) -or ($value -eq $Property)) {
+            return $Property, $false
+        } else {
+            return $Value, $true
+        }
+    } elseif ($Property -is [Array]) {
+        $Value = @($Value)
+        for ($i = 0; $i -lt [Math]::Min($Property.Length, $Value.Length); $i++) {
+            $Property[$i], $hasItemChanged = PropertyHelper -Property $Property[$i] -Value $Value[$i]
+            $hasChanged = $hasChanged -or $hasItemChanged
```
Ash258: Should not there be some `break`?

lib/autoupdate.ps1
```diff
@@ -431,3 +424,101 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+    .OUTPUTS
+        System.Object[]
+            The first element is new property, the second element is change flag
+    #>
+    param (
+        [Object]$Property,
+        [Object]$Value
+    )
+    $hasChanged = $false
+    if ($Property -is [String]) {
```
Ash258: Would switch look better?

lib/autoupdate.ps1
```diff
@@ -431,3 +424,101 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+    } elseif ($Property -is [Array]) {
+        $Value = @($Value)
+        for ($i = 0; $i -lt [Math]::Min($Property.Length, $Value.Length); $i++) {
+            $Property[$i], $hasItemChanged = PropertyHelper -Property $Property[$i] -Value $Value[$i]
+            $hasChanged = $hasChanged -or $hasItemChanged
+        }
+        return $Property, $hasChanged
+    } elseif ($Property -is [PSObject]) {
+        if ($Value -is [PSObject]) {
+            foreach ($Name in $Property.PSObject.Properties.Name) {
```
Ash258:
```suggestion
            foreach ($name in $Property.PSObject.Properties.Name) {
```

lib/autoupdate.ps1
Ash258: Because then you could potentionally skip some properties. Right?

You are returning only single property and boolean in #L463.

lib/autoupdate.ps1
```diff
@@ -281,43 +281,90 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
...
+        Hashtable of internal substitutable variables
+    .OUTPUTS
+        System.Boolean
+            Flag that indicate if there are any changed properties
+    #>
+    [CmdletBinding(SupportsShouldProcess = $true)]
+    [OutputType([Boolean])]
+    param (
+        [Parameter(Mandatory = $true, Position = 1)]
+        [PSObject]
```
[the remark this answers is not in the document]
niheaven: To be extact, this should be [PSCustomObject]. PSCustomObject < PSObject < Object, and the first two are both okay. I'll change it to [PSCustomObject]， since no type is not precise enough, I thought.

lib/autoupdate.ps1
```diff
@@ -281,43 +281,90 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
...
+        $Manifest,
+        [Parameter(ValueFromPipeline = $true, Position = 2)]
+        [String[]]
+        $Property,
+        [String]
+        $AppName,
+        [String]
+        $Version,
+        [Alias("Matches")]
+        [HashTable]
```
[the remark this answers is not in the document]
niheaven: Yes, this is limited to [HashTable], see `autoupdate` and `get_version_substitutions` in `lib/autoupdate.ps1`

lib/autoupdate.ps1
niheaven: Changed.

lib/autoupdate.ps1
```diff
@@ -431,3 +424,101 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+        [Object]$Property,
+        [Object]$Value
+    )
+    $hasChanged = $false
+    if ($Property -is [String]) {
+        $Value = $Value -as [String]
+        if (($null -eq $Value) -or ($value -eq $Property)) {
+            return $Property, $false
+        } else {
+            return $Value, $true
```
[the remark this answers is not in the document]
niheaven: Done.

lib/autoupdate.ps1
niheaven: Done.

lib/autoupdate.ps1
niheaven: This is for one property, and `$Property[$i]` is one item (or line) in this property. This `for` loop replace first `$Value.Length` lines or items of the property, or all lines or items if `$Property.Length` is less then or equal to `$Value.Length`.

Ash258: Then you could easily break as soon as one item in property was changed.  To skip remaining.

lib/core.ps1
```diff
@@ -839,17 +839,21 @@ function is_scoop_outdated() {
...
+    if ($entity -is [String]) {
         $params.GetEnumerator() | ForEach-Object {
-            if($regexEscape -eq $false -or $null -eq $_.Value) {
+            if ($regexEscape -eq $false -or $null -eq $_.Value) {
                 $entity = $entity.Replace($_.Name, $_.Value)
             } else {
                 $entity = $entity.Replace($_.Name, [Regex]::Escape($_.Value))
             }
         }
         return $entity
```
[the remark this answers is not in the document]
niheaven: Done.

lib/autoupdate.ps1
```diff
@@ -431,3 +424,101 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+    $hash = @()
+    for ($i = 0; $i -lt $URL.Length; $i++) {
+        if ($null -eq $HashExtraction) {
+            $currentHashExtraction = $null
+        } else {
+            $currentHashExtraction = $HashExtraction[$i], $HashExtraction[-1] | Select-Object -First 1
+        }
+        $hash += get_hash_for_app $AppName $currentHashExtraction $Version $URL[$i] $Substitutions
+        if ($null -eq $hash[$i]) {
+            abort "Could not update $AppName, hash for $($URL[$i]) failed!"
```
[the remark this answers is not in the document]
niheaven: Change to `throw`, just as current `master`'s

lib/autoupdate.ps1
niheaven: But we must go through all items to ensure they are updated, at least `$Value.Length`, at most `$Property.Length`. We cannot skip this check and replace.

lib/autoupdate.ps1
```diff
@@ -431,3 +423,107 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+    .OUTPUTS
+        System.Object[]
+            The first element is new property, the second element is change flag
+    #>
+    param (
+        [Object]$Property,
+        [Object]$Value
+    )
+    $hasChanged = $false
+    switch ($Property) {
```
Ash258:
```suggestion
    switch ($Property.GetType().Name) {
        'String' {
             .....
        }
```

lib/core.ps1
```diff
@@ -839,18 +839,28 @@ function is_scoop_outdated() {
...
 function substitute($entity, [Hashtable] $params, [Bool]$regexEscape = $false) {
-    if ($entity -is [Array]) {
-        return $entity | ForEach-Object { substitute $_ $params $regexEscape}
-    } elseif ($entity -is [String]) {
-        $params.GetEnumerator() | ForEach-Object {
-            if($regexEscape -eq $false -or $null -eq $_.Value) {
-                $entity = $entity.Replace($_.Name, $_.Value)
-            } else {
-                $entity = $entity.Replace($_.Name, [Regex]::Escape($_.Value))
+    switch ($entity) {
```
Ash258: $entity.GetType().Name

lib/autoupdate.ps1
Ash258: Ohh... good. Sorry I missed that recursive call.

lib/autoupdate.ps1
niheaven: Done.

lib/core.ps1
niheaven: Done.

lib/core.ps1
```diff
@@ -839,18 +839,28 @@ function is_scoop_outdated() {
...
+    switch ($entity.GetType().Name) {
+        'String' {
+            $params.GetEnumerator() | ForEach-Object {
+                if ($regexEscape -eq $false -or $null -eq $_.Value) {
+                    $entity = $entity.Replace($_.Name, $_.Value)
+                } else {
+                    $entity = $entity.Replace($_.Name, [Regex]::Escape($_.Value))
+                }
             }
+            break
```
Ash258: Are breaks needed? As soon as variable is set switch will be ended naturally (in all branches)

lib/core.ps1
```diff
@@ -839,18 +839,28 @@ function is_scoop_outdated() {
...
+        }
+        'Object[]' {
+            $entity = $entity | ForEach-Object { substitute $_ $params $regexEscape }
+            break
+        }
+        'PSCustomObject' {
+            $entity = $entity.PSObject.Copy()
```
Ash258: Would not make sense to create local variable instead of "overriding" parameter?

lib/core.ps1
Ash258: Break make sense only when you need to conditionally break from whole switch execution.

niheaven: No... `switch` will match all posible variables and run from top to bottom... And I've bugfixed a long time when I first used `switch` in PowerShell... So best solution is `break`.

lib/core.ps1
niheaven: That's okay, too. I'll change it.

niheaven: Maybe this function should be rewrited, but in another PR.

lib/core.ps1
Ash258: Yeah it is. But there should not be multiple matches right? 

As if you would have something like:

```
switch -wildcard ('testing') {
     'test*' { Write-Host 'A' }
      '*ing' { Write-Host 'B' }
}
```
This would print `A` and `B` as it will fit both, but in this case there should not be any entity, which is String  and Object[] (or PSCustomObject)

niheaven: Latest `break` could be omitted, but for better readability, I keep it.

Ash258: Ok. Keep it. I was just curious if there is some potentional use case for it.

niheaven: Got it, `xxx.GetType().Name` is just string, so multi match is imposible. I met error when I match between Type, and there are some inheritance.

lib/core.ps1
```diff
@@ -839,18 +839,26 @@ function is_scoop_outdated() {
...
+                    $newentity = $entity.Replace($_.Name, [Regex]::Escape($_.Value))
+                }
             }
         }
-        return $entity
+        'Object[]' {
+            $newentity = $entity | ForEach-Object { substitute $_ $params $regexEscape }
+        }
+        'PSCustomObject' {
+            $newentity = $entity.PSObject.Copy()
```
Ash258: Is this copying needed?

`$newentity` consist of `$entity` already from L842

lib/autoupdate.ps1
```diff
@@ -431,3 +424,104 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+    }
+    return $Property, $hasChanged
+}
+
+function HashHelper {
+    <#
+    .SYNOPSIS
+        Helper of getting file hash(es)
+    .DESCRIPTION
+        Get file hash(es) by hash extraction template(s).
```
Ash258: By `hash extraction templates` you mean hash property in autoupdate right?

lib/autoupdate.ps1
```diff
@@ -431,3 +424,104 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+        If hash extraction templates are less then URLs, the last template will be reused for the rest URLs.
+    .PARAMETER AppName
+        Software name
+    .PARAMETER Version
+        Given software version
+    .PARAMETER HashExtraction
+        Hash extraction template(s)
+    .PARAMETER URL
+        New download URL(s), used to calculate hash locally
```
Ash258: This is not accurate. It will be used for calculating hash locally or extracting hash.

> used to calculate hash locally

This implies it will be downloaded and calculated all the time, without possible hash extraction when defined and working.

lib/autoupdate.ps1
niheaven: Yes, what's your advice for this property? Or just `autoupdate.hash`?

lib/autoupdate.ps1
Ash258: Yes please. Sounds good.

lib/autoupdate.ps1
Ash258: What about:

```suggestion
        Get file hash(es) by hash extraction method.
```

```suggestion
        Extract or calculate file hash(es)
```

lib/autoupdate.ps1
```diff
@@ -431,3 +424,104 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+            The first element is new property, the second element is change flag
+    #>
+    param (
+        [Object]$Property,
+        [Object]$Value
+    )
+    $hasChanged = $false
+    switch ($Property.GetType().Name) {
+        'String' {
+            $Value = $Value -as [String]
```
Ash258: Is this cast needed?

lib/autoupdate.ps1
```diff
@@ -431,3 +424,104 @@ function autoupdate([String] $app, $dir, $json, [String] $version, [Hashtable] $
...
+            if (($null -ne $Value) -and ($value -ne $Property)) {
```
Ash258: What is use case of `$Value -ne $Property`?

The function will be called for example as `Property-Helper $property.Name $property.Value`, where name will be evaluated from manifest as `$manifest.autoupdate.extract_dir`. So name will be `extract_dir` and value could be `mocked v$version`

I do not see a use case when the actual value could be equal to property name itself

niheaven: `$Property` is the property value, or property value's pointer, so we could replace it with newer one (`$Value`).

lib/autoupdate.ps1
niheaven: This is for error handling, i.e., if `$Property` is `[String]` but `$Value` is `[Object]` in mistake, then `$Value` will be `$null` and just do nothing.

niheaven:
This is ready for merging into `develop`, or maybe `master`? @r15ch13 

Other fixes should be in another PR now.

Ash258:
Not working for kitty

A

Even with single URL

B

niheaven:
I'll test it again, may be due to some recent tweaks.

niheaven:
Located and fixed. `core/substite` has an error when we send `$null` to it, since `$null` cannot `GetType()`, and if `autoupdate` has no `hash`, we'll send `$null`. It was introduced by 73e67af, and now is fixed by null check and bypass.

I used to run the PR on `irfanview` and it has `au.hash`, so didn't find this before :cry: @Ash258 Please test `kitty` again, it works well now.

lib/autoupdate.ps1
```diff
@@ -351,77 +398,23 @@ function get_version_substitutions([String] $version, [Hashtable] $customMatches
...
     # update properties
-    update_manifest_prop "extract_dir" $json $substitutions
-
-    # update license
-    update_manifest_prop "license" $json $substitutions
+    $updatedProperties = @(@($json.autoupdate.PSObject.Properties.Name) -ne 'architecture')
+    if ($json.autoupdate.architecture) {
+        $updatedProperties += $json.autoupdate.architecture.PSObject.Properties | ForEach-Object { $_.Value.PSObject.Properties.Name } | Select-Object -Unique
+    }
+    if (!($updatedProperties -eq 'hash')) {
```
Ash258: This should be array all the time. (L404)

```suggestion
    if ($updatedProperties -notcontains 'hash') {
```

Ash258:
Force parameter is not respected correctly.

A

1. Create manifest bellow and update it.
1. Remove some hashes from manifest. 
1. Force update and recalculate deleted hashes.
1. Manifest is not updated with remaining hashes

```json
{
    "version": "0.72.0.6",
    "homepage": "http://kitty.9bis.net",
    "description": "A telnet / SSH client, which is a fork from version 0.72 of PuTTY.",
    "license": "MIT",
    "url": [
        "https://github.com/cyd01/KiTTY/releases/download/v0.72.0.6/genpass.exe",
        "https://github.com/cyd01/KiTTY/releases/download/v0.72.0.6/kittygen.exe",
        "https://github.com/cyd01/KiTTY/releases/download/v0.72.0.6/kageant.exe",
        "https://github.com/cyd01/KiTTY/releases/download/v0.72.0.6/kitty.exe",
        "https://github.com/cyd01/KiTTY/releases/download/v0.72.0.6/klink.exe",
        "https://github.com/cyd01/KiTTY/releases/download/v0.72.0.6/kscp.exe",
        "https://github.com/cyd01/KiTTY/releases/download/v0.72.0.6/ksftp.exe"
    ],
    "hash": [
        "13673f67164c1d0431ffdb8da93c97e64e18f0458eb7d30ebd52ee18f72ef94e",
        "caa4fe751ea4cee4ea5519f7c07489c990901e65d4e16726bebed5a6db8b5224",
        "9fe6bd585a5afc6c44bdccd1c9481c21c5680b5c91015c0224a575990f62211e",
        "3a719f9961c2b8bd493a56819c6dc816e1340e4ed3e20c61e565362c9a248ec2",
        "1263f5e679f6667e03cc1614c8a045c5406e79c23f004fa6577c1bdb3da6fa8c",
        "54c535598d61cc998af5591ada6d17e5e83a5308ba8aab9288ae3b9a1858cdf8",
        "2721eeba5d8fb77d0a8a0f661b5676092670cb9fd9ad02f32cb37575d42b62ce"
    ],
    "checkver": {
        "github": "https://github.com/cyd01/KiTTY"
    },
    "autoupdate": {
        "url": [
            "https://github.com/cyd01/KiTTY/releases/download/v$version/genpass.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/kittygen.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/kageant.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/kitty.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/klink.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/kscp.exe",
            "https://github.com/cyd01/KiTTY/releases/download/v$version/ksftp.exe"
        ]
    }
}
```

Ash258:
Also [this debug](https://github.com/niheaven/scoop/blob/multi-urls/lib/autoupdate.ps1#L411) is useless

B

Ash258:
Extract_dir issue

1. When original extract_dir is array then nothing is done
1. When it is string it is replaced with all replaced values joined with space

A

G

lib/autoupdate.ps1
niheaven: The following is more precise:

```powershell
    if ($updatedProperties -contains 'url') {
        $updatedProperties += 'hash'
    }
    $updatedProperties = $updatedProperties | Select-Object -Unique
```

But I don't think there will be `autoupdate` without `au.url`... So `$updatedProperties -notcontains 'hash'` should be okay.

niheaven:
> Force parameter is not respected correctly.

This is 'cos `hash.Length` is less then `au.hash.Length`, and `PropertyHelper` will just replace from top to bottom, with 'min(Length)' lines. So when all hash lines are replace with newer ones, and nothing changed, `checkver` will not overwrite manifest in such situation.

It's such a rare case that `url`, `hash`, `au.url` and `au.hash` don't have same length, it's a wrong manifest, right?

Ash258:
No it is not. This is use case when you are creting manifest. You just add autoupdate URLs and let checkver to update hashes automatically

niheaven:
> Also [this debug](https://github.com/niheaven/scoop/blob/multi-urls/lib/autoupdate.ps1#L411) [](/niheaven/scoop/blob/HEAD@{2019-10-25T19:37:50Z}/lib/autoupdate.ps1#L411) is not optimal and do not match style of other variables

It's a bug of `debug`

https://github.com/lukesampson/scoop/blob/6eb90c9c112c8c6c1790f1bd4e1c9cd67a6e70df/lib/core.ps1#L147-L151

L148 skip first two line of `$msg`, which may be usefull while `$obj` is not `[hashtable]`. Here `$updatedProperties` is `[Object[]]` and give no output if `debug $updatedProperties`...

lib/autoupdate.ps1
niheaven: One use case is when nothing changed in update, the manifest will not be updated, even with `-Force`.

Ash258: Manifest have to be updated all the time when Force is given.

lib/autoupdate.ps1
niheaven: This cause the following `extract_dir` bug, if `$Value` is `[Array]` and `$Property` is `[String]`. I'll check if `$Value = $Value[0]` works.

niheaven:
> No it is not. This is use case when you are creating manifest. You just add autoupdate URLs and let checkver to update hashes automatically

Okay, let's force replace `hash` when `hash`'s length is less then `au.url`'s.

Ash258:
No. When force is specified everything has to be replaced. It is nonse to limit it on hash.

When manfiest will be converted to support with multiple autoupdate properties. Everyone will just do this:

1. Add to autoupdate
1. [Optional] Convert current to array
1. Run checkver `Force`

And everything has to be updated

lib/autoupdate.ps1
niheaven: Even if nothing changed, with only the same file overwrited?

Ash258: Yes. That is how current checkver work.

Ash258: It is used mostly instead of `formatjson`. Since checkver is at the end calling ConvertToPrettyJson

niheaven:
Got it, if any property in `au` has length greater then the one outside, it should be replaced with new one when updated. If less then outside one, then only first xxx lines updated.

Ash258:
Exactly.

niheaven:
For the bug of `core/debug`, what's your opinion? @Ash258

Ash258:
I do not see a reason to not adopt same approach as here. If it is object keep as it is. When array it could be joined, or indented on each line with one item.

niheaven:
`debug $updatedProperties` will output `$updatedProperties (System.Object[])`, while `debug [String]$updatedProperties` will output `[String]$updatedProperties = [String]url hash`.

Without modification of `core/debug`, we must explicitly convert `[Array]` to `[String]` for correct output.

lib/autoupdate.ps1
niheaven: Fixed with length comparison.

lib/autoupdate.ps1
niheaven: This line is useless, I think...

https://github.com/lukesampson/scoop/blob/6eb90c9c112c8c6c1790f1bd4e1c9cd67a6e70df/lib/autoupdate.ps1#L431

niheaven:
@Ash258 A simple length comparison works for the update. If property in `au` has greater length, then it will replace the one outside.

Ash258:
Seems good on first test. I will get through edited code and do additional testin.

A

Ash258:
For some reason this PR is removing unusal characters.

A

```json
    "shortcuts": [
        [
            "ufm.exe",
            "UložTo File Manager"
        ]
    ],
```

niheaven:
The bug was introduced by `Set-Content -Encoding ASCII`, so change back to `[System.IO.File]::WriteAllLines`. `UTF8NoBOM` encoding in PowerShell 7 works well, we may change it later after updating minimal requirement to PS7.

Ash258:
If you will close debug param into brackets, then there is no redundant `[String]` inside value. But there are brackets in name.

A

niheaven:
`debug [$updatedProperties]` gives good output

```
DEBUG[1572282026.21928] [$updatedProperties] = [url hash]
```

lib/autoupdate.ps1
```diff
@@ -349,85 +403,156 @@ function get_version_substitutions([String] $version, [Hashtable] $customMatches
...
-            $hash = get_hash_for_app $app $json.autoupdate.hash $version $url $substitutions
-            if ($null -eq $hash) {
-                $valid = $false
-                Write-Host -f DarkRed "Could not find hash!"
-            }
-        }
+function Invoke-AutoUpdate {
+    param (
+        [String]
+        $App,
```
linsui: What's the difference between $App and $AppName ? :confused:

niheaven: One thing... I change them all to `AppName`, more formal.

rasa:
I’m comfortable merging this into develop. @r15ch13?

niheaven:
@r15ch13 @rasa @Ash258 Could this be merged into 'develop'?

r15ch13:
Could you squash some of the commits and change the commit messages to the new style?
I would like to merge this instead of squashing it to one big commit :smile:

niheaven:
Okay, I'll arrange the commits and rename them.

niheaven:
@r15ch13 I've force-pushed this PR, and original one is put at https://github.com/niheaven/scoop/tree/org-multi-urls.

There're some tweaks about updating notes, see diff below (already rebased to develop):

<details>
<summary>Diff between original PR and new one</summary>

```diff
diff --git a/lib/autoupdate.ps1 b/lib/autoupdate.ps1
index 62b01b9f..1248f695 100644
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -178,6 +178,9 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
     $substitutions.Add('$url', (strip_fragment $url))
     $substitutions.Add('$baseurl', (strip_filename (strip_fragment $url)).TrimEnd('/'))
     $substitutions.Add('$basename', $basename)
+    $substitutions.Add('$urlNoExt', (strip_ext (strip_fragment $url)))
+    $substitutions.Add('$basenameNoExt', (strip_ext $basename))
+
     debug $substitutions
 
     $hashfile_url = substitute $config.url $substitutions
@@ -442,9 +445,21 @@ function Invoke-AutoUpdate {
         # $Manifest | ConvertToPrettyJson | Set-Content -Path (Join-Path $Path "$AppName.json") -Encoding UTF8NoBOM
         [System.IO.File]::WriteAllLines((Join-Path $Path "$AppName.json"), (ConvertToPrettyJson $Manifest))
         # notes
+        $note = "`nUpdating note:"
         if ($Manifest.autoupdate.note) {
-            Write-Host ""
-            Write-Host $Manifest.autoupdate.note -ForegroundColor DarkYellow
+            $note += "`nno-arch: $($Manifest.autoupdate.note)"
+            $hasNote = $true
+        }
+        if ($Manifest.autoupdate.architecture) {
+            '64bit', '32bit' | ForEach-Object {
+                if ($Manifest.autoupdate.architecture.$_.note) {
+                    $note += "`n$_-arch: $($Manifest.autoupdate.architecture.$_.note)"
+                    $hasNote = $true
+                }
+            }
+        }
+        if ($hasNote) {
+            Write-Host $note -ForegroundColor DarkYellow
         }
     } else {
         # This if-else branch may not be in use.
diff --git a/lib/diagnostic.ps1 b/lib/diagnostic.ps1
index 4c67454c..9eb7579e 100644
--- a/lib/diagnostic.ps1
+++ b/lib/diagnostic.ps1
@@ -50,3 +50,14 @@ function check_long_paths {
 
     return $true
 }
+
+function check_envs_requirements {
+    if ($null -eq $env:COMSPEC) {
+        warn '$env:COMSPEC environment variable is missing.'
+        Write-Host "    By default the variable should points to the cmd.exe in Windows: '%SystemRoot%\system32\cmd.exe'."
+
+        return $false
+    }
+
+    return $true
+}
diff --git a/lib/git.ps1 b/lib/git.ps1
index d06aaa64..58556247 100644
--- a/lib/git.ps1
+++ b/lib/git.ps1
@@ -20,7 +20,7 @@ function git_checkout {
 }
 
 function git_pull {
-    git_proxy_cmd pull $args
+    git_proxy_cmd pull --rebase=false $args
 }
 
 function git_fetch {
diff --git a/lib/install.ps1 b/lib/install.ps1
index 41289501..4674f575 100644
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -1021,11 +1021,11 @@ function prune_installed($apps, $global) {
 
 # check whether the app failed to install
 function failed($app, $global) {
-    $ver = current_version $app $global
-    if(!$ver) { return $false }
-    $info = install_info $app $ver $global
-    if(!$info) { return $true }
-    return $false
+    if (is_directory (appdir $app $global)) {
+        return !(install_info $app (current_version $app $global) $global)
+    } else {
+        return $false
+    }
 }
 
 function ensure_none_failed($apps, $global) {
diff --git a/lib/manifest.ps1 b/lib/manifest.ps1
index acb11891..f2455d43 100644
--- a/lib/manifest.ps1
+++ b/lib/manifest.ps1
@@ -59,8 +59,20 @@ function install_info($app, $version, $global) {
 }
 
 function default_architecture {
-    if ([Environment]::Is64BitOperatingSystem) { return "64bit" }
-    "32bit"
+    $arch = get_config 'default-architecture'
+    $system = if ([Environment]::Is64BitOperatingSystem) { '64bit' } else { '32bit' }
+    if ($null -eq $arch) {
+        $arch = $system
+    } else {
+        try {
+            $arch = ensure_architecture $arch
+        } catch {
+            warn 'Invalid default architecture configured. Determining default system architecture'
+            $arch = $system
+        }
+    }
+
+    return $arch
 }
 
 function arch_specific($prop, $manifest, $architecture) {
diff --git a/libexec/scoop-checkup.ps1 b/libexec/scoop-checkup.ps1
index df4a8b6e..d515469c 100644
--- a/libexec/scoop-checkup.ps1
+++ b/libexec/scoop-checkup.ps1
@@ -12,6 +12,7 @@ $issues += !(check_windows_defender $false)
 $issues += !(check_windows_defender $true)
 $issues += !(check_main_bucket)
 $issues += !(check_long_paths)
+$issues += !(check_envs_requirements)
 
 if (!(Test-HelperInstalled -Helper 7zip)) {
     error "'7-Zip' is not installed! It's required for unpacking most programs. Please Run 'scoop install 7zip' or 'scoop install 7zip-zstd'."
diff --git a/libexec/scoop-info.ps1 b/libexec/scoop-info.ps1
index fd7ef3bb..f72416c8 100644
--- a/libexec/scoop-info.ps1
+++ b/libexec/scoop-info.ps1
@@ -35,6 +35,7 @@ if (!$manifest) {
 }
 
 $install = install_info $app $status.version $global
+$status.installed = $install.bucket -eq $bucket
 $version_output = $manifest.version
 if (!$manifest_file) {
     $manifest_file = manifest_path $app $bucket
diff --git a/libexec/scoop-list.ps1 b/libexec/scoop-list.ps1
index de366ddf..44c45a74 100644
--- a/libexec/scoop-list.ps1
+++ b/libexec/scoop-list.ps1
@@ -32,7 +32,7 @@ if($apps) {
         if (!$install_info) { Write-Host ' *failed*' -ForegroundColor DarkRed -NoNewline }
         if ($install_info.hold) { Write-Host ' *hold*' -ForegroundColor DarkMagenta -NoNewline }
 
-        if ($install_info.bucket -and ($install_info.bucket -ne 'main')) {
+        if ($install_info.bucket) {
             write-host -f Yellow " [$($install_info.bucket)]" -NoNewline
         } elseif ($install_info.url) {
             write-host -f Yellow " [$($install_info.url)]" -NoNewline
diff --git a/libexec/scoop-update.ps1 b/libexec/scoop-update.ps1
index 9227e6b9..05095fd2 100644
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -239,6 +239,7 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
     # endregion Workaround
 
     $dir = versiondir $app $old_version $global
+    $persist_dir = persistdir $app $global
 
     #region Workaround for #2952
     $processdir = appdir $app $global | Resolve-Path | Select-Object -ExpandProperty Path
diff --git a/schema.json b/schema.json
index ed45a847..246550f2 100644
--- a/schema.json
+++ b/schema.json
@@ -107,6 +107,9 @@
                 "bin": {
                     "$ref": "#/definitions/stringOrArrayOfStringsOrAnArrayOfArrayOfStrings"
                 },
+                "checkver": {
+                    "$ref": "#/definitions/checkver"
+                },
                 "env_add_path": {
                     "$ref": "#/definitions/stringOrArrayOfStrings"
                 },
@@ -217,6 +220,9 @@
                 },
                 "persist": {
                     "$ref": "#/definitions/stringOrArrayOfStringsOrAnArrayOfArrayOfStrings"
+                },
+                "note": {
+                    "$ref": "#/definitions/stringOrArrayOfStrings"
                 }
             },
             "type": "object"
```
</details>

niheaven:
Ping @r15ch13

batkiz:
Does this PR have been forgotten?
I'm experiencing autoupdate url arrays issue, hope this PR would be merged as soon as possible.

niheaven:
I'll check it again and wait for @r15ch13

rashil2000:
@niheaven Please go ahead and merge this

PR #3522, opened 2019-06-21
feat: load proxy from env for aria2

Cologler:


r15ch13:
aria2 recognizes these environment variables on its own. See documentation:
https://aria2.github.io/manual/en/html/aria2c.html#environment

Cologler:
It seems worked already (in the latest version).

PR #3527, opened 2019-06-22
fix(checkhashes): Do not call scoop directly

Ash258:
Opening to master since, it is not invansive and fast fix

# Background

I need to be able to run binaries from linux hosts
https://github.com/Ash258/GithubActionsBucketForTesting/runs/153802813

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,17 +132,17 @@ If you find Scoop useful and would like to support ongoing development and maint
 
 The following buckets are known to scoop:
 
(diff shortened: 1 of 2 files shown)


Ash258:
Also fix failing build introduced in 2849e0f96099004f761d7d8c715377e0d2c105f2

Ash258:
If someone is concerned about actions I am on good way 🎉🎉

https://github.com/Ash258/GithubActionsBucketForTesting/runs/153809021
https://github.com/Ash258/GithubActionsBucketForTesting/issues/15#issuecomment-504693152

PR #3533, opened 2019-06-27
refactor(core): Get rid of 'fullpath'

niheaven:
Get rid of `fullpath`, just get absolute path when initialing `$globaldir`, `$scoopdir` and `$cachedir`, other dirs are derived from these three root ones.

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -450,8 +450,8 @@ function Get-CommandPath {
     )
 
     begin {
-        $userShims = Convert-Path (shimdir $false)
-        $globalShims = fullpath (shimdir $true) # don't resolve: may not exist
+        $userShims = shimdir $false
+        $globalShims = shimdir $true
     }
 
     process {
@@ -571,9 +571,33 @@ function ensure($dir) {
     }
     Convert-Path -Path $dir
 }
+function Get-AbsolutePath {
+    <#
+    .SYNOPSIS
+        Get absolute path
+    .DESCRIPTION
+        Get absolute path, even if not existed
+    .PARAMETER Path
+        Path to manipulate
+    .OUTPUTS
+        System.String
+            Absolute path, may or maynot existed
+    #>
+    [CmdletBinding()]
+    [OutputType([string])]
+    param (
+        [Parameter(Mandatory = $true, ValueFromPipeline = $true)]
+        [string]
+        $Path
+    )
+    process {
+        return $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($Path)
+    }
+}
+
 function fullpath($path) {
-    # should be ~ rooted
-    $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($path)
+    Show-DeprecatedWarning $MyInvocation 'Get-AbsolutePath'
+    return Get-AbsolutePath -Path $path
 }
 function friendly_path($path) {
     $h = (Get-PSProvider 'FileSystem').Home
@@ -1300,7 +1324,7 @@ if ($pathExpected) {
     # ├─shims
     # ├─config.json
     # ```
-    $configPortablePath = fullpath "$coreRoot\..\..\..\config.json"
+    $configPortablePath = Get-AbsolutePath "$coreRoot\..\..\..\config.json"
     if (Test-Path $configPortablePath) {
         $configFile = $configPortablePath
     }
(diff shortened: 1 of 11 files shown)


lib/core.ps1
```diff
@@ -352,8 +352,12 @@ function url_remote_filename($url) {
...
 
 function ensure($dir) { if(!(test-path $dir)) { mkdir $dir > $null }; resolve-path $dir }
-function fullpath($path) { # should be ~ rooted
-    $executionContext.sessionState.path.getUnresolvedProviderPathFromPSPath($path)
+function Get-AbsolutePath {
+    param (
+        [String]
+        $Path
+    ) # should be ~ rooted
+    $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($Path)
```
Ash258:
```suggestion
    return $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($Path)
```

niheaven:
Removed `is_in_dir` test case will never be met when using `scoop`.

lib/core.ps1
```diff
@@ -352,8 +352,12 @@ function url_remote_filename($url) {
 }
 
 function ensure($dir) { if(!(test-path $dir)) { mkdir $dir > $null }; resolve-path $dir }
-function fullpath($path) { # should be ~ rooted
-    $executionContext.sessionState.path.getUnresolvedProviderPathFromPSPath($path)
+function Get-AbsolutePath {
+    param (
+        [String]
```
Ash258: Use mandatory parameter

lib/core.ps1
```diff
@@ -352,8 +352,17 @@ function url_remote_filename($url) {
...
+    param (
+        [Parameter(Mandatory = $true)]
+        [String]
+        $Path
+    ) # should be ~ rooted
+    return $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($Path)
+}
+function fullpath($path) {
+    Show-DeprecatedWarning $MyInvocation 'Get-AbsolutePath'
+    Get-AbsolutePath -Path $path
```
Ash258: It is not meant to be void

```suggestion
    return Get-AbsolutePath -Path $path
```

lib/core.ps1
```diff
@@ -352,8 +352,17 @@ function url_remote_filename($url) {
 }
 
 function ensure($dir) { if(!(test-path $dir)) { mkdir $dir > $null }; resolve-path $dir }
-function fullpath($path) { # should be ~ rooted
-    $executionContext.sessionState.path.getUnresolvedProviderPathFromPSPath($path)
+function Get-AbsolutePath {
+    param (
```
Ash258: This is one parameter. It could be one liner.

lib/core.ps1
```diff
@@ -352,8 +352,17 @@ function url_remote_filename($url) {
 }
 
 function ensure($dir) { if(!(test-path $dir)) { mkdir $dir > $null }; resolve-path $dir }
-function fullpath($path) { # should be ~ rooted
-    $executionContext.sessionState.path.getUnresolvedProviderPathFromPSPath($path)
+function Get-AbsolutePath {
```
Ash258: Add synopsis, Cmdletbinding, Outputtype, Pipeline support

lib/core.ps1
```diff
@@ -351,15 +351,32 @@ function url_remote_filename($url) {
...
+    [CmdletBinding()]
+    [OutputType([String])]
+    Param (
+        [Parameter(Mandatory = $true, ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true)]
         [String]
         $Path
-    ) # should be ~ rooted
-    return $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($Path)
+    )
+    Process {
```
Ash258: What is advantage of having process for one line body function?

niheaven: Just for pipeline. Indeed, there're only three usages in scoop now, and if not used in manifest's script, there is no need to support pipeline.

lib/core.ps1
```diff
@@ -351,9 +351,35 @@ function url_remote_filename($url) {
...
+            Get absolute path, even if not existed
+        .PARAMETER Path
+            Path to manipulate
+        .OUTPUTS
+            System.String
+                Absolute path, may or maynot existed
+    #>
+    [CmdletBinding()]
+    [OutputType([String])]
+    Param (
```
Ash258:
```suggestion
    param (
```

lib/core.ps1
```diff
@@ -351,9 +351,35 @@ function url_remote_filename($url) {
...
+        [Parameter(Mandatory = $true, ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true)]
+        [String]
+        $Path
+    )
+    Process {
```
Ash258:
```suggestion
    process {
```

lib/core.ps1
```diff
@@ -351,9 +351,35 @@ function url_remote_filename($url) {
...
 }
 
-function ensure($dir) { if(!(test-path $dir)) { mkdir $dir > $null }; resolve-path $dir }
-function fullpath($path) { # should be ~ rooted
-    $executionContext.sessionState.path.getUnresolvedProviderPathFromPSPath($path)
+function ensure($dir) { if (!(test-path $dir)) { mkdir $dir > $null }; resolve-path $dir }
+
+function Get-AbsolutePath {
+    <#
+        .SYNOPSIS
```
Ash258: Synopsis is not indented from comment start in while codebase

https://github.com/lukesampson/scoop/search?q=synopsis&unscoped_q=synopsis

Ash258:
This will break Excavator due to path glueing with hardcoded backslash.

image

image

PR #3502, opened 2019-06-10
feat(install): Add unpacker for 'wix' and 'nsis' installer

niheaven:
- Add a new property `installer.type` (`nsis|wix|inno`)
  - `nsis`: `Expand-Installer` accepts `$Architecture` (`64bit|32bit`) and unpack Electron's NSIS installer (`$PLUGINSDIR/app-32|64.7z`)
  - `wix`: `Expand-WixInstaller` use `Expand-DarkArchive` first and then `Expand-MsiArchive` all msi under `AttachedContainer` except those excluded by `installer.exclude`
  - `inno`: Move `innosetup` here for consistency, i.e. `innosetup:true` = `installer.type:inno` (autoupdate will remove `innosetup:true` and add `installer.type:inno` automatically)

Some Examples:

<details>
<summary>`python` manifest</summary>

```json
"installer": {
    "type": "wix",
    "exclude": [
        "launcher.msi",
        "path.msi",
        "pip.msi"
    ],
    "script": [
        "&(\"$dir\\python.exe\") -E -s -m ensurepip -U --default-pip | Out-Null",
        "if ($global) { $pathext = (env 'PATHEXT' $true) -replace ';.PYW?', ''; env 'PATHEXT' $true \"$pathext;.PY;.PYW\" }"
    ]
},
```

</details>

<details>
<summary>`mro` manifest</summary>

```json
"installer": {
    "type": "wix",
    "script": [
        "movedir \"$dir\\Microsoft\\MRO-$version.0\" \"$dir\" | Out-Null",
        "Expand-7zipArchive \"$dir\\Setup\\*.cab\"",
        "Remove-Item \"$dir\\bin\\x64\\Rblas.dll\", \"$dir\\bin\\x64\\Rlapack.dll\" -Force",
        "movedir \"$dir\\Setup\\bin\\x64\" \"$dir\\bin\\x64\" | Out-Null",
        "movedir \"$dir\\Setup\\library\" \"$dir\\library\" | Out-Null",
        "Remove-Item \"$dir\\Setup\", \"$dir\\Microsoft\" -Force -Recurse"
    ]
},
```

</details>

<details>
<summary>`simplenote` manifest</summary>

```json
"installer": {
    "type": "nsis"
},
```

</details>

PR #3534, opened 2019-06-27
opt(decompress): Add [Array] and pipeline support to `Expand-XXX`'s params

niheaven:
Now `Expand-xxx` functions accept [Array] params (`Path`, `DestinationPath`, `ExtractDir`), and work for pipeline params (`aa, bb, cc | Expand-xxx` or `@(aa, bb, cc) | Expand-xxx`.

No more `ForEach-Object { Expand-xxx $_ }`.

### DIFF
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -39,12 +39,13 @@ function Expand-7zipArchive {
     [CmdletBinding()]
     param (
         [Parameter(Mandatory = $true, Position = 0, ValueFromPipeline = $true)]
-        [String]
+        [String[]]
         $Path,
-        [Parameter(Position = 1)]
-        [String]
-        $DestinationPath = (Split-Path $Path),
-        [String]
+        [Parameter(Position = 1, ValueFromPipelineByPropertyName = $true)]
+        [String[]]
+        $DestinationPath,
+        [Parameter(Position = 2, ValueFromPipelineByPropertyName = $true)]
+        [String[]]
         $ExtractDir,
         [Parameter(ValueFromRemainingArguments = $true)]
         [String]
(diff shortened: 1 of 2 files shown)


lib/decompress.ps1
```diff
@@ -55,206 +56,275 @@ function Expand-7zipArchive {
...
+        if (!$DestinationPath) {
+            $DestinationPath = Split-Path -Path $Path
+        }
+        for ($i = 0; $i -lt $Path.Length; $i++) {
+            $aPath = $Path[$i]
+            if ($aDestinationPath = $DestinationPath[$i]) {
+                $aDestinationPath = $DestinationPath[-1]
+            }
+            if ($ExtractDir) {
+                if ($aExtractDir = $ExtractDir[$i]) {
```
Ash258: What are these check for?? It does not make sense at all.

Why do you need to check if assigning to variable was possible? For index out of bounds??

Sidenote: Naming `$aArray` is completely misleading and non-intuitive. Rather go with `currentDestination`

niheaven: That's for index out of bounds, e.g., `Expand-xxx -Path @(xxx) -Destination xxx` will extract all archives into one location, so it use `$DestinationPath[-1]`.

Assigning variable will be done without error, but `$aDestinationPath` may be `$null`, the `if` is checking not null. i.e., it is shorten for

```json
$aDestinationPath = $DestinationPath[$i]
if ($null -eq $aDestinationPath) {
    $aDestinationPath = $DestinationPath[-1]
}
```

I use `$aXXXXX` for noting singular variable, 'a DestinationPath'. Maybe `current` is better except for too long variable name...

Ash258: What is wrong with?

```powershell
if ($null -eq $DestinationPath[$i]) {
    .....
}
```

niheaven: That's okay, too. You mean?

```json
if ($null -eq $DestinationPath[$i]) {
    $aDestinationPath = $DestinationPath[$i]
} else {
    $aDestinationPath = $DestinationPath[-1]
}
```

lib/decompress.ps1
```diff
@@ -55,206 +56,291 @@ function Expand-7zipArchive {
...
+                if ($null -eq $ExtractDir[$i]) {
+                    $aExtractDir = $ExtractDir[$i]
+                } else {
+                    $aExtractDir = $ExtractDir[-1]
+                }
+            }
+            $LogPath = "$(Split-Path $aPath)\7zip.log"
+            $ArgList = @('x', "`"$aPath`"", "-o`"$aDestinationPath`"", '-y')
+            $IsTar = ((strip_ext $aPath) -match '\.tar$') -or ($aPath -match '\.t[abgpx]z2?$')
+            if (!$IsTar -and $aExtractDir) {
```
r15ch13: `$aExtractDir` is not initialized if `$ExtractDir` is not set.

niheaven: Oop, it should be `-ne` here...

Ash258:
@niheaven Could you also look into this?

```json
{
    "version": "5.4.6.0",
    "url": "http://www.vieas.com/software/dl/Vieas5460_32.zip#/cosi",
    "pre_install": "\"$dir\\$fname\" | Expand-7zipArchive; 'WRONG'",
    "post_install": "\"$dir\\$fname\" | Expand-7zipArchive -DestinationPath $dir; 'OK'"
}
```

When using pipeline, default DestinationPath is not resolved as `$Path` is $null and I do not know why, as `$Path` is position 0 and set to `ValueFromPipeline`

Ash258:
OK.Nevermind. This come from PowerShell precedence of execution.

Parameters are validated and binded before pipeline input processing. 

image

This could be solved with removing default from `$DestinationPath` parameter and then do simple check in begin block or function body `if(-not $DestinationPath) { $DestinationPath = (Split-Path $Path) }`

niheaven:
@Ash258 I've found it and fixed in this PR, it should be okay when this is merged. 😄

lib/decompress.ps1
```diff
@@ -55,206 +56,299 @@ function Expand-7zipArchive {
...
+                $ArgList += "-ir!`"$currentExtractDir\*`""
+            }
+            if ($Switches) {
+                $ArgList += (-split $Switches)
+            }
+            switch ($Overwrite) {
+                "All" { $ArgList += "-aoa" }
+                "Skip" { $ArgList += "-aos" }
+                "Rename" { $ArgList += "-aou" }
+            }
```
Ash258: Use single quotes

lib/decompress.ps1
```diff
@@ -55,206 +56,299 @@ function Expand-7zipArchive {
...
+                    $currentExtractDir = $ExtractDir[-1]
+                } else {
+                    $currentExtractDir = $ExtractDir[$i]
+                }
+            } else {
+                $currentExtractDir = $null
+            }
+            $currentDestinationPath = $currentDestinationPath.TrimEnd("\")
+            if ($currentExtractDir) {
+                $OriDestinationPath = $currentDestinationPath
```
Ash258: local variable

```suggestion
                $oriDestinationPath = $currentDestinationPath
```

lib/decompress.ps1
```diff
@@ -55,206 +56,299 @@ function Expand-7zipArchive {
...
+            } else {
+                $currentExtractDir = $null
+            }
+            $LogPath = "$(Split-Path $currentPath)\innounp.log"
+            $ArgList = @('-x', "-d`"$currentDestinationPath`"", "`"$currentPath`"", '-y')
+            switch -Regex ($currentExtractDir) {
+                "^[^{].*" { $ArgList += "-c{app}\$currentExtractDir" }
+                "^{.*" { $ArgList += "-c$currentExtractDir" }
+                Default { $ArgList += "-c{app}" }
+            }
```
Ash258: - single quotes
- default is lowercased when switch is lowercased

lib/decompress.ps1
```diff
@@ -55,206 +56,299 @@ function Expand-7zipArchive {
...
+                    $currentExtractDir = $ExtractDir[-1]
+                } else {
+                    $currentExtractDir = $ExtractDir[$i]
+                }
             } else {
-                abort "Unzip failed: Windows can't handle the long paths in this zip file.`nRun 'scoop install 7zip' and try again."
+                $currentExtractDir = $null
+            }
+            if ($currentExtractDir) {
+                $OriDestinationPath = $currentDestinationPath
```
Ash258: local varaible

lib/decompress.ps1
```diff
@@ -55,206 +56,299 @@ function Expand-7zipArchive {
...
+            # All methods to unzip the file require .NET4.5+
+            if ($PSVersionTable.PSVersion.Major -lt 5) {
+                Add-Type -AssemblyName System.IO.Compression.FileSystem
+                try {
+                    [System.IO.Compression.ZipFile]::ExtractToDirectory($currentPath, $currentDestinationPath)
+                } catch [System.IO.PathTooLongException] {
+                    # try to fall back to 7zip if path is too long
+                    if (Test-HelperInstalled -Helper 7zip) {
+                        Expand-7zipArchive $currentPath $currentDestinationPath -Removal
+                        return
```
Ash258: Why would you break whole foreach?

niheaven:
@Ash258 I would like to rebase this PR onto #3502 and check the whole code again after #3502 is merged. There're many merging conflicts between these two PRs...

Issue #2519, opened 2018-08-22
Add libsndfile formula


MartinDelille:
Homepage here: http://www.mega-nerd.com/libsndfile/

MartinDelille:
@tresf You mention *libsndfile* here: https://github.com/lukesampson/scoop/pull/2117#issue-174238690

Have you ever created a formula for this library?

PR #3540, opened 2019-07-02
fix(decompress): fix bugs on extract_dir

issaclin32:
This fixes the error when `extract_dir` contains spaces, such as "ISLC v1.0.1.2"
(encountered in: https://github.com/lukesampson/scoop-extras/pull/2398)

Please let me know if there are any problems. Thanks.

### DIFF
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -68,7 +68,7 @@ function Expand-7zipArchive {
     $ArgList = @('x', "`"$Path`"", "-o`"$DestinationPath`"", '-y')
     $IsTar = ((strip_ext $Path) -match '\.tar$') -or ($Path -match '\.t[abgpx]z2?$')
     if (!$IsTar -and $ExtractDir) {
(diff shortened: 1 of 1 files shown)


Ash258:
Change base branch to develop

r15ch13:
This would result in `"-ir!C:\bla\bla\"*`, but shouldn't it be `"-ir!C:\bla\bla\*"`?

issaclin32:
> This would result in `"-ir!C:\bla\bla\"*`,
> but shouldn't it be `"-ir!C:\bla\bla\*"` or `-ir!"C:\bla\bla\*"`?

Changing it to `-ir!"C:\bla\bla\*"` now. Thanks for the suggestion.

I thought that would result in taking `*` literally and won't work, but it works just fine.

Luckz:
I'd still really like to install ISLC some day 🥇

gothicx:
Closes #2519

Issue #3542, opened 2019-07-03
More challenge when uninstall scoop


Cologler:
Try to uninstall scoop require the user to input a random (or complex) string instead of a simple `Y`.

kroppt:
Would this be to prevent someone from deleting your scoop or from accidentally doing so?
A script could easily parse the required input.

Cologler:
Some people (like me) won't read warning messages and direct type `Y`.

kroppt:
Other apps do not ask for any input to do the uninstall, so I think the y/n accomplishes this for uninstalling `scoop`. In this, case there is also a warning message.

However, I think that this addition would be good, because `scoop` is simple enough for knuckle heads to use (not saying that OP is one). Someone who is not a native-English speaker would also benefit from an extra check.

It is a known phenomenon that when a tool is easy to use, the average skill of the users actually lowers as more people start using it.

If this were added, there should be a flag for skipping the query, so that it can be used more easily in scripts.

PR #3543, opened 2019-07-03
feat: random code for uninstall scoop

Cologler:
for #3542 
😭


rashil2000:
This is not required. The `yN` prompt is sufficient.

PR #3547, opened 2019-07-04
opt(Invoke-ExternalCommand): Small tweak

niheaven:
- Also redirect standard error to log file.
- Fix `-RunAs` error.
  - `-RunAs` and `-Log` cannot be used together.

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -373,7 +373,7 @@ function run($exe, $arg, $msg, $continue_exit_codes) {
 }
 
 function Invoke-ExternalCommand {
-    [CmdletBinding()]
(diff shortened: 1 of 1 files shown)


r15ch13:
Why did you add `ParameterSetName` if you don't use the feature? :smile:
https://blog.simonw.se/powershell-functions-and-parameter-sets/

niheaven:
> `-RunAs` and `-Log` cannot be used together.

`Verb = RunAs` needs `UseShellExcute = $true` while `RedirectStandardOutput` needs `UseShellExcute = $false`, so I add `ParameterSetName` to give an error.

r15ch13:
Oh, I understand it now. Sorry 😅

PR #3562, opened 2019-07-17
fix(install): allow TLSv1.0/1.1 for aria2

issaclin32:
This changes aria2c's argument `--min-tls-version` from **TLSv1.2** to ~SSLv3~ **TLSv1**.

SSLv3 and TLS v1.0/1.1 are deprecated due to security issues, but some sites are still using it. This causes `aria2` to fail for some packages. (such as: https://github.com/lukesampson/scoop-extras/issues/2449) **(SSL/TLS handshake error)**

Scoop does not restrict SSL/TLS version When `aria2` is disabled. Therefore, I don't see a reason to restrict TLS version when using `aria2`.

(Possible options for `--min-tls-version` can be found [here](https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-min-tls-version))

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -222,7 +222,7 @@ function dl_with_cache_aria2($app, $version, $manifest, $architecture, $dir, $co
         "--no-conf=true"
         "--follow-metalink=true"
         "--metalink-preferred-protocol=https"
-        "--min-tls-version=SSLv3"
+        "--min-tls-version=TLSv1"
         "--stop-with-process=$PID"
         "--continue"
     )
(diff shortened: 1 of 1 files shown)


r15ch13:
I tested all URLs found in Scoops manifest files and I couldn't find any that don't have at least one TLS version active. (See: https://github.com/lukesampson/scoop/issues/2040#issuecomment-369705709)

Your example of `dl5.oo-software.com` also supports TLSv1.0 but they use a broken Cypher. (See: https://www.ssllabs.com/ssltest/analyze.html?d=dl5.oo-software.com)

I don't see any reason to degrade to SSLv3. Making it configurable would be okay.
If they offer HTTPS then they should at least try to use modern standards...

issaclin32:
Thanks for your reply.

I see there is no need to downgrade to `SSLv3`. Would it be better to change `--min-tls-version` to `TLSv1`? or just keep it as it is?

In these kind of cases, the problem is caused by that site maintainers is not keeping their HTTPS protocols up to date. However, I think most "normal users" would not figure that out. They will just blame `aria2` for failing in some packages.

issaclin32:
I am closing the PR because it is outdated.

PR #3570, opened 2019-07-25
Fix checksum regexs.

NamelessUzer:
Fix checksum regexs.


niheaven:
What's the use case?

ssfjhh:
Make the regex more precise.

rashil2000:
This is not required. The current regex allows for more flexibility in length.

PR #3610, opened 2019-08-25
fix(update,uninstall): Regression from get-process workaround

Ash258:
Get-process with path on splited `$dir` resolved to `$env:SCOOP\apps` when application installation was broken.

Broken installations have path `$env:SCOOP\apps\<app>\` and splited result in location of all applications.

Since this is just workaround I splitted path twice to check leaf. Every opinon is welcomed. (Maybe `EndsWith` could work 🤔)

A

### DIFF
--- a/libexec/scoop-uninstall.ps1
+++ b/libexec/scoop-uninstall.ps1
@@ -56,10 +56,12 @@ if (!$apps) { exit 0 }
     $dir = versiondir $app $version $global
     $persist_dir = persistdir $app $global
 
-
     #region Workaround for #2952
     # Split-path for getting $env:SCOOP\apps\<app> and not current or specific version
-    if (Get-Process | Where-Object { $_.Path -like "$(Split-Path $dir)\*" }) {
+    $split = Split-Path $dir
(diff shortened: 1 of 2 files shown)


r15ch13:
Why not simply use `appdir()`?
https://github.com/lukesampson/scoop/blob/79f8538b57b9021db71a279879b9032fefd1ae52/lib/core.ps1#L179

```pwsh
$appdir = appdir $app $global
if (Get-Process | Where-Object { $_.Path -like "$appdir\*" }) {
# snip
```

Ash258:
Good catch. Thank you.

`Resolve-Path` is needed as appdir is returning `C:\SCOOP\\apps\` in my case

r15ch13:
Did you add `\` to the ending of `$env:SCOOP`?

Ash258:
Yes

r15ch13:
It shouldn't end with a `\`.
 https://github.com/lukesampson/scoop/blob/79f8538b57b9021db71a279879b9032fefd1ae52/lib/core.ps1#L938

Ash258:
That code snippet has nothing to do with restrictions of trailing slash.

PR #3613, opened 2019-08-28
Update version pattern to support latest flutter v1.7.8+hotfix.4

hanabi1224:


schema.json
```diff
@@ -519,7 +519,7 @@
             "$ref": "#/definitions/uriOrArrayOfUris"
         },
         "version": {
-            "pattern": "^[\\w\\.\\-_]+$",
+            "pattern": "^[\\w\\.\\+\\-_]+$",
```
Ash258:
```suggestion
            "pattern": "^[\\w.+-_]+$",
```

linsui: Why not `"^[\\w-.+_]+$"` ?

Ash258: I do not know precedence of evaluating in powershell. Because `-` is control sequence character in character classes `+-.` mean "Everything between plus and dot" == "Everything between 43 and 46 character index"

niheaven: It should be `"^[\\w.+_-]+$"`, i.e., putting `-` as latest character will get literal `-`. (https://regex101.com/r/ekwiw7/1)

Ash258: I think it could be even after `\w`, but that depends on implementation. You could put it on the end, but i would prefer to escape it as it will not be problem in any of implementations.

niheaven: Yes, `^[\\w-.+_]+$` should work, too. PowerShell (or .NET) follows PCRE flavor and `-` follows `\w` means `-`. (https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expressions#how-regular-expressions-work)

hanabi1224:
@Ash258 Fixed, plz let me know if there's anything else outstanding

Ash258:
This is working and good, but I cannot approve this PR as I do not know initial reason why there are some characters forbidden in version property.

hanabi1224:
@Ash258 Thx, is there anything I can do to move this forward?

hanabi1224:
kindly ping

hanabi1224:
Kindly ping

niheaven:
It is very strang for their usage of plus sign '+', and please wait for new version comparison function.

hanabi1224:
@niheaven They intentionally change - to +. explanation here. https://github.com/flutter/flutter/wiki/Release-process#applying-emergency-fixes

niheaven:
Fixed in #4504

PR #3615, opened 2019-08-29
fix(autoupdate): Decode basename when extract hash

linsui:
Firefox, Thunderbird and uget can benefit.

### DIFF
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -172,7 +172,7 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
     $hash = $null
 
     $hashmode = $config.mode
-    $basename = url_remote_filename($url)
+    $basename = [System.Web.HttpUtility]::UrlDecode((url_remote_filename($url)))
 
     $substitutions = $substitutions.Clone()
     $substitutions.Add('$url', (strip_fragment $url))


Ash258:
What does have basename string to do with hash extraction templates?? Do not add it into hash templates

linsui:
substitutions will be regexEscape.

https://github.com/lukesampson/scoop/blob/79f8538b57b9021db71a279879b9032fefd1ae52/lib/autoupdate.ps1#L58

Ash258:
Would not be more suitable to url decode it rather than specifiing all possible escapes?

`[System.Web.HttpUtility]::UrlDecode('uget-2.2.2-win32%2Bgtk3.7z')`

image

Relate ScoopInstaller/Main#471

linsui:
Good idea! Would you pr? Then I'll close this.

h404bi:
> Good idea! Would you pr? Then I'll close this.

Why not update your pull request?

linsui:
I don't want to steal credit :joy: . If @Ash258 doesn't want to open another pr I'm glad to update this.

h404bi:
> I don't want to steal credit 😂 . If @Ash258 doesn't want to open another pr I'm glad to update this.

You can commit with`Co-authored-by` like [this](https://github.com/ScoopInstaller/Main/commit/ccc24e57b6026f88d5a7b1c11150e83487a016e1)

linsui:
Thanks! I'll have a try.

Ash258:
I do not care for contributions. Edit this PR.

PR #3619, opened 2019-09-01
fix(install) execution policy check

copdips:
also with a tiny script formatting.


r15ch13:
@copdips thanks 👍 
@h404bi this should probably be ported to [new installer `install.ps1`](https://github.com/ScoopInstaller/Install/blob/56188a9bdbfc5d2c399a28c59ebd5b32b2fb7232/install.ps1#L139-L142) as well

PR #3628, opened 2019-09-05
Fix to stop all updates when only one app failed to update.

lyuha:


libexec/scoop-update.ps1
```diff
@@ -331,9 +331,14 @@ if (!$apps) {
         }
     }
 
-    $suggested = @{};
+    $suggested = @{ };
     # $outdated is a list of ($app, $global) tuples
-    $outdated | ForEach-Object { update @_ $quiet $independent $suggested $use_cache $check_hash }
+    $outdated | ForEach-Object { try {
```
Ash258:
```suggestion
    $outdated | ForEach-Object {
        try {
```

libexec/scoop-update.ps1
```diff
@@ -331,9 +331,14 @@ if (!$apps) {
...
+            update @_ $quiet $independent $suggested $use_cache $check_hash
+        } catch {
+            Write-Error $_
```
Ash258:
```suggestion
            error $_
```

Ash258: What do you want to show in this error? Name of manifest or error message?

Ash258:
I do not think this will work as there are `abort` calls.

Ash258:
```powershell
# File B.ps1
function Abort-Mock {
    write-host 'before calling exit'
    exit 1
}

# File A.ps1
. $psscriptroot\B.ps1

try {
	Abort-Mock
} catch {
    'ahoj'
}

Write-host 'after exit call'
```

Result is always only message "Before calling exit"

lyuha:
Thanks to review.

I can't find some way to trap or catch `abort` calls. But, why handles `abort` calls at same level process? `exit 1` means to terminate process and report error to parent process. It does not use to handle in same level.

libexec/scoop-update.ps1
lyuha: I want show error including error message for report issue.

Ash258:
I am just saying that this will not fix problem when there is some hash check fail or installation error. Because on hash check fail there is `abort` call which have `exit 1`, which will stop whole scoop process all the time with or without try catch block. So this "fix" will not fix anything.

lyuha:
Good description.

I checked codes to find error I met. Found it

If found new version app, scoop try download it using `dl_with_cache` function. `dl_with_cache` function calls `do_dl`. Inside `do_dl` function, it has `try-catch` and `throw $e`. This error is not handled while update.

This changes fix above problem. However I feels this pr is not right way.

Why `do_dl` catch error and throw agian?

PR #3669, opened 2019-10-03
docs(readme): Update badges, add awesome and deprecate gitter

chawyehsu:


trip23app:
**deprecate gitter**

are Discord chats public? is Discord web interface 1000 times better than the **native Discord app experience**?

PR #3676, opened 2019-10-05
feat(core/au): Handle mediafire.com downloads

r15ch13:
This would fix [Musicbee](https://github.com/lukesampson/scoop-extras/blob/master/bucket/musicbee.json)

### DIFF
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -224,6 +224,11 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
         $hashmode = 'sourceforge'
     }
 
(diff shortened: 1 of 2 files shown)


lib/autoupdate.ps1
```diff
@@ -224,6 +224,11 @@ function get_hash_for_app([String] $app, $config, [String] $version, [String] $u
         $hashmode = 'sourceforge'
     }
 
+    if (!$hashfile_url -and $url -match "^.*mediafire.com\/\w+\/(?<random>\w+)\/(?<filename>.*)\/file(?:#\/.*)?$") {
+        $hashmode = 'mediafire'
```
Ash258: Using hashmode variable, which is used for `hash` extraction, to extract URL is pretty confusing

Ash258: Would not be better to just do `Invoke-RestMethod $url -match ...` here?

PR #3680, opened 2019-10-11
fix some problems in core.ps1 - set_config

jantari:
- `Out-Null` is slow, casting to `[void]` or assigning to `$null` is much faster
- Add parameter names instead of relying on position
- Do not allow null or empty string for `$name` - function would previously thow an error if that occured, and it doesn't make sense to allow for it
- if `$scoopConfig` was `$null` and `$value` was a boolean-string ("True"/"False") it would previously not have been converted to a bool-type and saved in the JSON as a string-value instead. Fixed by moving the bool-conversion logic outside of the unrelated if-statement

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,10 +64,13 @@ If you've built software that you'd like others to use, Scoop is an alternative 
 
 ## Installation
 
-Run this command from your PowerShell to install scoop to its default location (`C:\Users\<user>\scoop`)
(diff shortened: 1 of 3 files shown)


r15ch13:
Ah crap! Changing the branch to develop destroyed the PR. My bad 😢
Please create a new one, but base the changes on the develop branch.

PR #3681, opened 2019-10-11
fix(config): Fix `set_config` bugs

jantari:
Same as: https://github.com/lukesampson/scoop/pull/3680 but in develop branch -->

---

- `Out-Null` is slow, casting to `[void]` or assigning to `$null` is much faster
- Add parameter names instead of relying on position
- Do not allow null or empty string for `$name` - function would previously thow an error if that occured, and it doesn't make sense to allow for it
- if `$scoopConfig` was `$null` and `$value` was a boolean-string ("True"/"False") it would previously not have been converted to a bool-type and saved in the JSON as a string-value instead. Fixed by moving the bool-conversion logic outside of the unrelated if-statement


lib/core.ps1
```diff
@@ -54,27 +54,33 @@ function get_config($name, $default) {
...
-        $scoopConfig = New-Object PSObject
+function set_config {
+    Param (
+        [ValidateNotNullOrEmpty()]
+        $name,
+        $value
+    )
+
+    if ($null -eq $scoopConfig -or $scoopConfig.Count -eq 0) {
+        $null = ensure (Split-Path -Path $configFile)
```
r15ch13: This function doesn't get called often. A millisecond performance boost would not make any difference.
I would go for readability in this case and use `| Out-Null`

jantari: I personally find both equally readable and if we were to standardize the entire codebase on one method of discarding STDOUT choosing the most performant one could add up.

I can see how that should be a seperate issue/PR though, so for the time being I'll change it back.

niheaven (approved):

PR #3685, opened 2019-10-13
feature(update): allow updating apps from local manifest or URL

### LINKED ISSUE #3607: Update of manifests installed by url fails
Reproducible steps:

1. `scoop bucket rm Ash258` in case you already have this bucket added.
1. `scoop install 'https://raw.githubusercontent.com/Ash258/scoop-Ash258/master/bucket/RPCS3.json'`
    - RPCS3 is good example as it has frequent updates
1. Wait for update of manifest
1. `scoop update`
1. `scoop status`
    - Status should find out there was update of manifest
1. `scoop update RPCS3`
1. `Could not find manifest for RPCS3`

Same problem with Rider manifest

image

For debug purpose I suggest to create custom gist and just change version property.

### LINKED ISSUE #3683: Issue when using scoop update from a manifest url
**What:**
We can install an application from a local manifest (url), we should be able to update it from the same url

**Use case:**
1. Create an application manifest on your machine `C:/scoop_manifests/appA.json`
2. install an app from the local manifest on your machine:
`scoop install C:/scoop_manifests/appA.json`
3. Update the local manifest `C:/scoop_manifests/appA.json`
4. Update the app with scoop
`scoop update appA`

**Expected:**
This should update the app with your new manifest value

**Actual result:**
Scoop look in its buckets if the app exist, if so it update the app to its latest bucket version

**Investigation:**
in scoop-update.ps1 we don't concider if the manifest contains an url instead of an bucket name

stephanec1:
- Closes #3683 
- Closes #3607

### DIFF
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -173,6 +173,11 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
 
     if (!$independent) {
         # check dependencies
+        if ($install.url) {
+            $deps = @(deps $install.url $architecture) | Where-Object { !(installed $_) }
+        } else {
+            $deps = @(deps $app $architecture) | Where-Object { !(installed $_) }
+        }
         $deps = @(deps $app $architecture) | Where-Object { !(installed $_) }
         $deps | ForEach-Object { install_app $_ $architecture $global $suggested $use_cache $check_hash }
     }
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -178,7 +178,8 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
         } else {
             $deps = @(deps $app $architecture) | Where-Object { !(installed $_) }
         }
-        $deps = @(deps $app $architecture) | Where-Object { !(installed $_) }
+		$man = if ($url) { $url } else { $app }
+        $deps = @(deps $man $architecture) | Where-Object { !(installed $_) }
         $deps | ForEach-Object { install_app $_ $architecture $global $suggested $use_cache $check_hash }
     }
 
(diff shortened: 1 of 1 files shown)


Ash258:
This is not enough. You need to cover dependcy resolver too.

Update will fail on function `update` on `$(deps $app $architecture)`

stephanec1:
moreover its not an elseif but an else, we always set the bucket name to "main" if it doesn't exist

stephanec1:
Dependencies do not support url manifest for updates. 
The code just check if there is an app with the same name than the dependency. But if the dependency is a manifest.json, if didn't find anything

libexec/scoop-update.ps1
```diff
@@ -173,6 +173,11 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
 
     if (!$independent) {
         # check dependencies
+        if ($install.url) {
+            $deps = @(deps $install.url $architecture) | Where-Object { !(installed $_) }
+        } else {
+            $deps = @(deps $app $architecture) | Where-Object { !(installed $_) }
+        }
         $deps = @(deps $app $architecture) | Where-Object { !(installed $_) }
```
Ash258:
```suggestion
		$man = if ($url) { $url } else { $app }
        $deps = @(deps $man $architecture) | Where-Object { !(installed $_) }
```

libexec/scoop-update.ps1
```diff
@@ -173,7 +173,13 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
 
     if (!$independent) {
         # check dependencies
-        $deps = @(deps $app $architecture) | Where-Object { !(installed $_) }
+        if ($install.url) {
+            $deps = @(deps $install.url $architecture) | Where-Object { !(installed $_) }
+        } else {
+            $deps = @(deps $app $architecture) | Where-Object { !(installed $_) }
+        }
```
Ash258: Remove these obsolete lines.

stephanec1: oups sorry

Ash258:
Add snippet bellow as PR description

```txt
- Closes #3683 
- Closes #3607
```

PR #3690, opened 2019-10-17
fix(core): Use [Environment]::Is64BitOperatingSystem instead of [intptr]::size

linsui:
[intptr]::size is 4 in powershell x86 on 64bit system.


r15ch13:
Related: https://github.com/lukesampson/scoop/issues/3105 https://github.com/lukesampson/scoop/issues/2114

Issue #2952, opened 2019-01-05
Refactor uninstallation


Ash258:
- Part of #488

1. [ ] Remove similar uninstallation code
	- `libexec/uninstall.ps1` and `bin/uninstall.ps1` use similar/duplicated code for uninstallation
		- https://github.com/lukesampson/scoop/blob/99b5afafa086246aac65d63160218c38b014650d/bin/uninstall.ps1#L18-L44
		- https://github.com/lukesampson/scoop/blob/99b5afafa086246aac65d63160218c38b014650d/libexec/scoop-uninstall.ps1#L42-L80
	- Move adequate uninstallation functions into lib/uninstall.ps1
	- Duplicated function:
		- Scoop-UninstallApplication
			- params:
				- [String] $AppName
					- Maybe default to *
				- [Switch] $Global
			- return:
				- Boolean if some error occured
1. [ ] Implement uninstallation of running applications
	- function Scoop-GetProcess($appname, $global) into core/install/uninstall
		- Return whole Get-Process
		```powershell
		<#
		.SYNOPSIS
			Get all running processes for given app.
		.PARAMETER AppName
			Name of the application.
			Default to return all running applications. (Mainly for uninstallatino of whole scoop)
		.PARAMETER Global
			Get process for globally installed applications.
		#>
		function Scoop-GetProcess {
			param(
				[String] $AppName = '*',
				[Switch] $Global
			)

			$procs = Get-Process # Avoid calling twice for each path
			if ($Global) {
				return $procs | Where-Object { $_.Path -like "$globaldir\apps\$AppName\*" }
			} else {
				return $procs | Where-Object { $_.Path -like "$scoopdir\apps\$AppName\*" }
			}
		}
		```
	- Adopt libexecs for Get-Process
		- Uninstall
		- Update
1. [ ] ❓ Add force parameter to uninstall for killing running applications

TheRandomLabs:
Not sure if this is the right place for this, but would it be possible to add an option to run the uninstaller script regardless of whether the application is still running? Corsair iCUE's uninstaller automatically kills all of its processes before uninstalling, but it's unnecessarily difficult to have to kill all Corsair iCUE processes before uninstalling as they also include services.

PR #3692, opened 2019-10-17
fix(update): support $persist_dir in uninstaller.script

linsui:
$persist_dir in uninstaller.script only works when uninstall but not when update.


Ash258:
I see what you mean. Yeath this could go in for small period of time.

I am currently working on #2952, which will cover this use case.

PR #3648, opened 2019-09-19
Revert 'Compare-Version' (#3333, #3578)

niheaven:

PR #3694, opened 2019-10-18
fix(install): Allow installing specific version if latest is installed

### LINKED ISSUE #3693: How do I update an app that I installed with a specific version (without uninstalling and reinstalling)?
Here's a specific case:

Let's say I installed terraform, version 0.11.14 because that's what the team I work with is using now:
`scoop install terraform@0.11.14`

That creates an appropriate manifest in my scoop workspace and installs the correct version. All fine and dandy.

The team decides to use the newest version of terraform (0.12.11) for a new project, but not to upgrade the version used by an old project because... reasons.

What I was expecting to be able to do:
`scoop update terraform` to update terraform, but keep the old version in its own version folder;
then use `scoop reset terraform@0.11.14` when working on the old project, and `scoop reset terraform@0.12.11` when working on the new one.

But that doesn't work. And if I try to `scoop install terraform@0.12.11`, or `scoop update terraform -f`, I get a message that the latest version is already installed (because it's relying on the generated manifest for 0.11.14 instead of the one in the bucket)
So, I am stuck with uninstalling and reinstalling the version each time I swap to a different project.

Is there something I am missing that would make the process easier? The wiki is not very exhaustive about scoop's commands.

r15ch13:
closes #3693
Should be hotfixed to `master`

# Expected behavior
> scoop install terraform

Installs version 0.12.11
> scoop install terraform@0.11.14

Installs version 0.11.14
> scoop reset terraform

Resets to version 0.12.11 and it should be updateable

# Actual behavior
image

# Fixed behavior
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,10 +64,13 @@ If you've built software that you'd like others to use, Scoop is an alternative 
 
 ## Installation
 
-Run this command from your PowerShell to install scoop to its default location (`C:\Users\<user>\scoop`)
+Run the following command from your PowerShell to install scoop to its default location (`C:\Users\<user>\scoop`)
 
 ```powershell
-iex (new-object net.webclient).downloadstring('https://get.scoop.sh')
+Invoke-Expression (New-Object System.Net.WebClient).DownloadString('https://get.scoop.sh')
+
+# or shorter
+iwr -useb get.scoop.sh | iex
 ```
 
 Once installed, run `scoop help` for instructions.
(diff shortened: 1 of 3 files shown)


Ash258:
> Should be hotfixed to master

Not needed. Just merge this to develop. Then merge #3648 and sync master.
Master is behind for too long. and only compare versions refactor is blocking it to go live.

r15ch13:
... killed PR :grin:

PR #3697, opened 2019-10-18
Remove unnecessary git_proxy_cmd() calls for local commands

r15ch13:


Ash258:
I would rather avoid invoke-expression due to security reason in favor of custom Invoke-External command or custom git wrapper.

Pretty old post regarding Invoke-Expression https://devblogs.microsoft.com/powershell/invoke-expression-considered-harmful/

Cannot find updated one right now.

r15ch13:
Hm, aria2 gets called via `iex`, but it also uses the `&` call operator.

https://github.com/lukesampson/scoop/blob/08af9ff6e7f7b017701c0c9114294b13e1e83fb8/lib/install.ps1#L281-L286

PR #3730, opened 2019-11-05
fix bucket check

linsui:
- Abort immediately when bucket is specified but there is no such a bucket.
- Do not check manifest from local path when bucket is specified.

### DIFF
--- a/lib/depends.ps1
+++ b/lib/depends.ps1
@@ -21,13 +21,13 @@ function deps($app, $arch) {
 
 function dep_resolve($app, $arch, $resolved, $unresolved) {
     $app, $bucket, $null = parse_app $app
+    if(((Get-LocalBucket) -notcontains $bucket) -and $bucket) {
+        abort "Bucket '$bucket' not installed. Add it with 'scoop bucket add $bucket' or 'scoop bucket add $bucket <repo>'."
+    }
     $unresolved += $app
     $null, $manifest, $null, $null = Find-Manifest $app $bucket
 
     if(!$manifest) {
-        if(((Get-LocalBucket) -notcontains $bucket) -and $bucket) {
-            warn "Bucket '$bucket' not installed. Add it with 'scoop bucket add $bucket' or 'scoop bucket add $bucket <repo>'."
-        }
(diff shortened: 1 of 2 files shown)


lib/depends.ps1
```diff
@@ -21,13 +21,13 @@ function deps($app, $arch) {
 
 function dep_resolve($app, $arch, $resolved, $unresolved) {
     $app, $bucket, $null = parse_app $app
+    if(((Get-LocalBucket) -notcontains $bucket) -and $bucket) {
```
Ash258: Excahange these two conditions.
In case there is no $bucket is not specified, then it will fail immediatelly and do not check the second condition. (Short circuiting)

```powershell
function testBucket($buck) { Write-Host $buck; return $true }
```

image

See image above. Function is not evaluated in case first condition is false. And in second you can see `ahoj` message popuped from evaluation

```suggestion
    if ($bucket -and ((Get-LocalBucket) -notcontains $bucket)) {
```

lib/install.ps1
```diff
@@ -86,9 +86,13 @@ function Find-Manifest($app, $bucket) {
         $manifest = url_manifest $url
     } else {
         # check buckets
-        $manifest, $bucket = find_manifest $app $bucket
+        if (!$bucket) {
+            $manifest, $bucket = find_manifest $app $bucket
```
Ash258: I do not understand. Why are you passing $bucket when it is $null or do not exists

```suggestion
            $manifest, $bucket = find_manifest $app
```

lib/install.ps1
```diff
@@ -86,9 +86,13 @@ function Find-Manifest($app, $bucket) {
         $manifest = url_manifest $url
     } else {
         # check buckets
-        $manifest, $bucket = find_manifest $app $bucket
+        if (!$bucket) {
```
Ash258: In general, when you are using both branches of `if` statement, then it is not suitable to negate the result.

```suggestion
        if ($bucket) {
            $manifest, $null = find_manifest $app $bucket
        } else {
            $manifest, $bucket = find_manifest $app
        }
```

PR #3740, opened 2019-11-09
fix(scoop-info): check bucket of installed app

linsui:
When bucket is specified, check if the installed app belong to the same bucket. If app is not installed from the specified bucket, output `Installed: No`

Also fix another bug. When an app hash is wrong there is no manifest.json in app directory. So `$install.bucket` is none. `Find-BucketDirectory` will return main bucket path as default and `manifest_path` will return a wrong path.

### DIFF
--- a/libexec/scoop-info.ps1
+++ b/libexec/scoop-info.ps1
@@ -35,7 +35,7 @@ if (!$manifest) {
 }
 
 $install = install_info $app $status.version $global
-if ($install.bucket -ne $bucket) {
+$status.installed = $install.bucket -eq $bucket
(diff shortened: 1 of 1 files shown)


libexec/scoop-info.ps1
```diff
@@ -35,6 +35,9 @@ if (!$manifest) {
 }
 
 $install = install_info $app $status.version $global
+if ($install.bucket -ne $bucket) {
```
Ash258: Would not make sense to go with:?

```suggestion
$status.installed = $install.bucket -eq $bucket
```

PR #3751, opened 2019-11-18
support symlink in movedir

linsui:
Fix https://github.com/lukesampson/scoop/issues/3750
See https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/robocopy


Ash258:
Is this something you want to? Could it lead into excessive duplication of files and hogging up too much space?

linsui:
This is a corner case. I just want to fix the issue and for archive without symlink there should be no difference. `movedir` is only used in decompress and some manifests, I don't know why not use `Move-Item` directly which is more robust.

PR #3763, opened 2019-11-22
fix(decompress): Remove parent dir when extract_dir is nested dir

linsui:
When extract_dir is nested dir, the parent dir won't be removed by movedir.

Test with `kubectl` and `wire` .

### DIFF
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -65,11 +65,15 @@ function Expand-7zipArchive {
         $7zPath = Get-HelperPath -Helper 7zip
     }
     $LogPath = "$(Split-Path $Path)\7zip.log"
-    $ArgList = @('x', "`"$Path`"", "-o`"$DestinationPath`"", '-y')
     $IsTar = ((strip_ext $Path) -match '\.tar$') -or ($Path -match '\.t[abgpx]z2?$')
+    $DestinationPath = $DestinationPath.TrimEnd("\")
+    $ArgList = @('x', "`"$Path`"", '-y')
     if (!$IsTar -and $ExtractDir) {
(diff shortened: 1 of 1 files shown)


lib/decompress.ps1
```diff
@@ -65,11 +65,15 @@ function Expand-7zipArchive {
         $7zPath = Get-HelperPath -Helper 7zip
     }
     $LogPath = "$(Split-Path $Path)\7zip.log"
-    $ArgList = @('x', "`"$Path`"", "-o`"$DestinationPath`"", '-y')
     $IsTar = ((strip_ext $Path) -match '\.tar$') -or ($Path -match '\.t[abgpx]z2?$')
+    $DestinationPath = $DestinationPath.TrimEnd("\")
+    $ArgList = @('x', "`"$Path`"", '-y')
     if (!$IsTar -and $ExtractDir) {
+        $OriDestinationPath = $DestinationPath
+        $DestinationPath = "$DestinationPath\_tmp"
```
niheaven: You could use `Join-Path` to avoid using `TrimEnd("\")`.

linsui: I just keep it consistent with other exppander. Should I change all of them?

niheaven: Could you try if the script works without `TrimEnd("\")`? i.e. if 7-zip recognize `aaa\\temp`. I trim `\` in msi's since https://github.com/lukesampson/scoop/pull/3466#issuecomment-492989875, and `Expand-Archive` use `aaa\\_tmp` as `aaa\_tmp` so it don't need trim.

linsui: It works. I got, I'll remove it. Thanks!

PR #3778, opened 2019-12-04
feat: Add configuration option for default architecture

### LINKED ISSUE #3770: How to configure scoop to default to 32bit architecture?
Currently scoop only suports 64bit and 32bit architectures and there are many Surface Pro X users who'd like to use this tool to help setup their machines (even if running emulated win32 apps instead of native). It would be good if there was a way to configure scoop to default to 32bit apps because always typing `--arch 32bit` gets tiresome really fast.

Ash258:
- Closes #3770

Configuration name is subject of change. I am following dash notation for naming, until config is reworked and united (Related #3401)

~~Is boolean enough or it is prefered to go with `scoop config 'default-architecture' '32bit|64bit'` which could be more suitable for additional arch support (arm for example)~~

image

### DIFF
--- a/lib/manifest.ps1
+++ b/lib/manifest.ps1
@@ -59,8 +59,15 @@ function install_info($app, $version, $global) {
 }
 
 function default_architecture {
-    if ([Environment]::Is64BitOperatingSystem) { return "64bit" }
-    "32bit"
+    if (get_config 'prefer-32bit' $false) {
+        $arch = '32bit'
+    } elseif ([Environment]::Is64BitOperatingSystem) {
+        $arch = '64bit'
+    } else {
+        $arch = '32bit'
+    }
+
+    return $arch
 }
 
 function arch_specific($prop, $manifest, $architecture) {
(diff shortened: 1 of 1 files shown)


niheaven:
`default-architecture` is preferred for me.

r15ch13:
`default-architecture` with `32bit|64bit|arm` sounds good to me. 👍

niheaven:
scoop should throw when user use some words other than `32bit` or `64bit`.

Ash258:
I could theoretically add else branch with [ensure_architecture function](https://github.com/lukesampson/scoop/blob/08af9ff6e7f7b017701c0c9114294b13e1e83fb8/lib/core.ps1#L631) check

Resulting to:

```powershell
   if ($null -eq $arch) {
		if .... is64..
   } else {
       $arch = ensure_architecture $arch
   }
```

or to be more user friendly with something like this

```powershell
try {
   $arch = ensure_architecture $arch
} catch {
	error 'Invalid default architecture configured'
}
```

niheaven:
Prefer `try...cache` :smile:

PR #3782, opened 2019-12-05
feat: Present download command

### LINKED ISSUE #1804: Download command proposal
It would be nice if I can just download the installer file and run it by myself without adding it to scoop list.

 i. e. Browsers have autoupdate system that updates for me, so when i type `scoop status` it shows that my browser is outdated, but it is not.

Or just silent install it without adding to installed packages list. (maybe some option to install command)

### LINKED ISSUE #2293: Add command to download and cache without installing
Might be useful for situations where you don't want to install the software immediately but you want the installer handy for later.

Ash258:
- Closes #1804
- Closes #2293

New command for only downloading of files defined in manifest.

UNC paths, url, specific versions are supported

A

B

### DIFF
--- a/libexec/scoop-download.ps1
+++ b/libexec/scoop-download.ps1
@@ -5,11 +5,11 @@
 #
 # Options:
 #   -s, --skip                      Skip hash check validation.
-#   -a, --arch <32bit|64bit>        Use the specified architecture.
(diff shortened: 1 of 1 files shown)


niheaven:
Why not just add trigger `-DownloadOnly` in `scoop install` and `scoop update`?

PR #3784, opened 2019-12-07
fix(install): fix wrong output of 'install/failed'

niheaven:
Fix https://github.com/ScoopInstaller/Main/issues/642

This was caused by https://github.com/lukesampson/scoop/blob/cb2ff446e6a3b5367db7ed85c284ac083fd5f450/lib/install.ps1#L603, should be `return $true`... Replace the `failed` function with single line.

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -1021,11 +1021,7 @@ function prune_installed($apps, $global) {
 
 # check whether the app failed to install
 function failed($app, $global) {
-    $ver = current_version $app $global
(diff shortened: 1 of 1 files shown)


h404bi:
So is this fix ready to merge?
ping: 
https://github.com/ScoopInstaller/Main/issues/824
https://github.com/ScoopInstaller/Main/issues/786

niheaven:
Yes, ping @r15ch13

PR #3788, opened 2019-12-10
fix(install): let pathes in 'env_add_path' be added ascendantly

niheaven:
Fix https://github.com/lukesampson/scoop/issues/3785

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -939,7 +939,7 @@ function find_dir_or_subdir($path, $dir) {
 function env_add_path($manifest, $dir, $global, $arch) {
     $env_add_path = arch_specific 'env_add_path' $manifest $arch
     # GH-3785: Add path in ascending order.
-    $env_add_path = $env_add_path[-1..-$env_add_path.Length]
+    [Array]::Reverse($env_add_path)
     $env_add_path | Where-Object { $_ } | ForEach-Object {
         $path_dir = Join-Path $dir $_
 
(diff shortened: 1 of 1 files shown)


lib/install.ps1
```diff
@@ -938,6 +938,7 @@ function find_dir_or_subdir($path, $dir) {
 
 function env_add_path($manifest, $dir, $global, $arch) {
     $env_add_path = arch_specific 'env_add_path' $manifest $arch
+    $env_add_path = $env_add_path[-1..-$env_add_path.Length]
```
h404bi:
```suggestion
    # GH-3785: Add path in ascending order.
    $env_add_path = $env_add_path[-1..-$env_add_path.Length]
```

lib/install.ps1
```diff
@@ -938,6 +938,8 @@ function find_dir_or_subdir($path, $dir) {
 
 function env_add_path($manifest, $dir, $global, $arch) {
     $env_add_path = arch_specific 'env_add_path' $manifest $arch
+    # GH-3785: Add path in ascending order.
+    $env_add_path = $env_add_path[-1..-$env_add_path.Length]
```
Ash258: Would not be more cleaner to use `[Array]::Reverse($env_add_path)`?

Ash258:
Is there any possibility of breaking current behaviour? I cannot think about anything right now.

niheaven:
Just order in %PATH%, which shouldn't break anything. Maintainer typically write `env_add_path` ascendantly, I thought.

h404bi:
ping @r15ch13

PR #3780, opened 2019-12-04
feat(config): Add all configuration options into help

Ash258:
These are all config options, which I found in code base.

`NO_JUNCTIONS` Anyone have clean description?

PR #3819, opened 2020-01-10
fix(install): add config to skip aria2 cert check

copdips:
When using aria2 to download the packages, and if the target certificate check is failed for x reason (for ex: `80092013` as per https://github.com/aria2/aria2/issues/1362), aria2 exit code will be `1` which is marked as unknown code as per: https://github.com/lukesampson/scoop/blob/a9fa775d59b14e7dce335313faa0eff855469764/lib/install.ps1#L137

In such case, we can provide a possibility to let aria2 to ignore the certificate check by `--check-certificate=false`

It seems that the list of settings that can be set by `scoop config` is not well documented, users might not aware how to ignore the aria2 certificate check, that's why in the `if (Test-Aria2SkipCertificateCheck) {` block, I added two `Write-Host` to guide users to set the `aria2-skipCertificateCheck`.

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -286,6 +286,10 @@ function Test-Aria2Enabled {
     return (Test-HelperInstalled -Helper Aria2) -and (get_config 'aria2-enabled' $true)
 }
 
+function Test-Aria2SkipCertificateCheck {
+    return (Test-HelperInstalled -Helper Aria2) -and (get_config 'aria2-skipCertificateCheck' $false)
+}
+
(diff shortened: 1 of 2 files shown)


copdips:
@Ash258 
Any reason for the thumbs down ?

Ash258:
Adding more bloat, which could be easily done with config now and which will be obsolete after #3780.

image

image

image

copdips:
great, thx for the info, I close the PR.

PR #3896, opened 2020-03-25
Two ways to set encoding and cookie when checkver

L-Trump:
use parameter Cookies,Encode or $json.checkver.cookie&$json.checkver.encode


Ash258:
What are you trying to do with all the PRs opened??

Why would you need separate user agent passing when checkver already support user agent?

PR #3910, opened 2020-04-02
Using hard link instead of copy shim.exe every time creating shim

kiennq:
We should use hard link instead of copying `shim.exe` every time. This will greatly reduce disk storage when you have multiple shim link.

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -518,6 +518,35 @@ function movedir($from, $to) {
     }
 }
 
+function rm_file($path) {
+    try {
+        Remove-Item $path -ErrorAction Stop
+    }
+    catch {
+        if (Test-Path -Path $path -PathType Leaf) {
+            # Move to temp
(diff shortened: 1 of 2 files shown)


kiennq:
@Ash258 Not sure how I get thumb down. Can you help explain why?

Deide:
I would also like to explore the advantages vs disadvantages with this.

I personally think the extra storage is negligible for _me_, but I also don't know of a downside to hard linking instead - maybe there's some unexpected behaviour?

kiennq:
I think I can list a few:
Advantages:
- Storage footprint
- When `scoop` app get update with new `shim.exe` (may a better c/c++ version), all hard linked shim will also be updated instantly.

But that also come with disadvantages:
- Since hard linked file cannot be deleted when in used, we can only move it to temporary folder and then make it removed on reboot. Scoop command has been updated to accommodate update/install/uninstall with hard link already.

rashil2000:
This PR adds unnecessary complexity by calling native functions. The storage space gained by this is negligible by today's standards.

PR #3946, opened 2020-04-26
Update scoop-alias.ps1

archit9169:
Add include commands.ps1 to get the function command_path to work properly. Otherwise it gives error on scoop alias list


Ash258:
What error?

Commands are included in main scoop binary

archit9169:
> What error?
> 
> Commands are included in main scoop binary

$content = Get-Content (command_path $_.Name)
this line was giving me error. Was resolved when i added commands.ps1 in include

this command_path is present in commands.ps1 file which is not defined in core.ps1, install.ps1, help.ps1

PR #3961, opened 2020-05-05
core(install): Readd "Don't send referer to portableapps.com"

NilsonPark:
Close https://github.com/lukesampson/scoop/issues/3960
Do not send referer to portableapps.com.
There are still many software hosted in https://download3.portableapps.com


NilsonPark:
I have no idea how to fix my commit to pass the failed checks. Help needed. @r15ch13, you are most familiar with this part, please jump in at your convenience.

PR #4006, opened 2020-05-31
If scoop update fails, scoop itself will not be removed now (fixes #4005)

### LINKED ISSUE #4005: [BUG] If 'scoop update' failed, scoop itself will be removed.
## Description
Scoop is installed with 'bin/install.ps1' ('\~/scoop/apps/scoop/current/.git' doesn't exist), if I run 'scoop update' and use 'Ctrl + C' to interrupt it, '\~/scoop/apps/scoop/current' will be removed silently.

## Solution
There is a bug in 'libexec/scoop-update.ps1', when '$currentdir/.git' doesn't exist and 'git clone' fails, the currentdir will always be removed. I fixed it in my pull request.

GalaxySnail:
Fixes #4005: If 'scoop update' failed, scoop itself will be removed.

### DIFF
--- a/lib/buckets.ps1
+++ b/lib/buckets.ps1
@@ -101,7 +101,7 @@ function add_bucket($name, $repo) {
     write-host 'Checking repo... ' -nonewline
     $out = git_ls_remote $repo 2>&1
     if ($lastexitcode -ne 0) {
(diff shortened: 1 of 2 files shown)


Romyanmar4:
gh PR ကုန်ပစ္စည်း 4006

PR #4071, opened 2020-07-24
[WIP] feature: Add support for ARM64 architecture

ktos:
I am not 100% sure if it should be done that way, but I've tried myself in adding basic support for ARM64 architecture, apart from "32bit" and "64bit" scoop has already. The discussion was started in https://github.com/lukesampson/scoop/issues/3146.

It requires of course changes to the manifest - apart from two already existing architectures, third one was added, with "arm64" moniker (while "aarch64" and "armv8" are also supported in parameter list).

To recognize if we are running on Windows on ARM (WoA), the `$env:PROCESSOR_ARCHITECTURE` is used, as [WOW64 documentation](https://docs.microsoft.com/en-us/windows/win32/winprog64/wow64-implementation-details?redirectedfrom=MSDN) suggests.

I have added warnings when trying to install ARM64 application on non-ARM system (but there should be nothing to stop the user at all - it's their decision) and when there is no architecture given in manifest - on ARM64 system it will require explicit `scoop install -a 32bit foobar` for example.

I am marking this as WIP at the moment, as checkhashes/checkurls.ps1 has not been yet modified to work with ARM64.

What do you think?

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -34,6 +34,10 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
         return
     }
 
+    if ((default_architecture -neq "arm64") -and ($architecture -eq "arm64")) {
+        warn "ARM64 application probably won't run on this system."
+    }
+
     write-output "Installing '$app' ($version) [$architecture]"
 
     $dir = ensure (versiondir $app $version $global)
--- a/bin/checkurls.ps1
+++ b/bin/checkurls.ps1
@@ -89,6 +89,7 @@ foreach ($man in $Queue) {
     if ($manifest.url) {
         $manifest.url | ForEach-Object { $urls += $_ }
     } else {
+        url $manifest 'arm64' | ForEach-Object { $urls += $_ }
         url $manifest '64bit' | ForEach-Object { $urls += $_ }
         url $manifest '32bit' | ForEach-Object { $urls += $_ }
     }
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -635,6 +635,7 @@ function ensure_architecture($architecture_opt) {
     $architecture_opt = $architecture_opt.ToString().ToLower()
     switch($architecture_opt) {
         { @('64bit', '64', 'x64', 'amd64', 'x86_64', 'x86-64')  -contains $_ } { return '64bit' }
+        { @('arm64', 'aarch64', 'armv8')  -contains $_ } { return 'arm64' }
         { @('32bit', '32', 'x86', 'i386', '386', 'i686')  -contains $_ } { return '32bit' }
         default { throw [System.ArgumentException] "Invalid architecture: '$architecture_opt'"}
     }
--- a/lib/manifest.ps1
+++ b/lib/manifest.ps1
@@ -58,9 +58,14 @@ function install_info($app, $version, $global) {
     parse_json $path
 }
 
+function system_architecture {
+    if ($env:PROCESSOR_ARCHITECTURE -eq "ARM64") { return "arm64" }
+    if ([Environment]::Is64BitOperatingSystem) { return "64bit" } else { return "32bit" }
+}
+
 function default_architecture {
     $arch = get_config 'default-architecture'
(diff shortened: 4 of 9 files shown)


ktos:
Okay, I have checked this code today on the "real" Windows 10 on ARM devices, and it seems I was mistaken - PowerShell 5 is running in x86 emulation mode, thus `PROCESSOR_ARCHITECTURE` is `x86`. I have no idea how to proceed with this.

I have tested on PSCore 7.0.3 ARM64 build and now it works, properly recognized architecture, works with my custom manifests for ARM64 apps (like Notepad2), yet there are still some rough edges (testing dependencies, warning on installing ARM apps is shown on ARM platforms), and so on.

rasa:
If there isn’t an ARM architecture in the manifest, will it default to installing the 32bit architecture, if one exists in the manifest, (as ARM64 will run 32bit in emulated mode)?

ktos:
That's actually a thing to consider. At the moment I've implemented it such it won't install - just throws "<package> does not support 'arm64' architecture". You have to use explicit `scoop install --arch 32bit <package>` to install x86/32bit version.

lib/install.ps1
```diff
@@ -34,6 +34,10 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
         return
     }
 
+    if ((default_architecture -neq "arm64") -and ($architecture -eq "arm64")) {
```
zhzy0077, line 37: `(default_architecture) -ne "arm64")`  otherwise the "arm64" will be passed to the function as parameter.

zhzy0077:
I did a rough test on my ARM64 windows. I installed firefox and vscode. Both work for me:
```
        "arm64": {
            "url": "https://archive.mozilla.org/pub/firefox/releases/83.0/win64-aarch64/en-US/Firefox%20Setup%2083.0.exe#/dl.7z",
            "hash": "sha512:c5b16ee3b3bbd329c2e5440468aa0a61f79ad279200de3ac5c50ef586cac148bf5adf1e90e68d2c84baf147c1f586eb276eae89ea515663d8b02904e9f96e182"
        }
```
```
        "arm64": {
            "url": [
                "https://vscode-update.azurewebsites.net/1.51.1/win32-arm64-archive/stable#/dl.7z",
                "https://raw.githubusercontent.com/lukesampson/scoop-extras/master/scripts/vscode-install-context.reg",
                "https://raw.githubusercontent.com/lukesampson/scoop-extras/master/scripts/vscode-uninstall-context.reg"
            ],
            "hash": [
                "4387ba2dea41da8f2df921bc937d693d690bc66de290ee6fcf4645f056746627",
                "b65d66860d9ccc18bfb05237b03e06db0d6c574be3d4b946c2a5e6865c08cb28",
                "df2a5162e72e2518e2b75b4337c8e7b46c136554872af90fa0de1cf6ebef376f"
            ]
        }
```

Hopefully this will get some progress.

ktos:
I am sorry I left this one for such a long time without any progress (or even response to @zhzy0077). It's all because of some personal issues and lack of access to WoA device anymore.

It seems that apart from `PROCESSOR_ARCHITECTURE`, there is also `PROCESSOR_ARCHITEW6432`, showing "real" architecture, even if run in WOW mode, which solves a problem for us to distinguish between architectures. 

It seems that also [x64 emulation is planned](https://blogs.windows.com/windowsexperience/2020/09/30/now-more-essential-than-ever-the-role-of-the-windows-pc-has-changed/) -- but I do not know if is already working. If it is, we can now only decide what is the preferred architecture: I think it should be ARM64, x64 and x86, in that order. That means if there is no arm64 in the manifest, it will try to install x64.

zhzy0077:
@ktos What I'm concerning is the installation order. For example, there will never be 64bit GPU drivers for Qualcomm 835/850, which means apps using GPU can work well in x86 but badly in amd64. 

My suggestion is to give some notes when the ARM version is not found and let user to decide the fallback architecture.
(Or even integrate with https://github.com/nvbn/thefuck to make it faster.)

rashil2000:
Superceded by #5154

PR #4073, opened 2020-07-25
feature(shim): hide console window when started from gui.

RustamGamidov:
stdout is still visible for cmd tools but for gui apps started from
   the win+r dialog or file explorer there's no additional console
   window


RustamGamidov:
looks like 6df80c1aeef72a3528ae3a5e37faa3de840318a2 is missing in the develop branch

PR #4117, opened 2020-09-21
Added simple no proxy functionality

### LINKED ISSUE #1820: can proxy be applied conditionally based on the url?
I have found scoop awesome in maintaining software,  so much that now i even want to encourage my friends to use it. while making use of it behind proxy is there anything that  can help me to skip proxy for a list of urls and apply for the others, something like a white-list/black-list. can i configure proxy per bucket so that proxy doesn't apply for all app references inside the bucket. i have created custom manifests hosted in an internal repository but the scoop proxy settings create issues when i update this bucket or install apps from it. on the other hand if i remove proxy i'm not able to update/install from the main bucket.

zie-pao:
I tried to implement a simnple no proxy mechanism. It sets the no proxy environment variable for git requests and removed the proxy on installation of the url matches  (like) the no proxy host.

I am not aware of othere parts in the code where these changes have to be done but so far it works for me. This should also fix https://github.com/lukesampson/scoop/issues/1820

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -353,17 +353,18 @@ function dl_with_cache_aria2($app, $version, $manifest, $architecture, $dir, $co
 
 # download with filesize and progress indicator
 function dl($url, $to, $cookies, $progress) {
+    $no_proxy = get_config 'no_proxy'
+
     $reqUrl = ($url -split "#")[0]
     $wreq = [net.webrequest]::create($reqUrl)
-    if($wreq -is [net.httpwebrequest]) {
-        $wreq.useragent = Get-UserAgent
-        if (-not ($url -imatch "sourceforge\.net" -or $url -imatch "portableapps\.com")) {
-            $wreq.referer = strip_filename $url
-        }
-        if($cookies) {
-            $wreq.headers.add('Cookie', (cookie_header $cookies))
+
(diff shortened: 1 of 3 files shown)


lib/install.ps1
```diff
@@ -353,18 +353,19 @@ function dl_with_cache_aria2($app, $version, $manifest, $architecture, $dir, $co
...
-    if($wreq -is [net.httpwebrequest]) {
-        $wreq.useragent = Get-UserAgent
-        if (-not ($url -imatch "sourceforge\.net" -or $url -imatch "portableapps\.com")) {
-            $wreq.referer = strip_filename $url
-        }
-        if($cookies) {
-            $wreq.headers.add('Cookie', (cookie_header $cookies))
+
+    if($no_proxy) {
+        if($reqUrl -like "*$no_proxy*") {
```
c33s, line 362: i think the matching should be improved. having the url `example.org` and `NO_PROXY`= `192.168.0.1,localhost,www.example.org` would lead to a match but should not match. also `NO_PROXY` should also support wildcards `*,example.org`

philippe-granet:
A new updated PR to fix this problem: https://github.com/ScoopInstaller/Scoop/pull/6550

PR #4121, opened 2020-09-24
gh-pages: move installation guide to top

MatthewCroughan:
This is a small change, but I think it would make a big difference. Most websites like this tend to have the installation instructions immediately visible to the user's eyes, because people today have short attention spans and are quick to enter the first thing they see into a terminal. All your fave CLI apps have web pages that tell you how to install them immediately, but for scoop you have to scroll down. So I think it would be much better like this. Thanks so much for this amazing package manager that keeps me sane.

### DIFF
--- a/index.md
+++ b/index.md
@@ -2,6 +2,25 @@
 layout: default
 ---
 
+# Installs in seconds
+
+Make sure [PowerShell 5](https://aka.ms/wmf5download) (or later, include [PowerShell Core](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-windows?view=powershell-6))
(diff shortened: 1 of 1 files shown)


rashil2000:
Hey @MatthewCroughan, can you enable GH Pages deployment for your PR's branch? 

This will enable reviewers to preview your changes live.

MatthewCroughan:
@rashil2000 I believe I have done this correctly. But the CSS is not visible on the page. https://matthewcroughan.github.io/scoop/

rashil2000:
image


Here is a web capture of the PR.

I think the Installation section should be moved below the intro (above the video), as @dwettstein said.

rasa (approved):

Valinor (approved):

PR #4130, opened 2020-10-06
Fix VirusTotal results querying: switch to v3 API and offer links to manually check when things fail

pcrama:
virustotal started reporting 429 HTTP errors (related to captchas)

Instead of just reporting an error, provide the URL so that the user may easily check for themselves (especially if their terminal allows clicking URLs).

To avoid the 429 HTTP errors, I also upgraded the querying of the VirusTotal to use their v3 REST API when the API key is configured.  This is much more reliable than the old way.

### DIFF
--- a/libexec/scoop-virustotal.ps1
+++ b/libexec/scoop-virustotal.ps1
@@ -115,7 +115,7 @@ Function Get-VirusTotalResult($hash, $app) {
         }
     }
     $stats = json_path $result $stats_json_path
-    if ($stats -eq $null) {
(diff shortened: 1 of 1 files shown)


pcrama:
I am aware that the AppVeyor checks failed, but I don't see the file I have edited `libexec\scoop-virustotal.ps1` in the error messages, so I don't think I caused the errors with this patch, nor that fixing the remaining warnings would belong in this pull request.

rashil2000:
Superceded by #4983

Issue #3990, opened 2020-05-19
Support zstd for extraction


trajano:
Msys2 repo uses zstd for their new packages, can you add support for zstd decompression?

Calinou:
Scoop depends on 7-zip to extract formats other than ZIP, but the official 7-zip distribution doesn't support Zstandard. Many third-party forks do, though.

r15ch13:
Scoop can use 7zip-zstd.
```
scoop bucket add versions
scoop install 7zip-zstd
scoop uninstall 7zip
```

https://github.com/lukesampson/scoop/blob/96de9c14bb483f9278e4b0a9e22b1923ee752901/lib/core.ps1#L252-L258

Augiiz:
Still no luck extracting zst packages upon installation, they just sit there.

r15ch13:
Do you have some example files?

Augiiz:
![](https://i.imgur.com/oLbUwAa.png)

Since main buckets gcc still on 9.3 trying to set my own bucket for gcc 10, but mingw moved some of their archives to zst starting gcc 10.

niheaven:
Need some manual tweaks:

1. Add `zst` to L19 of `lib/decompress.ps1` of scoop (`%SCOOP%/apps/scoop/current/lib/decompress.ps1`)
1. `scoop reset 7zip-zstd`
1. Install your gcc10

Augiiz:
Did it for me, thanks a bunch.

trajano:
> Scoop depends on 7-zip to extract formats other than ZIP, but the official 7-zip distribution doesn't support Zstandard. Many third-party forks do, though.

Is there a PR on the 7Zip mainline to integrate it?

Calinou:
@trajano I think there have been several, but none of them was merged so far: https://sourceforge.net/p/sevenzip/discussion/45797/thread/a7e4f3f3/

trajano:
@Calinou from the discussions you linked it seems that there's no further work on integrating it and it had started in 2016.  Maybe scoop should consider an alternative to 7z or at least use the official distribution to do the decompression at least it would decouple scoop from 7zip in case there are new compression formats that it does not support.

wstein:
msys2 has purged many older file version.
Now many packages of msys2 are only available as zstd compressed files.

If this issue will not be fixed, the msys2 based packages like gcc and openssh should be dropped.

psydvl:
There are also `zstd` package in `main` bucket
[...]

PR #4133, opened 2020-10-11
add .zst to extension list (part to solve #3990)

wstein:
Please integrate this PR as 1st step to make the workaround for #3990 some easier.
So only command line instructions are necessary and not patching a source file.

Workaround on my installation (in addition to the patch):
```
scoop bucket add versions
scoop install 7zip-zstd
scoop uninstall 7zip
scoop reset 7zip-zstd
```

Needed to fix openssh package: ScoopInstaller/Main#1494


wstein:
I will create a new pull request which solves #3990

PR #4155, opened 2020-10-28
Show manifest on installation

MrNuggelz:
Closes https://github.com/lukesampson/scoop/issues/2735

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -9,6 +9,10 @@ function nightly_version($date, $quiet = $false) {
     "nightly-$date_str"
 }
 
+function Test-ManifestReviewEnabled {
+    return get_config 'manifest-review' $false
+}
+
 function install_app($app, $architecture, $global, $suggested, $use_cache = $true, $check_hash = $true) {
     $app, $bucket, $null = parse_app $app
     $app, $manifest, $bucket, $url = Find-Manifest $app $bucket
@@ -34,7 +38,18 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
         return
     }
 
-    write-output "Installing '$app' ($version) [$architecture]"
+    if (Test-ManifestReviewEnabled) {
+        $show_manifest = read-host -Prompt "Show manifest? [y/N]"
+        if ($show_manifest -eq "y") {
+            write-output "manifest:"
+            write-output $manifest | ConvertToPrettyJson
+            $answer = read-host -Prompt "Continue installation? [Y/n]"
+
+            if ($answer -eq "n") {
+                return
+            }
+        }
+    }
 
     $dir = ensure (versiondir $app $version $global)
     $original_dir = $dir # keep reference to real (not linked) directory
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -9,10 +9,6 @@ function nightly_version($date, $quiet = $false) {
     "nightly-$date_str"
 }
 
-function Test-ManifestReviewEnabled {
-    return get_config 'manifest-review' $false
-}
-
 function install_app($app, $architecture, $global, $suggested, $use_cache = $true, $check_hash = $true) {
     $app, $bucket, $null = parse_app $app
     $app, $manifest, $bucket, $url = Find-Manifest $app $bucket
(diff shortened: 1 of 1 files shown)


rashil2000:
I think a better and more granular approach would be implementing something like #4272. This would allow users to view manifests even when they aren't installing an app.

MrNuggelz:
The use case for this pull request is so you can review the manifest before updating an app, similar to what many AUR helpers.
This is especially usefull for third party buckets, were manifests are less monitored.
I additionally created a pull request for a `cat` command like it was proposed in the pull request you mentioned, see: #4532

lib/install.ps1
```diff
@@ -34,7 +38,18 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
         return
     }
 
-    write-output "Installing '$app' ($version) [$architecture]"
+    if (Test-ManifestReviewEnabled) {
```
rashil2000, line 41:
```suggestion
    if (get_config 'manifest-review' $false) {
```

The function is not being used anywhere else, you can directly check the config value. No need for a new function here.

lib/install.ps1
```diff
@@ -9,6 +9,10 @@ function nightly_version($date, $quiet = $false) {
     "nightly-$date_str"
 }
 
+function Test-ManifestReviewEnabled {
+    return get_config 'manifest-review' $false
+}
+
```
rashil2000, line 15:
```suggestion

```

rashil2000 (changes requested):

issaclin32:
I think it won't hurt to add this small feature, even if I personally prefer to use `scoop cat` to check first.

issaclin32 (approved):

niheaven:
Don't annoy user with Y/N question. If someone really likes to see manifest before installation, trigger config.

lib/install.ps1
```diff
@@ -34,7 +34,18 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
         return
     }
 
-    write-output "Installing '$app' ($version) [$architecture]"
+    if (get_config 'manifest-review' $false) {
+        $show_manifest = read-host -Prompt "Show manifest? [y/N]"
```
rashil2000, line 38: This statement block can be removed, because if the user has already set `manifest-review` in config, no need to ask again.

lib/install.ps1
```diff
@@ -34,7 +34,18 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
         return
     }
 
-    write-output "Installing '$app' ($version) [$architecture]"
```
niheaven, line 37: Why delete this line?

lib/install.ps1
```diff
@@ -34,7 +34,18 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
...
+        if ($show_manifest -eq "y") {
+            write-output "manifest:"
+            write-output $manifest | ConvertToPrettyJson
+            $answer = read-host -Prompt "Continue installation? [Y/n]"
+
+            if ($answer -eq "n") {
+                return
+            }
+        }
+    }
```
niheaven, line 48:
```suggestion
    if (get_config 'manifest-review' $true) {
        Write-Output 'Manifest:'
        Write-Output $manifest | ConvertToPrettyJson
        $answer = Read-Host -Prompt "Continue installation? [Y/n]"
        if (($answer -eq 'n') -or ($answer -eq 'N')) {
            return
        }
    }
```

niheaven (changes requested):

rashil2000 (approved):

lib/install.ps1
```diff
@@ -34,6 +34,14 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
         return
     }
 
+    if (get_config 'manifest-review' $false) {
```
niheaven, line 37:
```suggestion
    if (get_config 'manifest-review' $true) {
```

niheaven (changes requested):
If and only if you want to `manifest-review`

lib/install.ps1
rashil2000: I'm confused. Don't we want to disable the review if a user has not set anything in the config?

Passing `true` as the second argument here will always enable the review, unless a user specifically disables it.

niheaven: Yes, `get_config` function's param is really strange..., Ignore it.

niheaven (approved):

rashil2000:
One side-effect of this is that it asks for manifest review even when running the `scoop update *` command, which becomes very annoying. @niheaven is there a way/flag to detect if the `scoop update` command is being used, in lib/install.ps1?

niheaven:
Please test the PR.

PR #4192, opened 2020-12-02
fix(decompress): use innoextract if innounp fails

PorridgePi:
Right now, the latest version of innounp (0.49) does not support newer versions of Inno Setup (like version 6.1.0). While the issue has already been raised on its SourceForge ([here](https://sourceforge.net/p/innounp/discussion/353235/thread/23cf78c92c/) and [here](https://sourceforge.net/p/innounp/discussion/353235/thread/4a7e2b1f19/#6f3b)), an update to innounp does not seem imminent.

As a result, many applications could not be installed due to `decompress error`. These include lukesampson/scoop-extras#5223 lukesampson/scoop-extras#5213 and many more which I would list at the end.

I found [dscharrer/innoextract](https://github.com/dscharrer/innoextract). It is well supported and updated, and it works with the newer versions of Inno Setup. Thus, I edited `decompress.ps1` so that if innounp fails, innoextract will be tried. I have also edited the code of `depends.ps1` and `core.ps1` to make innoextract dependent.

I apologise in advance if my pull request and/or commit is not up to the standard because I am somewhat inexperienced. However, I still hope this would help others who are facing this issue.

Thanks!

Known Issues:
lukesampson/scoop-extras#5223
lukesampson/scoop-extras#5213
lukesampson/scoop-extras#5218
lukesampson/scoop-extras#5136
lukesampson/scoop-extras#5127
lukesampson/scoop-extras#5186
lukesampson/scoop-extras#5225

### DIFF
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -191,17 +191,17 @@ function Expand-InnoArchive {
     }
     $Status = Invoke-ExternalCommand (Get-HelperPath -Helper Innounp) $ArgList -LogPath $LogPath
     if (!$Status) {
-		$InnoextractLogPath = "$(Split-Path $Path)\innoextract.log"
-		$InnoextractArgList = @($Path, "-d", $DestinationPath)
-		$NewStatus = Invoke-ExternalCommand (Get-HelperPath -Helper innoextract) $InnoextractArgList -LogPath $InnoextractLogPath
-		
-		if (!$NewStatus) {
-			abort "Failed to extract files from $Path.`nLog file:`n  $(friendly_path $LogPath)`n  $(friendly_path $InnoextractLogPath)`n$(new_issue_msg $app $bucket 'decompress error')"
-		} else {
-			dir $DestinationPath\app\* | mv -dest $DestinationPath
-			rm $DestinationPath\app\
-		}
-	}
+        $InnoextractLogPath = "$(Split-Path $Path)\innoextract.log"
+        $InnoextractArgList = @($Path, "-d", $DestinationPath)
+        $NewStatus = Invoke-ExternalCommand (Get-HelperPath -Helper innoextract) $InnoextractArgList -LogPath $InnoextractLogPath
+        
+        if (!$NewStatus) {
+            abort "Failed to extract files from $Path.`nLog file:`n  $(friendly_path $LogPath)`n  $(friendly_path $InnoextractLogPath)`n$(new_issue_msg $app $bucket 'decompress error')"
+        } else {
+            dir $DestinationPath\app\* | mv -dest $DestinationPath
+            rm $DestinationPath\app\
+        }
+    }
     if (Test-Path $LogPath) {
         Remove-Item $LogPath -Force
     }
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -198,8 +198,8 @@ function Expand-InnoArchive {
         if (!$NewStatus) {
             abort "Failed to extract files from $Path.`nLog file:`n  $(friendly_path $LogPath)`n  $(friendly_path $InnoextractLogPath)`n$(new_issue_msg $app $bucket 'decompress error')"
         } else {
-            dir $DestinationPath\app\* | mv -dest $DestinationPath
-            rm $DestinationPath\app\
+            Get-ChildItem $DestinationPath\app\* | Move-Item -Destination $DestinationPath
+            Remove-Item $DestinationPath\app\
         }
     }
     if (Test-Path $LogPath) {
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -191,16 +191,42 @@ function Expand-InnoArchive {
     }
     $Status = Invoke-ExternalCommand (Get-HelperPath -Helper Innounp) $ArgList -LogPath $LogPath
     if (!$Status) {
-        $InnoextractLogPath = "$(Split-Path $Path)\innoextract.log"
-        $InnoextractArgList = @($Path, "-d", $DestinationPath)
-        $NewStatus = Invoke-ExternalCommand (Get-HelperPath -Helper innoextract) $InnoextractArgList -LogPath $InnoextractLogPath
-        
-        if (!$NewStatus) {
-            abort "Failed to extract files from $Path.`nLog file:`n  $(friendly_path $LogPath)`n  $(friendly_path $InnoextractLogPath)`n$(new_issue_msg $app $bucket 'decompress error')"
-        } else {
+        abort "Failed to extract files from $Path.`nLog file:`n  $(friendly_path $LogPath)`n$(new_issue_msg $app $bucket 'decompress error')"
+    }
+    if (Test-Path $LogPath) {
+        Remove-Item $LogPath -Force
+    }
+    if ($Removal) {
+        # Remove original archive file
+        Remove-Item $Path -Force
+    }
+}
+
+function Expand-InnoExArchive {
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory = $true, Position = 0, ValueFromPipeline = $true)]
+        [String]
+        $Path,
+        [Parameter(Position = 1)]
+        [String]
+        $DestinationPath = (Split-Path $Path),
+        [String]
+        $ExtractDir,
+        [Parameter(ValueFromRemainingArguments = $true)]
+        [String]
+        $Switches,
+        [Switch]
+        $Removal
+    )
+    $LogPath = "$(Split-Path $Path)\innoextract.log"
+    $ArgList = @($Path, "-d", $DestinationPath)
+    $Status = Invoke-ExternalCommand (Get-HelperPath -Helper Innoextract) $ArgList -LogPath $LogPath
+    if (!$Status) {
+        abort "Failed to extract files from $Path.`nLog file:`n  $(friendly_path $LogPath)`n  $(friendly_path $InnoextractLogPath)`n$(new_issue_msg $app $bucket 'decompress error')"
+    } else {
             Get-ChildItem $DestinationPath\app\* | Move-Item -Destination $DestinationPath
             Remove-Item $DestinationPath\app\
-        }
     }
     if (Test-Path $LogPath) {
         Remove-Item $LogPath -Force
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -177,9 +177,11 @@ function Expand-InnoArchive {
         [String]
         $Switches,
         [Switch]
-        $Removal
+        $Removal,
+        [Switch]
+        $UseInnoextract
     )
-    if (get_config 'INNOSETUP_USE_INNOEXTRACT') {
+    if ((get_config 'INNOSETUP_USE_INNOEXTRACT') -or ($UseInnoextract)) {
         $LogPath = "$(Split-Path $Path)\innoextract.log"
         $ArgList = @($Path, "-d", $DestinationPath)
         $Status = Invoke-ExternalCommand (Get-HelperPath -Helper Innoextract) $ArgList -LogPath $LogPath
@@ -311,9 +313,9 @@ function extract_msi($path, $to, $removal) {
     Expand-MsiArchive -Path $path -DestinationPath $to -Removal:$removal
 }
 
-function unpack_inno($path, $to, $removal) {
+function unpack_inno($path, $to, $removal, $useinnoextract) {
     Show-DeprecatedWarning $MyInvocation 'Expand-InnoArchive'
-    Expand-InnoArchive -Path $path -DestinationPath $to -Removal:$removal @args
+    Expand-InnoArchive -Path $path -DestinationPath $to -Removal:$removal -UseInnoextract:$useinnoextract @args
 }
 
 function extract_zip($path, $to, $removal) {
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -179,54 +179,29 @@ function Expand-InnoArchive {
         [Switch]
         $Removal
     )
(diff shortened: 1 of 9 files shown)


Ash258 (changes requested):
Unacceptable to have two same dependencies for one thing. Allow it as config option.

Same as for lessmsi

- Do not use tabs
- Never ever use 'linux-like' commands/aliases in powershell scripts

PorridgePi:
I have changed according to your suggestions.
I am only unsure of the config name. For now, I left it as `innoex` as the prefix instead of the original `inno`.

lib/install.ps1
```diff
@@ -582,6 +582,8 @@ function dl_urls($app, $version, $manifest, $bucket, $architecture, $dir, $use_c
         $extract_fn = $null
         if ($manifest.innosetup) {
             $extract_fn = 'Expand-InnoArchive'
+        } elseif($manifest.innoexsetup) {
+            $extract_fn = 'Expand-InnoExArchive'
         } elseif($fname -match '\.zip$') {
             # Use 7zip when available (more fast)
             if (((get_config 7ZIPEXTRACT_USE_EXTERNAL) -and (Test-CommandAvailable 7z)) -or (Test-HelperInstalled -Helper 7zip)) {
```
Ash258, line 589: You need to use something like this:

Best possible add simple switch parameter (`-UseInnoExtract` or just `-InnoExtract`) to expand-innoarchive function

Ash258 (changes requested):

lib/install.ps1
Ash258: I take it back. Sorry.

Only use this

`if ($manifest.innosetup) { if (get_config 'INNOSETUP_USE_INNOEXTRACT' $false) { $depends += 'innoextract' } } else { $depends += 'innounp' }`

And then inside function Expand-InnoArchive use same logic:

```
if get_config ''... {
 $Stratus = INvoke-external Innoextract
} else  {
$status = Invoke-External innounp }
```

PorridgePi: Hi, to add the switch parameter/option, do I edit `scoop-install.ps1` and `scoop-update.ps1` in scoop\libexec?

rasa:
To implement a complete solution, I searched for  [innounp](https://github.com/lukesampson/scoop/search?q=innounp) and found we should consider also updating:
1. [appveyor.yml](https://github.com/lukesampson/scoop/blob/08af9ff6e7f7b017701c0c9114294b13e1e83fb8/appveyor.yml#L14)
1. [test/bin/init.psl](https://github.com/lukesampson/scoop/blob/6df80c1aeef72a3528ae3a5e37faa3de840318a2/test/bin/init.ps1#L21)
1. [libexec/scoop-checkup.ps1](https://github.com/lukesampson/scoop/blob/643cfb0da8369dbbe2964511698763c5ff179492/libexec/scoop-checkup.ps1#L22)

lib/install.ps1
PorridgePi: @Ash258 May I know how exactly do I add a switch parameter? Thanks.

PorridgePi:
Hi, may I know if there is anything else for me to improve on? Maybe other than the switch parameter @Ash258 suggested?

Ash258:
Well this will not work for most of the packages.

Since innoextract is only simple implementation it cannot be used the same way as innounp. See all manifests, which have two different architectures coming from single inno archive.

It does not support extraction the same way as innounp, which will break lots of other stuff if this is configured.

It would make sense to force it only for the relevant. So introducing new switch to the Expand-Inno function and then change all the relevant manifests should solve it.

The configuration option could stay for some experiments and for future when innoextract catch up.

So I would propose to just add new switch `UseInnoextract` to the function and then apply simple condition `if get_config -or $UseInnoextract`, then use innoextract, with message to user, that it could fail in some use cases

PorridgePi:
Apologies for not responding.

Thank you so much for your guidance and explanation. 

Following your instructions, I have added the new switch parameter `UseInnoextract` to both `scoop install` and `scoop update`, as well as an warning when the user tries to use this parameter to install and/or update an application.

lib/decompress.ps1
```diff
@@ -311,9 +313,9 @@ function extract_msi($path, $to, $removal) {
     Expand-MsiArchive -Path $path -DestinationPath $to -Removal:$removal
 }
 
-function unpack_inno($path, $to, $removal) {
+function unpack_inno($path, $to, $removal, $useinnoextract) {
     Show-DeprecatedWarning $MyInvocation 'Expand-InnoArchive'
-    Expand-InnoArchive -Path $path -DestinationPath $to -Removal:$removal @args
+    Expand-InnoArchive -Path $path -DestinationPath $to -Removal:$removal -UseInnoextract:$useinnoextract @args
```
Ash258, line 318: It does not make sense to maintain deprecated functions

lib/install.ps1
```diff
@@ -9,7 +9,7 @@ function nightly_version($date, $quiet = $false) {
     "nightly-$date_str"
 }
 
-function install_app($app, $architecture, $global, $suggested, $use_cache = $true, $check_hash = $true) {
+function install_app($app, $architecture, $global, $suggested, $use_cache = $true, $check_hash = $true, $useinnoextract) {
```
Ash258, line 12: It will be patched in manifests itself.

It is nonsense to have additional switch, and 20 checks for this.

Remove

lib/install.ps1
```diff
@@ -40,7 +40,7 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
     $original_dir = $dir # keep reference to real (not linked) directory
     $persist_dir = persistdir $app $global
 
-    $fname = dl_urls $app $version $manifest $bucket $architecture $dir $use_cache $check_hash
+    $fname = dl_urls $app $version $manifest $bucket $architecture $dir $use_cache $check_hash $useinnoextract
```
Ash258, line 43: remove

lib/install.ps1
```diff
@@ -520,10 +520,10 @@ function dl_progress($read, $total, $url) {
     [console]::SetCursorPosition($left, $top)
 }
 
-function dl_urls($app, $version, $manifest, $bucket, $architecture, $dir, $use_cache = $true, $check_hash = $true) {
+function dl_urls($app, $version, $manifest, $bucket, $architecture, $dir, $use_cache = $true, $check_hash = $true, $useinnoextract) {
```
Ash258, line 523: remvoe

lib/install.ps1
```diff
@@ -520,10 +520,10 @@ function dl_progress($read, $total, $url) {
...
     # we only want to show this warning once
     if(!$use_cache) { warn "Cache is being ignored." }
-
+    if($useinnoextract) { warn "Use of innoextract is experimental as it may not work in all scenarios" }
```
Ash258, line 526: remove

lib/install.ps1
```diff
@@ -604,7 +604,7 @@ function dl_urls($app, $version, $manifest, $bucket, $architecture, $dir, $use_c
             Write-Host "Extracting " -NoNewline
             Write-Host $fname -f Cyan -NoNewline
             Write-Host " ... " -NoNewline
-            & $extract_fn -Path "$dir\$fname" -DestinationPath "$dir\$extract_to" -ExtractDir $extract_dir -Removal
+            & $extract_fn -Path "$dir\$fname" -DestinationPath "$dir\$extract_to" -ExtractDir $extract_dir -Removal -UseInnoextract:$useinnoextract
```
Ash258, line 607: remove

libexec/scoop-checkup.ps1
```diff
@@ -24,6 +24,11 @@ if (!(Test-HelperInstalled -Helper Innounp)) {
     $issues++
 }
 
+if (!(Test-HelperInstalled -Helper Innoextract)) {
+    error "'innoextract' is not installed! It's required for unpacking InnoSetup files. Please run 'scoop install innoextract'."
+    $issues++
+}
+
```
Ash258, line 31: remove

libexec/scoop-install.ps1
```diff
@@ -47,14 +47,15 @@ function is_installed($app, $global) {
     return $false
 }
 
-$opt, $apps, $err = getopt $args 'gfiksa:' 'global', 'force', 'independent', 'no-cache', 'skip', 'arch='
+$opt, $apps, $err = getopt $args 'gfiksa:' 'global', 'force', 'independent', 'no-cache', 'skip', 'arch=', 'useinnoextract'
```
Ash258, line 50: remove

libexec/scoop-install.ps1
```diff
@@ -47,14 +47,15 @@ function is_installed($app, $global) {
...
 if ($err) { "scoop install: $err"; exit 1 }
 
 $global = $opt.g -or $opt.global
 $check_hash = !($opt.s -or $opt.skip)
 $independent = $opt.i -or $opt.independent
 $use_cache = !($opt.k -or $opt.'no-cache')
 $architecture = default_architecture
+$useinnoextract = $opt.useinnoextract
```
Ash258, line 58: remove

libexec/scoop-install.ps1
```diff
@@ -123,7 +124,7 @@ if (Test-Aria2Enabled) {
     warn "Scoop uses 'aria2c' for multi-connection downloads."
     warn "Should it cause issues, run 'scoop config aria2-enabled false' to disable it."
 }
-$apps | ForEach-Object { install_app $_ $architecture $global $suggested $use_cache $check_hash }
+$apps | ForEach-Object { install_app $_ $architecture $global $suggested $use_cache $check_hash $useinnoextract }
```
Ash258, line 127: remove

libexec/scoop-update.ps1
```diff
@@ -27,14 +27,15 @@
 
 reset_aliases
 
-$opt, $apps, $err = getopt $args 'gfiksq:' 'global', 'force', 'independent', 'no-cache', 'skip', 'quiet'
+$opt, $apps, $err = getopt $args 'gfiksq:' 'global', 'force', 'independent', 'no-cache', 'skip', 'quiet', 'useinnoextract'
```
Ash258, line 30: remove

libexec/scoop-update.ps1
```diff
@@ -27,14 +27,15 @@
...
 if ($err) { "scoop update: $err"; exit 1 }
 $global = $opt.g -or $opt.global
 $force = $opt.f -or $opt.force
 $check_hash = !($opt.s -or $opt.skip)
 $use_cache = !($opt.k -or $opt.'no-cache')
 $quiet = $opt.q -or $opt.quiet
 $independent = $opt.i -or $opt.independent
+$useinnoextract = $opt.useinnoextract
```
Ash258, line 38: remove

libexec/scoop-update.ps1
```diff
@@ -153,11 +154,10 @@ function update_scoop() {
     success 'Scoop was updated successfully!'
 }
 
-function update($app, $global, $quiet = $false, $independent, $suggested, $use_cache = $true, $check_hash = $true) {
+function update($app, $global, $quiet = $false, $independent, $suggested, $use_cache = $true, $check_hash = $true, $useinnoextract) {
```
Ash258, line 157: remove

libexec/scoop-update.ps1
```diff
@@ -153,11 +154,10 @@ function update_scoop() {
...
     $old_version = current_version $app $global
     $old_manifest = installed_manifest $app $old_version $global
     $install = install_info $app $old_version $global
-
```
Ash258, line 160: revert

libexec/scoop-update.ps1
```diff
@@ -170,7 +170,7 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
         # check dependencies
         $man = if ($url) { $url } else { $app }
         $deps = @(deps $man $architecture) | Where-Object { !(installed $_) }
-        $deps | ForEach-Object { install_app $_ $architecture $global $suggested $use_cache $check_hash }
+        $deps | ForEach-Object { install_app $_ $architecture $global $suggested $use_cache $check_hash $useinnoextract }
```
Ash258, line 173: remove

libexec/scoop-update.ps1
```diff
@@ -273,7 +273,7 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
         # use the url of the install json if the application was installed through url
         $app = $install.url
     }
-    install_app $app $architecture $global $suggested $use_cache $check_hash
+    install_app $app $architecture $global $suggested $use_cache $check_hash $useinnoextract
```
Ash258, line 276: remove

libexec/scoop-update.ps1
```diff
@@ -334,7 +334,7 @@ if (!$apps) {
 
     $suggested = @{};
     # $outdated is a list of ($app, $global) tuples
-    $outdated | ForEach-Object { update @_ $quiet $independent $suggested $use_cache $check_hash }
+    $outdated | ForEach-Object { update @_ $quiet $independent $suggested $use_cache $check_hash $useinnoextract }
```
Ash258, line 337: remove

Ash258 (changes requested):

lib/install.ps1
PorridgePi: Sorry, I did not understand the part about the patch being in the manifest instead. Really sorry for the hassle caused.

PorridgePi: The patches in the manifest should be something similar to the instaler script in [this manifest](https://github.com/ScoopInstaller/Main/blob/42f1e0956330d19cc44f38375b18528187f89ff7/bucket/plex-server.json), right?

niheaven:
Not needed.

PR #4268, opened 2021-02-24
Added zstd archive extraction based on tar installed by git package

Airr:
Added new function 'Expand-ZstdArchive'.

The only prerequisite for this to work is the installation of the zstd package.

Accesses the git-provided tar via `"$scoopdir\apps\git\current\usr\bin\tar.exe"`

Neither the native tar that comes with Windows, nor the tar package in scoop support zst extraction; only the git-provided version does.

I still need to put together a test-case for this, but it worked with a tweaked gcc package I have here.

### DIFF
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -81,7 +81,7 @@ function Expand-ZstdArchive {
     catch [System.Management.Automation.CommandNotFoundException] {
         abort "Cannot find external Zstd (zstd.exe). Install Zstd (sccop install zstd) manually and try again."
     }
-    
+
     $LogPath = "$(Split-Path $Path)\zstd.log"
     $ArgList = @('-d', '-v',"`"$Path`"" )
 
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -75,26 +75,36 @@ function Expand-ZstdArchive {
         $Removal
     )
 
-    $TarPath = "$scoopdir\apps\git\current\usr\bin\tar.exe"
+    try {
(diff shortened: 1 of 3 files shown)


lib/decompress.ps1
```diff
@@ -35,6 +54,53 @@ function Test-LessmsiRequirement {
...
+        [String]
+        $Switches,
+        [ValidateSet("All", "Skip", "Rename")]
+        [String]
+        $Overwrite,
+        [Switch]
+        $Removal
+    )
+
+    $TarPath = "$scoopdir\apps\git\current\usr\bin\tar.exe"
```
rasa, line 78: This assumes that git was installed locally (`$scoopdir`) and not globally (`$globaldir`). See https://github.com/lukesampson/scoop/wiki/Pre--and-Post-install-scripts

Also, not all users will install git via scoop. I don't per https://github.com/lukesampson/scoop/issues/3877.

See also https://github.com/ScoopInstaller/Main/pull/680

Airr: I see your points, thanks Rasa!

I pushed a change, to use zstd and the 'Expand-7zipArchive' function instead of using git-tar.

nikitalita:
This is a giant blocker for updating packages, several of which are now broken. Please merge this.

scrouthtv:
I just had a look at it myself, for me the tar.exe that comes with the tar package works if you also install the zstd package

niheaven:
Thanks for your job! Zstandard support has been added in #4372.

PR #4292, opened 2021-03-13
fix(install): fallback to default downloader when aria2 fails

### LINKED ISSUE #4472: scoop fails to install anything with aria2
```
C:\Users\zachd>scoop update nano
nano: 5.7-34 -> 5.8-44
WARN  Scoop uses 'aria2c' for multi-connection downloads.
WARN  Should it cause issues, run 'scoop config aria2-enabled false' to disable it.
Updating one outdated app:
Updating 'nano' (5.7-34 -> 5.8-44)
Downloading new version
Set-Content : Could not find a part of the path 'C:\Users\zachd\scoop\cache\nano.txt'.
At C:\Users\zachd\scoop\apps\scoop\current\lib\install.ps1:279 char:9
+         Set-Content -Path $urlstxt $urlstxt_content
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\Users\zachd\scoop\cache\nano.txt:String) [Set-Content], DirectoryNotFoundException
    + FullyQualifiedErrorId : GetContentWriterDirectoryNotFoundError,Microsoft.PowerShell.Commands.SetContentCommand

Starting download with aria2 ...
Exception caught
Exception: [download_helper.cc:562] errorCode=1 Failed to open the file C:\Users\zachd\scoop\cache\nano.txt, cause: File not found or it is a directory


ERROR Download failed! (Error 1) An unknown error occurred
ERROR https://files.lhmouse.com/nano-win/nano-win_9672_v5.8-44-gc04962b5e.7z
    referer=https://files.lhmouse.com/nano-win/
    dir=C:\Users\zachd\scoop\cache
    out=nano#5.8-44#https_files.lhmouse.com_nano-win_nano-win_9672_v5.8-44-gc04962b5e.7z

ERROR & 'C:\Users\zachd\scoop\apps\aria2\current\aria2c.exe' --input-file='C:\Users\zachd\scoop\cache\nano.txt' --user-agent='Scoop/1.0 (+http://scoop.sh/) PowerShell/5.1 (Windows NT 10.0; Win64; x64; Desktop)' --allow-overwrite=true --auto-file-renaming=false --retry-wait=2 --split=5 --max-connection-per-server=5 --min-split-size=5M --console-log-level=warn --enable-color=false --no-conf=true --follow-metalink=true --metalink-preferred-protocol=https --min-tls-version=TLSv1.2 --stop-with-process=12620 --continue --summary-interval 0

Please try again or create a new issue by using the following link and paste your console output:
https://github.com/ScoopInstaller/Main/issues/new?title=nano%405.8-44%3a+download+via+aria2+failed
```

kiennq:
Try to fallback to default downloader when aria2 download failed.
This make the download with aria2 more stable since there's some download links that doesn't support multi-connection.

Closes #4472

### DIFF
--- a/lib/download.ps1
+++ b/lib/download.ps1
@@ -466,7 +466,7 @@ function Invoke-CachedAria2Download ($app, $version, $manifest, $architecture, $
                     Invoke-CachedDownload $app $version $url "$($data.$url.target)" $cookies $use_cache
                 }
             } catch {
-                write-host -f darkred $_
+                Write-Host $_ -ForegroundColor DarkRed
                 abort "URL $url is not valid"
             }
         }
--- a/lib/download.ps1
+++ b/lib/download.ps1
@@ -138,7 +138,7 @@ function Invoke-Download ($url, $to, $cookies, $progress) {
         # Handle manual file rename
         if ($url -like '*#/*') {
             $null, $postfix = $url -split '#/'
-            $newUrl = "$newUrl#/$postfix"
+            $newUrl = "$newUrl`#/$postfix"
         }
 
         Invoke-Download $newUrl $to $cookies $progress
@@ -454,10 +454,21 @@ function Invoke-CachedAria2Download ($app, $version, $manifest, $architecture, $
         Write-Host ''
 
         if ($lastexitcode -gt 0) {
-            error "Download failed! (Error $lastexitcode) $(aria_exit_code $lastexitcode)"
-            error $urlstxt_content
-            error $aria2
-            abort $(new_issue_msg $app $bucket 'download via aria2 failed')
+            warn "Download failed! (Error $lastexitcode) $(aria_exit_code $lastexitcode)"
+            warn $urlstxt_content
+            warn $aria2
+            warn $(new_issue_msg $app $bucket "download via aria2 failed")
+
+            Write-Host "Fallback to default downloader ..."
+
+            try {
+                foreach ($url in $urls) {
+                    Invoke-CachedDownload $app $version $url "$($data.$url.target)" $cookies $use_cache
+                }
+            } catch {
+                write-host -f darkred $_
+                abort "URL $url is not valid"
+            }
         }
 
         # remove aria2 input file when done
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Bug Fixes
 
+- **scoop-download|install|update:** Fallback to default downloader when aria2 fails ([#4292](https://github.com/ScoopInstaller/Scoop/issues/4292))
 - **decompress**: `Expand-7zipArchive` only delete temp dir / `$extractDir` if it is empty ([#6092](https://github.com/ScoopInstaller/Scoop/issues/6092))
 
 ### Code Refactoring


rashil2000:
It would be better to not make this a default, and control it with a config variable instead. @kiennq can you fix the conficts?

niheaven:
A refactoring of `install.ps1` by separate downloading, extracting, and installing will be done after this release, and could you do the resolving after that?

brian6932:
Would it not be better to just set `aria2-max-connection-per-server` to `1` if the download fails on retry?

kiennq:
> It would be better to not make this a default, and control it with a config variable instead. @kiennq can you fix the conficts?

I think it should be by default as this only fallback when the aria2 downloading fail. The warning and instruction to file a bug for aria2 is still showing though.


> Would it not be better to just set `aria2-max-connection-per-server` to `1` if the download fails on retry?

I think once aria2 fails, it would be better to fall back into the default one instead of retrying with aria2 again, no?


> A refactoring of `install.ps1` by separate downloading, extracting, and installing will be done after this release, and could you do the resolving after that?

Done

brian6932:
> I think once aria2 fails, it would be better to fall back into the default one instead of retrying with aria2 again, no?

Yea it's prob better to fallback to the default DLer. I remember when I had some wonky NIC driver settings, aria2c would fail (not really aria2c's fault tbf), but the default DLer worked.

lib/download.ps1
```diff
@@ -138,7 +138,7 @@ function Invoke-Download ($url, $to, $cookies, $progress) {
         # Handle manual file rename
         if ($url -like '*#/*') {
             $null, $postfix = $url -split '#/'
-            $newUrl = "$newUrl#/$postfix"
+            $newUrl = "$newUrl`#/$postfix"
```
niheaven, line 141: Why add this `?

lib/download.ps1
```diff
@@ -454,10 +454,21 @@ function Invoke-CachedAria2Download ($app, $version, $manifest, $architecture, $
...
+            warn $(new_issue_msg $app $bucket "download via aria2 failed")
+
+            Write-Host "Fallback to default downloader ..."
+
+            try {
+                foreach ($url in $urls) {
+                    Invoke-CachedDownload $app $version $url "$($data.$url.target)" $cookies $use_cache
+                }
+            } catch {
+                write-host -f darkred $_
```
niheaven, line 469:
```suggestion
                Write-Host $_ -ForegroundColor DarkRed
```

lib/download.ps1
kiennq: It's done to escape the `#` symbol. The output string is still the in two cases but by escaping the `#` symbol, the [powershell tree sitter module](https://github.com/airbus-cert/tree-sitter-powershell) can properly recognize the syntax and highlight it correctly in the text editor

niheaven: Okay, so just for tree sitter module? I think PowerShell itself could handle this...

kiennq: Yes, purely for code highlighting. PowerShell handles both case the same

niheaven:
Now you could add a changlog entry now.

PR #4382, opened 2021-06-10
added sudo to Set-ItemProperty incl note about sudo

axtens:
Changed lib\diagnostic.ps1 to specify `sudo` on the Set-ItemProperty call as it won't work without elevated privileges. Also added the line that follows the other sudos about needing to install sudo if you don't have it.


niheaven:
@rashil2000 You may forget update CHANGELOG.

rashil2000:
Oh, right. Will add it in a subsequent PR.

PR #4393, opened 2021-06-23
合并最新版本

jovercao:
11


jovercao:
11

9uanhuo:
Does this pull request just change all the github link to the github mirror? That's inappropriate for other users who do not need a github mirror.  In general, the mirror sites are mostly used by the Chinese user ( also include me), and actually they are fine.  But the pull request which replaces  the source site is not a good idea.

PR #4405, opened 2021-07-10
config.json file Nested configuration Conversion exception problem

jilliss:
### When the `config.json` file is converted to hashtable, the keys, values ​​and attributes will be lost

For example
The content of `config.json` is as follows
```json
{
  "SCOOP_BRANCH": "master",
  "lastupdate": "2021-07-11T01:03:58.096058+08:00",
  "SCOOP_REPO": "https://github.com/lukesampson/scoop",
  "aria2-enabled": true,
  "list": "all",
  "test-map": {
    "k1":"v1",
    "k2":"v2",
    "k3":"v3"
  },
  "test-list": [
    "v1",
    "v2",
    "v3"
  ]
}
```
image

### After adding this parameter
image

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -41,7 +41,7 @@ function load_cfg($file) {
     }
 
     try {
-        return (Get-Content $file -Raw | ConvertFrom-Json -ErrorAction Stop)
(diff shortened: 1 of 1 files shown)


jilliss:
It seems that this parameter is
Supported only in 6.0  😭

Issue #4264, opened 2021-02-19
HELP: Another way to install scoop other than downloading from raw.githubusercontent.com


Yaojian:
In China, the [Great FireWall](https://en.wikipedia.org/wiki/Great_Firewall) blocks `raw.githubusercontent.com` so the powershell installation command fails.

Is their another way to install ?

HUMORCE:
This domain has DNS cache poisoning in some areas of China.

temporary solution for this problem, 
get actual IP: https://ip.sb/domain/raw.githubusercontent.com
then add to hosts file:
`actual_IP raw.githubusercontent.com`

without a proxy,
will also be blocked or slowed when downloading most app using scoop.

if you have a proxy server, install this app when scoop installation done:
`scoop install https://github.com/HUMORCE/scoop-nuke/blob/master/bucket/proxychains.json`
for use proxychains, read the docs:
https://github.com/shunf4/proxychains-windows/

for url from github, you can replace the original address with https://fastgit.org or other mirrors.

e.g. proxychains.json
```
replace:
"url": "https://github.com/shunf4/proxychains-windows/releases/download/0.6.8/proxychains_0.6.8_win32_x64.zip",
to:
"url": "https://hub.fastgit.org/shunf4/proxychains-windows/releases/download/0.6.8/proxychains_0.6.8_win32_x64.zip",
```

Dragon1573:
@Yaojian [**CODE.CHINA Mirror**](https://codechina.csdn.net/mirrors) provides Scoop itself and 10 "known buckets". To install Scoop via **CODE.CHINA**, you can act as follows:

1. The installation process uses 2 important files: `bin/install.ps1` and `lib/core.ps1`. Download them to your device before installation. 
2. Modify `install.ps1` file. At line 31, change the URL from `raw.githubusercontent.com` to your local `core.ps1`. 
3. Also, modify the URL at line 44 and line 60 to CODE.CHINA mirror. 
4. That's it! Excecute your local `install.ps1` and Scoop will be installed via that mirror. 
5. "Scoop known buckets" are only refered to those repositories on GitHub. When adding them with `scoop bucket add`, specify them to CODE.CHINA mirror. If they already added, enter the bucket directory (such as `$SCOOP/buckets/main`) and change the Git URL with `git remote set-url origin`. Scoop will update these buckets through the mirror instead of GitHub. 

[...]

PR #4406, opened 2021-07-11
Feature/support mirror configuration

jilliss:
Sometimes the address in the bucket becomes invalid or inaccessible
Or it may be slow to visit in some areas, such as China

I modified the url parsing script and now it will parse the mirrirs property in config.json. which is a map type property
Below is my `config.json`file
```json
{
  "SCOOP_BRANCH": "master",
  "lastupdate": "2021-07-11T10:33:24.5707181+08:00",
  "SCOOP_REPO": "https://github.com/lukesampson/scoop",
  "aria2-enabled": false,
  "list": "all",
  "proxy": "127.0.0.1:10809",
  "mirrors": {
    "http://repo.msys2.org/": " https://repo.huaweicloud.com/msys2/"
  }
}

```
At this time, scoop will download files from `repo.huaweicloud.com` instead of `repo.msys2.org`

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -41,7 +41,7 @@ function load_cfg($file) {
     }
 
     try {
-        return (Get-Content $file -Raw | ConvertFrom-Json -AsHashtable -ErrorAction Stop)
+        return (Get-Content $file -Raw | ConvertFrom-Json -ErrorAction Stop)
     } catch {
         Write-Host "ERROR loading $file`: $($_.exception.message)"
     }
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -41,7 +41,7 @@ function load_cfg($file) {
     }
 
     try {
-        return (Get-Content $file -Raw | ConvertFrom-Json -ErrorAction Stop)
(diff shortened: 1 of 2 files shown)


jilliss:
@r15ch13   Can you help me see if there is a problem?
There seems to be something wrong with the automated test script

niheaven:
Please search https://github.com/chawyehsu/dorado for mirrors in China, and submit PR if any need.

jilliss:
What I want is to change the location of the package download through configuration. Instead of downloading the json file in the bucket from gitee or from github
This PR is to replace the url by reading the configuration file

StarsbySea (commented):
I am very supportive of this addition, at times the network may incorrectly block certain domains and with this feature it is possible to configure mirror sites for certain domains so that the download and installation can run successfully. Of course a specific example is speeding up downloads from Github in China, but obviously there are many more use cases.
However, this may need more work, such as providing an option in the scoop config to manually add the domains that need to be mirrored. See the discussion at https://github.com/ScoopInstaller/Scoop/issues/4264#issuecomment-1008052093.
I wonder if the original author would be interested in continuing to implement this feature?
Also please consider changing the target branch to develop

PR #4410, opened 2021-07-16
fix(shortcuts): Fix network drive shortcut creation

cyingfan:
Calls Convert-Path on Filesystem object (returned `by shortcut_folder()`) so it works on start menu mounted on network drive

Closes https://github.com/lukesampson/scoop/issues/3489


dwettstein (dismissed):
Basically this looks fine for me. I would need just two things:
- Could you please change the target branch from `ScoopInstaller:master` to `ScoopInstaller:develop`.
- Could you please confirm that this works on network drives as well as on normal drives.

Thanks for your PR!

issaclin32 (approved):

PR #4420, opened 2021-08-03
Improve local search performance

### LINKED ISSUE #4239: Improve Search Performance
https://github.com/shilangyu/scoop-search has changed my experience with scoop, the slow search has always been one of my pain points as a user.

I'd be willing to PR improvement to the search engine to incorporate the lessons learnd here if there is interest.

Bios-Marcel:
We are not doing a re-lookup of the bucker directory anymore, which saves a lot of time.

I've simply inlined what "manifest" did before. Would you prefer if I were to make another function called `manifest_by_dir` which i could also refer to in `manifest_path`?

Fixes #4239

### DIFF
--- a/libexec/scoop-search.ps1
+++ b/libexec/scoop-search.ps1
@@ -25,7 +25,8 @@ function bin_match($manifest, $query) {
 }
 
 function search_bucket($bucket, $query) {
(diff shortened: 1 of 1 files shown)


Bios-Marcel:
I've made a major mistake here, sorry :D

PR #4469, opened 2021-10-01
updating parsing

tukanos:
Adding support underscore '_' in bucket's name.  Also the dot '.' before filename suffix should be escaped.


tukanos:
@niheaven  Ok.  Just note this patch contains also other small fix of the regexp.

PR #4342, opened 2021-04-28
Change `url()` scope to avoid conflict with global aliases (e.g. Pansies)

lediur:
Pansies, a popular PowerShell module, creates an alias from `url` to its [`New-Hyperlink`](https://github.com/PoshCode/Pansies/blob/master/Source/Assembly/Commands/NewHyperlinkCommand.cs) function. This conflicts with usages of the `url` function in [`./lib/manifest.ps1`](https://github.com/lukesampson/scoop/blob/master/lib/manifest.ps1#L114).

This PR specifies the scopes of those usages to the script level so those usages can find the function in the scoop library.

Fixes https://github.com/lukesampson/scoop/issues/4185

PR #4013, opened 2020-06-08
Fix #4000, handle arch-specific env_add_path

sinloss:
This fixes #4000

PR #4493, opened 2021-11-03
feat(scoop-prefix): remove unused imports and functions. Load when required

pratikpc:
Unused imports removed:-
1. manifest.ps1
2. buckets.ps1

Unused function removed:-
1. reset_aliases:- We do not use aliases here.

Load when required
1. help needs to be loaded only when app is not specified
2. core needs to be loaded only when app is specified

Gains:-
1. Improved speed.


rashil2000:
Retarget to develop branch. Master is for release.

rashil2000:
@niheaven would be able to suggest more on this.

niheaven:
@pratikpc Except for #4013 #4342 #4492, other merged PRs in `master` are only for text works. The recent commit (#3518) in `develop` is on the main develop line of scoop itself.

There are many feathers under development and they all should be committed into `develop` and tested by advanced users (incl. maintainers), only critical hotfixes that may affect normal users should be merged directly into `master` and synchronized to `develop`.

These commits that ahead `develop` now could be merged into `develop` after new commits in `master` and `develop` are all carefully tested, and then we can sync the branches and merge new PRs into `develop` and test them again.

pratikpc:
That makes sense. 

Although we really should first rebase develop on master so that we may not end up with future conflicts.

PR #3721, opened 2019-10-29
refactor(versions): Refactor 'versions.ps1'

niheaven:
Reworking of #3333, #3578, #3648 

Fix #2720, fix #3687, fix #4236, fix #4244, fix #4306

Get correct current version from `apps\xxx\current\manifest.json`, so works well after `scoop reset` or version roll-back. If `NO_JUNCTIONS` is set, use creating time of `install.json` to find last installed version as current version.

Add `force-update` config option that forces to update to manifest version (for version roll-back).

Now use '+' as post-release delimiter, and manifest maintainer should trim metadata in 'checkver.regex'. #3613

PR #4504, opened 2021-11-09
Fix schema to support + in version

rasa:
See 	https://github.com/ScoopInstaller/Main/blob/3590444367de0178ad6a35d861a24c9f105fd466/bucket/x265.json#L2


Valinor (approved):

niheaven:
For version comparison, see https://github.com/ScoopInstaller/Scoop/pull/3721

PR #2456, opened 2018-08-03
Include license and shortcuts in autoupdate in schema.json

se35710:
Fixes #2454

PR #4528, opened 2021-11-22
feat(schema): Enable autoupdate for 'license'

### LINKED ISSUE #2454: license property in autoupdate missing in schema.json
From AppVeyor;
```
    [*] wlp-javaee.json: Property 'license' has not been defined and the schema does not allow additional properties.
      [-] wlp-javaee.json 144ms
        Expected 0, but got 1.
        186:                         $validator.Errors.Count | should be 0
        at Invoke-LegacyAssertion, C:\Program Files\WindowsPowerShell\Modules\Pester\4.4.0\Functions\Assertions\Should.ps1: line 188
        at <ScriptBlock>, C:\projects\scoop-ibm\Scoop-Bucket.Tests.ps1: line 186
```

In [`wlp-javaee.json`](https://github.com/se35710/scoop-ibm/blob/master/wlp-javaee.json):

```
    "autoupdate": {
        "url": "https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/downloads/wlp/$version/wlp-javaee7-$version.zip",
        "license": "https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/downloads/wlp/$version/lafiles/runtime/en.html"
},
```

During updates, the license property is updated as expected.

niheaven:
- Fix #2454
- Close #2456
- Also make `description`, `homepage` and `license` properties required.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -220,6 +220,9 @@
                 },
                 "persist": {
                     "$ref": "#/definitions/stringOrArrayOfStringsOrAnArrayOfArrayOfStrings"
+                },                
+                "license": {
+                    "$ref": "#/definitions/license"
                 },
                 "note": {
                     "$ref": "#/definitions/stringOrArrayOfStrings"
@@ -593,7 +596,10 @@
         }
     },
     "required": [
-        "version"
+        "version",
+        "description",
+        "homepage",
+        "license"
     ],
     "title": "scoop app manifest schema",
     "type": "object"


niheaven:
If there are some properties that really need to be updated, please add a comment.

rashil2000:
`url` is also a mandatory property, right?

niheaven:
It isn't if we want some meta manifest later.

rashil2000:
But for now we have no such feature for meta manifest. For now we should make `url` required. We can change it back later.

niheaven:
arch structure has arch url and no noarch url, so this could not be requied...

pratikpc:
> Also make description, homepage and license properties required.

This won't cause problems with repos which are using Scoop to maintain and update their own buckets right?

niheaven:
Yes, this is just a schema, and modification doesn't change user experience.

rasa:
Making the `description`, `homepage` and `license` properties required, has caused hundreds of outside manifests to fail validation. For internal buckets, we can certainly require these fields, but they not are required for scoop to operate correctly. 

At the very least, I like us to consider making the `description` field optional. Thoughts?

niheaven:
So the only field that make scoop workable is `version`, right? Now `url` or `arch.url` and hashes are required, too, but the schema cannot properly check it.

I'll remove `description` and keep `homepage` and `license` since they are essential for an app, and could discuss if removing these two.

PR #2260, opened 2018-05-16
Option to skip hash validation and error message improvements

r15ch13:
- Add `show_app()` function for displaying `app/bucket@version`
- Add option (`-s` | `--skip`) to skip hash validation on `scoop install` and `scoop update` to allow installation when a manifest provides an outdated hash
- Add `new_issue_msg()` function for nicer error messages with a link to the buckets repository. clicking on the link opens a new issue with a predefined title
- Show filename instead of unreadable `robocopy` output when file moving fails and use `new_issue_msg()`
- Handle hash validation errors with `new_issue_msg()`

Hm, maybe I put too much in this pull-request ... one change leads to another. 🤷‍♂️ 😁 

Example output: 

```
λ .\bin\scoop.ps1 install wox
Installing 'wox' (1.3.524) [64bit]
Wox-1.3.524-full.nupkg (5,0 MB) [===========================] 100%
Checking hash of Wox-1.3.524-full.nupkg... ERROR Hash check failed!
App:       extras/wox
URL:       https://github.com/Wox-launcher/Wox/releases/download/v1.3.524/Wox-1.3.524-full.nupkg#/dl.7z
Expected:  bfb73e46c3c054c00c24ab0f03ed441ab1308e07
Actual:    9025a4d5264b9cf73cd48a2679b91f1127e46c10

Please create a new issue by using the following link and paste your console output:
https://github.com/lukesampson/scoop-extras/issues/new?title=wox%401.3.524%3a+hash+check+failed
```

Ultimate goal: Use GitHub Issue API to find reported hash check errors and let [Excavator](https://github.com/r15ch13/Excavator) resolve them automagically.

PR #4531, opened 2021-11-23
fix(scoop-update): Fix scoop update -a requiring arguments

### LINKED ISSUE #4530: fix "scoop update -a"
found [this one](https://github.com/ScoopInstaller/Scoop/commit/0d721abc2f200f9d0dda2fa26d7b68710a3a9840) finally merged (THANKS!!!) and was super excited to test it... 

PRO:
* ```scoop help update``` shows already the merged help info:
```powershell
λ  scoop help update
Usage: scoop update <app> [options]

'scoop update' updates Scoop to the latest version.
'scoop update <app>' installs a new version of that app, if there is one.

You can use '*' in place of <app> to update all apps.

Options:
  -f, --force               Force update even when there isn't a newer version
  -g, --global              Update a globally installed app
  -i, --independent         Don't install dependencies automatically
  -k, --no-cache            Don't use the download cache
  -s, --skip                Skip hash validation (use with caution!)
  -q, --quiet               Hide extraneous messages
  -a, --all                 Update all apps (alternative to '*')
  ```
* ```scoop update --all``` works great

CON: ```scoop update -a``` doesn't.. getting

```powershell
λ  scoop update -a
scoop update: Option -a requires an argument.
```

filmor:
Fixes #4530 

I misunderstood how `:` acts in `getopt`. I have removed it entirely, I assume that the `-q` parameter was also not supposed to get an argument. The error on `-q` was introduced with #2260.

### DIFF
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -28,7 +28,7 @@
 
 reset_aliases
 
-$opt, $apps, $err = getopt $args 'gfiksqa:' 'global', 'force', 'independent', 'no-cache', 'skip', 'quiet', 'all'
+$opt, $apps, $err = getopt $args 'gfiksqa' 'global', 'force', 'independent', 'no-cache', 'skip', 'quiet', 'all'
 if ($err) { "scoop update: $err"; exit 1 }
 $global = $opt.g -or $opt.global
 $force = $opt.f -or $opt.force


niheaven (approved):

niheaven:
ac71fccbecb3d4158f249db9c1b9bb043cb8e966 fixes issue that `scoop update` with no apps (even with `--all`) update scoop core only.

pratikpc:
@niheaven should we merge this one on stable as well?

niheaven:
Is it a critical fix? If so, we could open a merging PR.

Issue #4220, opened 2021-01-09
Request: Fish Shell


avarun42:
Would be awesome to have a package for `fish` shell. There's a package compiled for Windows maintained by [the folks over at MSYS2](https://packages.msys2.org/base/fish).

HUMORCE:
This request has nothing to do with scoop.

but you can install fish shell in MSYS2, then call fish by command: `msys2 -c fish`

or create a function to do this, example:
```powershell
[HUMOR@W ~]$ function fish {msys2 -c fish}
[HUMOR@W ~]$ fish
Welcome to fish, the friendly interactive shell
Type `help` for instructions on how to use fish
HUMOR@W /c/U/HUMOR> echo $FISH_VERSION
3.1.2
```

define fish function in powershell $PROFILE, then you can call fish shell anywhere.

bridystone:
Hi Humorce,

i don't understand the answer. i also would like to have fish installed on my Windows machine. 
i also thought that scoop could be a replacement of msys2 for fish and any other package out there.

is this a misunderstanding of the whole concept?

tahir-hassan:
Yes, I also wanted `fish` to be added to `scoop`, but then after reading about POSIX etc, it seems as though it is not possible until the `fish` people compile a native version for Windows.

As the above respondents are describing, you can install and run `fish` within `msys2`. 

Therefore you would first install `msys2`:

```powershell
scoop install msys2
```

Then you run `msys2` (it may prompt you to re-run it twice), and then you use `msys2`'s own package manager `pacman` to install `fish`:

```bash
pacman -S fish
```

If you wish to have your `scoop` installed programs available while in MSYS,  you could add the following lines to your `~/.bashrc`:

PATH=$PATH:/c/Users/$USER/scoop/shims

Then run `fish` within `msys2`.

Issue #3937, opened 2020-04-22
Feature request: Fish Shell


lucianmarin:
```
scoop install fish
```

Pretty please! Cygwin version is a nightmare!

Ash258:
Will not be implemented

Duplicate of ScoopInstaller/Main#903

rasa:
Duplicate of ScoopInstaller/Main#903

Issue #3053, opened 2019-01-31
Request: zsh


stuaxo:
Request for zsh, hopefully Oh my zsh will work inside it.

Calinou:
Shouldn't one install zsh in the Windows Subsystem for Linux instead?

stuaxo:
I'm using a work machine, although I'm a developer it is locked down.

stuaxo:
I'm using bash for Windows, I'd prefer to use zsh built for mingw

okonomiyaki3000:
@stuaxo I think you're barking up the wrong tree. If you want to use zsh on Windows, you should use MSys2. It includes pacman which lets you install zsh and a lot of other things. oh-my-zsh also works. It's what I use on my work PC because I agree with you, the usual Windows command line is an abomination.

rasa:
Closing as this issue is stale. Please reopen if this is still an issue.

stuaxo:
Will reopen if I find a build of zsh it's easy to integrate

PR #4533, opened 2021-11-24
feat(checkver): add msys support

pratikpc:
Supports:

1. Custom URLs for alternate MSYS or MSYS like repos
2. Custom Regex. For alternate MSYS and MSYS like repos with different naming conventions

Usage: 

```json
    "checkver": {
        "msys": [
            "diffutils",
            "msys2-runtime",
            "libiconv",
            "libintl"
        ]
    },
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": [
                    "http://repo.msys2.org/msys/x86_64/diffutils-$matchDiffutils-x86_64.pkg.tar.zst",
                    "http://repo.msys2.org/msys/x86_64/msys2-runtime-$matchMsys2-Runtime-x86_64.pkg.tar.zst",
                    "http://repo.msys2.org/msys/x86_64/libiconv-$matchLibiconv-x86_64.pkg.tar.zst",
                    "http://repo.msys2.org/msys/x86_64/libintl-$matchLibintl-x86_64.pkg.tar.xz"
                ]
            }
        }
    }
```

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -114,6 +114,7 @@ $Queue | ForEach-Object {
     $jsonpath = ''
     $xpath = ''
     $replace = ''
+    $msys = @()
 
     if ($json.checkver -eq 'github') {
         if (!$json.homepage.StartsWith('https://github.com/')) {
@@ -149,6 +150,20 @@ $Queue | ForEach-Object {
         $replace = $json.checkver.replace
     }
 
+    if ($json.checkver.msys -and $json.checkver.msys -is [array]) {
+        # Allow users to setup alternate MSYS repos
+        if (!$json.checkver.url) {
+            $url = "https://repo.msys2.org/msys/x86_64/"
+        }
+        # Allow users to support other regexes
+        # The ScoopMSYSSubstituteAppName is replaced by an app name
+        # Name has to be chosen to ensure minimum clashes
+        if (!$regex) {
+            $regex = "<a href=""ScoopMSYSSubstituteAppName-(?<version>\d[\w.-]+)-(x86_64|i686|any).pkg.tar.(xz|zst).sig"""
+        }
+        $msys = $json.checkver.msys
+    }
+
     if (!$jsonpath -and !$regex -and !$xpath) {
         $regex = $json.checkver
     }
@@ -166,6 +181,7 @@ $Queue | ForEach-Object {
         xpath    = $xpath;
         reverse  = $reverse;
         replace  = $replace;
+        msys = $msys;
     }
 
     $wc.Headers.Add('Referer', (strip_filename $url))
@@ -250,7 +266,23 @@ while ($in_progress -gt 0) {
         $ver = ''
     }
 
-    if ($regexp) {
+    if ($msys) {
+        $ver = ''
+        $matchesHashtable = @{}
+        $msys | ForEach-Object{
+            $regexp_msys = $regexp.Replace('ScoopMSYSSubstituteAppName', $_)
+            $regex_msys = New-Object System.Text.RegularExpressions.Regex($regexp_msys)
+            $match = $regex_msys.Matches($page) | Select-Object -Last 1
+            if ($match -and $match.Success) {
+                $msys_app_ver = $match.Groups['version'].Value
+                $matchesHashtable.Add($_, $msys_app_ver)
+                $ver += "$msys_app_ver-"
+            }
+        }
+        $ver = $ver.TrimEnd('-')
+    }
+
+    if ($regexp -and !$msys) {
         $regex = New-Object System.Text.RegularExpressions.Regex($regexp)
         if ($reverse) {
             $match = $regex.Matches($page) | Select-Object -Last 1


pratikpc:
@ScoopInstaller/maintainers need some help on this one

With checkver msys active, I presume we also need to update the schema

pratikpc:
Example PR with sample usage here

https://github.com/ScoopInstaller/Main/pull/3000

pratikpc:
Advantages of merging PR
1. OpenSSH on Main bucket is currently a beta of an older version
2. GCC on main bucket is currently older
3. Diff Utils is older
4. Can switch to MSYS for quite a few packages without the need to update manually instead of relying on 3rd party sources

rashil2000:
TL;DR: this is a bad idea. 

Note that there were reasons other than manifest maintenance that I mentioned in my comment in the relevant thread (to which you didn't reply). I'd urge you (and any future maintainers that stumble upon this) to properly discuss the pros and cons of a such a big change so that no one has to spend time on making PRs beforehand.

My main reasons for opposing this are:
- **Space bloat**: Even if this gets implemented, let's say I want to install 9 packages that depend on the msys2-runtime and bash. I will then have 9 different installations of the POSIX emulation layer, and soon my Scoop folder will be bigger than C:\Windows. It's not like Scoop is going to aggregate the msys2-runtime dependency: it doesn't care about other apps. A proper implementation of this PR would be make msys2-runtime and bash standalone packages themselves, and then add them as dependencies for packages like lftp etc. But then, you see where this is going - you just created (a very poor) implementation of pacman.
- **Runtime incompatibilities**: Now if some of those 9 apps have different runtime DLL versions (because the manifest didn't get updated or some other reason), all my 9 apps will become incompatible with each other. This will increase the maintenance burden 10x now that people will start opening issues regarding this.
- **MSYS2 quirks**: POSIX emulation layers like Cygwin and MSYS2 have enough quirks and annoyances that make it sometime difficult to use it outside of the MSYS2/Cygwin shell. I really really don't want people complaining that their fish shell doesn't work because of some UNIX-Windows compatibility mixes. You should realize that once you start supporting MSYS2 programs, people will make requests about adding POSIX programs to Scoop. This has already happened in the past - ScoopInstaller/Main#903, #3053, #3937, #4220.
- **Why give preference to MSYS2?**: The Scoop project is not affiliated to MSYS2 in any way, why would you give preference to it? What if tomorrow someone comes and asks for similar packages in Cygwin? Chocolatey? Or some \<insert some other package database here\>?
- **Unnecessary complexity**: Implementation of this PR is not trivial at all, and neither is maintenance as a result. There's a reason why Scoop doesn't handle NPM, PIP or GEM packages. Because they are handled far better by their own respective package managers. Besides, the Scoop repo has far more outstanding and necessary issues to be fixed, I think those should be prioritized much more.

> OpenSSH on Main bucket is currently a beta of an older version

It is NOT a beta. That win32 version works a far better with native Windows applications than the POSIX-emulated MSYS2 version. That win32 version is currently installed in 80% of the world's desktop computers, I'm not sure what qualifies as "stable enough" for you.

As for MSYS2, it uses a rolling release model following Arch-Linux - it is always going to have the latest and greatest of everything. That doesn't mean it's going to be the most stable/compatible for all Windows-native apps (by Windows-native apps I mean apps that don't rely on the POSIX emulation layer).

Scoop does one job - installing archives/installers - and it does that _bloody well_. Please don't try to make it do more stuff for the sake of feature bloat - and let's be frank, it's never going to be as robust as pacman or apt-cyg, at a job that suits pacman or apt-cyg.

pratikpc:
> Space bloat

Decision to accept or reject PRs with MSYS presence is up to the maintainers of the respective buckets and beyond the remit of this simple PR.

>  A proper implementation of this PR would be make msys2-runtime and bash standalone packages themselves, and then add them as dependencies for packages like lftp etc

That implementation is beyond the remit of this PR.

> Now if some of those 9 apps have different runtime DLL versions (because the manifest didn't get updated or some other reason), all my 9 apps will become incompatible with each other. 

A single package failing to update, single regex check not working would lead to failure for them all. As is tradition with most Scoop packages. I don't see any issues being faced by this PR regarding that.

> fish shell doesn't work because of some UNIX-Windows semantic mixes

In such cases, I believe maintainers should not add packages which can have a few expected incompatibilities. Using what this PR provides is up to the maintainers.

> You should realize that once you start supporting MSYS2 programs, people will make requests about adding POSIX programs to Scoop

It's only fair to make requests. Scoop still to date supports MSYS2, especially given the ZSTD support in stable. This only allows packages to be updated to reduce efforts that maintainers who wish to accept and support MSYS packages have to take.

> This has already happened in the past

It is up to the maintainers to accept or reject MSYS packages with specific reasoning as can be seen. 

> Why give preference to MSYS2?

It is a popular repository of packages. Their repo is easy to parse (single page) and support and contains quite a few packages which are popular among users. Also as can be seen, adding support for them is fairly simple.

> What if tomorrow someone comes and asks for similar packages in Cygwin? Chocolatey?

I am unsure if the implementation would be as simple as MSYS2 (Single page repo). If it is as simple, this would already provide support for it.

> Implementation of this PR is not trivial at all, and neither is maintenance as a result.

Could you please tell me where simplifications could be made? I did notice that MSYS and regex sections share a bit of their code but was unsure if a combination of them both would be the best of ideas.

I am willing to modify and cleanup the code for obvious reasons. It is fairly important to Scoop that it be done.

> Scoop repo has far more outstanding and necessary issues to be fixed, I think those should be prioritized much more.

That would be a decision I support. New features such as this would definitely be on a lower priority scale compared to fixing existing bugs. If there are simplifications to be made due to complexity as you mention, it is fairly important for this PR to make them and that would obviously take its own sweet time.

> it is always going to have the latest and greatest of everything.

Similar to Scoop I would assume. Scoop too tries our best to ship with the newest of versions irrespective of support or compatibility unless maintainers choose at their own discretion to do otherwise.

> it's never going to be as robust as pacman or apt-cyg, at a job that suits pacman or apt-cyg.

That would be beyond the remit of this PR. This only attempts to update the MSYS packages.

> It is not unstable though.

It is a beta of an older version according to MS themselves. If this PR allows maintainers to add support for newer versions according to their own discretion, then it is certainly helpful. It is of course upto the maintainers to accept the PRs

rashil2000:
> Decision to accept or reject PRs with MSYS presence is up to the maintainers of the respective buckets and beyond the remit of this simple PR.

No, the existence of this PR and feature implies that people will request, and therefore rightly expect maintainers to merge those PRs. You can't just add a feature and disallow people to use it.

> That implementation is beyond the remit of this PR.

It is the _proper_ implementation. We can't pretend that all packages are dumb (in the sense they are unaware of what else exists, especially in MSYS2's context).

> In such cases, I believe maintainers should not add packages which can have a few expected incompatibilities. Using what this PR provides is up to the maintainers.

Who has got the time to stress-test all MSYS2 packages if they work well outside of the MSYS2 shell? This is a false expectation. No one knows what "expected incompatibilities" the future MSYS2 packages will have. We don't want to keep redirecting people to the MSYS2 repo.

> This only allows packages to be updated to reduce efforts that maintainers who wish to accept and support MSYS packages have to take.

Again, as I said, the existence of such feature implies the expectation of people to merge such PRs. You can't confine people to use third-party buckets if such a feature exists in the Scoop core itself.

> It is up to the maintainers to accept or reject MSYS packages with specific reasoning as can be seen.

Who are these maintainers? You speak as if maintainers are some separate entity independent of Scoop core. _We_ are the maintainers, and the points I've written are written as a part of the maintainers team, not as an outsider.

> Their repo is easy to parse (single page) and support and contains quite a few packages which are popular among users. Also as can be seen, adding support for them is fairly simple.

Would you mind explaining how so? For example, the [lftp](https://packages.msys2.org/package/lftp?repo=msys&variant=x86_64) page here just lists the version of lftp. How will you track the version numbers of dependencies of lftp? These are the dependencies of lftp:

image

Now I have the same question for all recursive dependencies of _these_ dependencies. How is this simple in any sense? Are you planning to implement a topological sort in this PR? If so, again, you're trying to do something that pacman already beautifully does. And pacman is not simple as far as I'm aware.

> I am unsure if the implementation would be as simple as MSYS2 (Single page repo)

It is exactly the same for Cygwin. I'm fully expecting people to request support for Cygwin if the MSYS2 support gets added.

> Could you please tell me where simplifications could be made? I did notice that MSYS and regex sections share a bit of their code but was unsure if a combination of them both would be the best of ideas.

I explained the dependency-tree problem above.

> It is a beta of an older version according to MS themselves. If this PR allows maintainers to add support for newer versions according to their own discretion, then it is certainly helpful. It is of course upto the maintainers to accept the PRs

Again I'm not sure what your expectations after getting this PR merged are. You yourself are the maintainer.

Also, I would vastly prefer that the official version of a package that Scoop provides (openssh in this case), is the one from an official source, which is guaranteed to work with Windows-native applications (looking at you, Node), over some third-party package, that too which is POSIX-emulated. Even if that means being a version behind. Windows's philosophy has always been about compatibility. Always having the latest of everything brings little benefit most of the time.

rashil2000:
One more thing, I assume this PR is not complete, so mark it as a draft for now.

rasa:
@pratikpc While I applaud your goals with this PR, I think the drawbacks of implementing it in this way, outway its benefits. 

A better option might be a script that creates an msys bucket by parsing msys' files, and generating manifests for each of its apps.

erwin:
What about just making a symlink from the scoop's msys2/current/usr/bin/pacman.exe into scoop's shim directory, and then anybody that needs an msys library just runs something like:

```
pacman.exe -S mingw-w64-x86_64-gtk3
```

Very interesting debate on this thread. As just a scoop user I was initially thinking "yeah, merge it, more easily installed packages!" but reading through the debate the dependency issue is very real...

And I don't think you  guys even got into the "build time dependencies" vs "run time dependencies" which is it's own major issue. I'm pretty sure most of the dependencies in `lftp` example given are build time dependencies.

rashil2000:
> What about just making a symlink from the scoop's msys2/current/usr/bin/pacman.exe into scoop's shim directory, and then anybody that needs an msys library just runs something like:
> 
> ```
> pacman.exe -S mingw-w64-x86_64-gtk3
> ```

That can definitely be done, but there's a caveat: since pacman will now be accessible from anywhere, this will give users the false impression that the packages installed through pacman are also available everywhere - which is not true since they'll only be available inside the Msys2 shell. And pacman can be accessed through the Msys2 shell anyway, no need for a shim there. 

> And I don't think you guys even got into the "build time dependencies" vs "run time dependencies" which is it's own major issue. I'm pretty sure most of the dependencies in `lftp` example given are build time dependencies.

No, the dependencies I mentioned are runtime dependencies only - the build time dependencies are different, see below:

Screenshot_20211217-130647_Edge_Canary

But since pacman packages are pre-built, build time dependencies are not required.

rashil2000:
Closing as 'wontfix'. 

@rasa's workaround of an external script/program to scrape MSYS2-packages website is the best way forward for this.

PR #3851, opened 2020-02-14
Make sure exit code is returned from .ps1 shim

ocitrev:
This PR will fix issue #3850.
When a script returns a non zero value, the automatic variable `$?` is set to `$false`
Adding the line `exit $LastExitCode` will properly set the `$?` automatic variable when a .ps1 shim is executed.

Issue #2230, opened 2018-04-27
Special characters in Windows user names will cause command scripts in "shims" to fail.


bjornakr:
My Windows username, automatically created when logging into Microsoft at install time, contains the letter 'ø'.

This is the real path of my recently installed application:
`c:\Users\Bjørn\scoop\apps\sbt\current\bin\`

This is the contents of the shim command:
`@"C:\Users\Bj?rn\scoop\apps\sbt\current\bin\sbt.bat"  %*`

When trying to run my application, it gives the following message:
```
>sbt
The system cannot find the path specified.
```

rasa:
Shouldn’t
https://github.com/lukesampson/scoop/blob/master/lib/core.ps1#L292
etc, use utf-8 encoding, instead of ascii?

pendenaor:
I've got the same issue, but this bug seems to be related to an already fixed issue https://github.com/lukesampson/scoop/issues/572

Issue #3915, opened 2020-04-06
cmd extension files have a  full path to their binary instead of a relative path in the shims directory


gigi206:
In the shims directory, all cmd file extensions have a full path to the binary instead of a relative path like the ps1 extension.
For example, this is the case for vscode or vscodium.

codium.cmd:
```@"C:\Users\gigi\Desktop\VSCode-Anywhere\apps\scoop\apps\vscodium\current\bin\codium.cmd"  %*
```
codium.ps1:
```
if (!(Test-Path Variable:PSScriptRoot)) { $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent }
$path = join-path "$psscriptroot" "..\apps\vscodium\current\bin\codium.cmd"
if($myinvocation.expectingInput) { $input | & $path  @args } else { & $path  @args }
```

PR #4543, opened 2021-11-28
feat(shim): Rework shimming logic

### LINKED ISSUE #2018: ffmpeg installation doesn't include 'prores_ks'
When trying to encode using prores_ks as the output codec under ffmpeg installed via scoop, it returns:

`Unable to find a suitable output format for 'prores_ks'`
`prores_ks: Invalid argument`

Manually installing and using the static binary of ffmpeg 3.4.1 from their main site works as expected with ffmpeg_ks.

Any thoughts as to why prores_ks is unsupported in scoop's install of ffmpeg?

Full command in use was:

`ffmpeg -start_number 1001 -f image2 -r 23.98 -i inFile.%04d.dpx -codec:v prores_ks -profile:v 4444 -pix_fmt yuva444p10le -r 23.976 outFile_WinKeepAlpha.mov`

### LINKED ISSUE #3699: Shim generation breaks in Powershell 7 when using powershell.exe as the invoking executable
Example installing the Clojure cli from https://github.com/littleli/scoop-clojure.

Resulting shim when installing under Powershell 5

#### **`clj.ps1`**
```Powershell
$path = "C:\Windows\System32\WindowsPowerShell\v1.0\\powershell.exe"
if($myinvocation.expectingInput) { $input | & $path -NoProfile -ExecutionPolicy Bypass -Command Invoke-Clojure @args } else { & $path -NoProfile -ExecutionPolicy Bypass -Command Invoke-Clojure @args }
```

Resulting shim when installing under Powershell 7

#### **`clj.ps1`**
```Powershell
if (!(Test-Path Variable:PSScriptRoot)) { $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent }
$path = join-path "$psscriptroot" "C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe"
if($myinvocation.expectingInput) { $input | & $path -NoProfile -ExecutionPolicy Bypass -Command Invoke-Clojure @args } else { & $path -NoProfile -ExecutionPolicy Bypass -Command Invoke-Clojure @args }
```

which results in the path concatenating two absolute paths. Resulting error running the clj shim

```
& : The term 'D:\Users\Brian\.scoop\shims\C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At D:\Users\Brian\.scoop\shims\clj.ps1:3 char:129
+ ... nPolicy Bypass -Command Invoke-Clojure @args } else { & $path -NoProf ...
+                                                             ~~~~~
+ CategoryInfo          : ObjectNotFound: (D:\Users\Brian\.sco…v1.0\powershell.exe:String) [], CommandNotFoundException + FullyQualifiedErrorId : CommandNotFoundException
```

### LINKED ISSUE #3749: .ps1 shims breaking applications that call program via start-process
Simple one:
scoop install git
Invoke-Git

`Exception calling "Start" with "0" argument(s): "The specified executable is not a valid application for this OS platform."`

This happens because the program looks for "git" and "git.ps1" is returned as the first result, which is not an exe.

Is there a way to make git.exe the first in the path search? maybe two paths for shims with the .ps1 in a lower order?

Is there a setting to disable .ps1 shim creation? Why do they exist anyways if the .exe shims are created?

### LINKED ISSUE #3795: .ps1 shims cause lots of problems
Are the .ps1 shims really necessary? Aren't they already met by the .cmd and .exe shims?

Often when discovering or trying to run a command, the .ps1 shim gets resolved and breaks. For instance, & (get-command git) you have to know to add -CommandType application to not have it fail.

### LINKED ISSUE #3850: .ps1 shims don't work as expected with PowerShell 7 chain operators
With PowerShell 7 you can use chain operators. Without a proper value in `$?` it does not work as expected.

When I run this command line:
```
PS> should-fail || echo "Failed"
```
I get no output.
With current shims, PowerShell always see success (with automatic variable `$?`) when running .ps1 shims.

As a workaround, I can add .exe to the name of the command.  This will bypass the .ps1 shim and work as expected.
```
PS> should-fail.exe || echo "Failed"
Failed
```

### LINKED ISSUE #3877: Git spawns thousands of times
On two different systems now, git spawns thousands of times bringing the system to a crawl:
```bat
c:>tasklist | grep git.exe | wc -l
  14279
```

Perhaps the issue is that scoop's `apps/shims/git.exe` runs `apps/git/current/cmd/git.exe`, but instead of it running `apps/git/current/mingw64/libexec/git-core/git.exe` it runs `apps/shims/git.exe`  again, in a vicious cycle?

Any idea what exactly the issue is, and how to fix it?

### LINKED ISSUE #3890: ps1 shims for git are reseting $? PowerShell variable
Shims for `git` package are breaking GitLab runner. When runner tries to fetch git repo it checks for exit code of git but shims are resetting `$?` variable.
So even if git command fails `$?` will report no error.

### LINKED ISSUE #4003: .ps1 shim for Git masks command failure ($? is True, $LASTEXITCODE is non-zero)
## What's the issue?

When running a Git command which fails, I want to check for that failure using the `$?` variable.  (More concretely, I'd like to use it implicitly with the `&&` operator added in PowerShell 7.)

This works as expected if the `git` command resolves to `git.exe`, but fails if it resolves to `git.ps1`.

This is surprising behavior, and makes the shims a leaky abstraction over the command they represent.

## Steps to reproduce

- Run `Get-Command git` and verify that the `Name` property is `git.ps1`
- In a directory which is not a Git repository, run a Git command such as `git status`
- Observe that the value of the `$?` variable is `True`
- Remove or rename the `git.ps1` shim
- Run `Get-Command git` and verify that the `Name` property is `git.exe`
- Run the same command as above in the same directory
- Observe that the value of the `$?` variable is `False`

### LINKED ISSUE #4216: .ps1 shim passes incorrect arguments to app
I was calling ffmpeg (without any extension) in PowerShell (the shim directory is in PATH), and noticed ffmpeg gives me error even though the command line was correct in cmd.exe. After some digging, I found that the ffmpeg.ps1 shim is passing the arguments incorrectly. And in PowerShell it prioritizes ffmpeg.ps1 over ffmpeg.exe if they both exist, while it's the other way in cmd.exe.

Consider this command line:
```
test.ps1 -a:b c
```
The expected result is that `test` receives two arguments: `-a:b` and `c`.
The actual result is that `test` receives three arguments: `-a:`, `b` and `c`.
To get the correct result in PowerShell, this has to be used:
```
test.ps1 '-a:b' c
```

I understand this is caused by PowerShell's handling of named arguments (there are quite some [issues](https://github.com/PowerShell/PowerShell/issues/6360) in PowerShell repo), but it manifests in Scoop. I'm sure Scoop can add some special handing in the .ps1 shim, but I don't know if you guys are willing to go that route.

Also, thanks to https://github.com/lukesampson/scoop/issues/1606, I really don't see the point of .ps1 shim. Removing them thus "forcing" PowerShell to use .exe shim actually solves the problem. Is there a way to tell Scoop not to generate .ps1 shim?

### LINKED ISSUE #4333: The kiennq shim should be the default
I commented in #3634 but since it's closed, I thought it deserved some fresh exposure.

I've had trouble with shims leaving blank console windows open for some apps and some particular use cases. PR #3998 adds alternative shims which solve this and many other problems that users have reported. In particular, I have been using the "kiennq" shim with great results. It's working perfectly for all my apps.

On Dec 23, 2020 @rasa [said](https://github.com/lukesampson/scoop/issues/3634#issuecomment-750371232): "Eventually, one of those shims (probably kiennq) will be made the default." There seems to be virtually unanimous agreement that it is the preferred shim among all the options.

Considering the number of issues that it resolves, I think it's more than time for "kiennq" to become the default shim. I'd really prefer not having to change the config on every system I setup just to enable it.

### LINKED ISSUE #4376: Scoop Shims Interfere with Child Process Creation in Node.js
Installing Scoop programs with shims forces two instances to be created when Node.js 'child_process' methods are called regardless of their configuration parameters. For example:

```shell
scoop install ffmpeg
```
Creates ~\scoop\shims\
```
ffmpeg.exe
ffmpeg.ps1
ffmpeg.shim
ffplay.exe
ffplay.ps1
ffplay.shim
```

Writing code:
```js
const child_process = require('child_process');
const ffplay = child_process.spawn('ffplay', ['audio.mp3']);
```
Creates two 'ffplay' processes when only one should have been created.

### LINKED ISSUE #4413: FEATURE: Use existing `sh` script if one exists
VSCode (and possibly other applications) installs a shell script, `code`, with no file extension and a valid shebang. The primary use for this script is for WSL integration.

`scoop` currently can't expose this script, because it generates its own `code` shim that invokes `code.cmd`. This is despite the fact that the manifest for VSCode [includes `code` as a `bin` entry](https://github.com/lukesampson/scoop-extras/blob/master/bucket/vscode.json#L39).

Symlinking the VSCode `code` script _in place of_ the Scoop-generated shim appears to work as expected, both for the standard use-case of the shim (i.e. launching the application from something like Git Bash) and for WSL.

I suggest that when a file with no extension is included in the `bin` entry of a package manifest, a symlink to this file should be generated in place of a normal shim.

### LINKED ISSUE #4473: ffprobe.ps1 can't run a command
I want to use `ffprobe` get Width x Height (resolution) from video
but have error show
```txt
ffprobe -v error -select_streams v:0 -show_entries stream=width,height -of default=nw=1:nk=1 .\123.mp4
Argument '.\123.mp4' provided as input filename, but 'height' was already specified.
```
However use `ffprobe.exe`
```txt
ffprobe.exe -v error -select_streams v:0 -show_entries stream=width,height -of default=nw=1:nk=1 .\123.mp4
1920
1080
```
will be get right.

And `ffprobe -version
```
ffprobe -version
ffprobe version n4.4-7-g023bc5d926 Copyright (c) 2007-2021 the FFmpeg developers
built with gcc 10-win32 (GCC) 20210408
configuration: --prefix=/ffbuild/prefix --pkg-config-flags=--static --pkg-config=pkg-config --cross-prefix=x86_64-w64-mingw32- --arch=x86_64 --target-os=mingw32 --enable-gpl --enable-version3 --disable-debug --disable-w32threads --enable-pthreads --enable-iconv --enable-libxml2 --enable-zlib --enable-libfreetype --enable-libfribidi --enable-gmp --enable-lzma --enable-fontconfig --enable-libvorbis --enable-opencl --enable-libvmaf --enable-vulkan --enable-amf --enable-libaom --enable-avisynth --enable-libdav1d --enable-libdavs2 --enable-ffnvcodec --enable-cuda-llvm --enable-libglslang --enable-libgme --enable-libass --enable-libbluray --enable-libmp3lame --enable-libopus --enable-libtheora --enable-libvpx --enable-libwebp --enable-lv2 --enable-libmfx --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopenjpeg --enable-librav1e --enable-librubberband --enable-schannel --enable-sdl2 --enable-libsoxr --enable-libsrt --enable-libsvtav1 --enable-libtwolame --enable-libuavs3d --enable-libvidstab --enable-libx264 --enable-libx265 --enable-libxavs2 --enable-libxvid --enable-libzimg --extra-cflags=-DLIBTWOLAME_STATIC --extra-cxxflags= --extra-ldflags=-pthread --extra-libs=-lgomp
libavutil      56. 70.100 / 56. 70.100
libavcodec     58.134.100 / 58.134.100
libavformat    58. 76.100 / 58. 76.100
libavdevice    58. 13.100 / 58. 13.100
libavfilter     7.110.100 /  7.110.100
libswscale      5.  9.100 /  5.  9.100
libswresample   3.  9.100 /  3.  9.100
libpostproc    55.  9.100 / 55.  9.100
```

### LINKED ISSUE #4486: Shims for Bash?
When I install vscode, my shims folder gets `code.ps1`, `code.cmd`, and `code`, which are suitable for running from powershell, cmd, and bash. This is great but it seems like VSCode is alone in this. When I install other apps, I get only `xxx.exe`, `xxx.ps1`, and `xxx.shim`. Why this discrepancy? Why is VSCode so special and why don't other apps behave this way?

### LINKED ISSUE #4545: .ps1 shims interfere in PowerShell on WSL
Since Scoop creates ps1 shims for all program files, and ps1 scripts have a greater priority of execution, many times if the same application is installed on Windows and on WSL, using PowerShell on WSL causes the Windows binary to execute. This causes immense slowdowns, because cross-filesystem performance of WSL is extremely slow. Apart from being slow, the output is also wrong since the application reports UNC paths instead of Linux paths.

rashil2000:
The .ps1 shims are now only created when the bin entry itself is a .ps1 script.

Closes #3749 
Closes #2018 
Closes #4486 
Closes #3850
Closes #3795
Closes #4003
Closes #3890
Closes #4216
Closes #4473
Closes #3699
Closes #4413
Closes #4545
Closes #3851

Closes #4333
Closes #3877
Closes #4376

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -646,6 +646,7 @@ function get_shim_path() {
         '71' { $shim_path = "$(versiondir 'scoop' 'current')\supporting\shims\71\shim.exe"; Break }
         'scoopcs' { $shim_path = "$(versiondir 'scoop' 'current')\supporting\shimexe\bin\shim.exe"; Break }
         'rshim' { $shim_path = "$(versiondir 'scoop' 'current')\supporting\shims\rshim\shim.exe"; Break }
+        'kiennq' { Break } # for backward compatibility
         'default' { Break }
         default { warn "Unknown shim version: '$shim_version'" }
     }
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -575,28 +575,35 @@ function shim($path, $global, $name, $arg) {
     } elseif($path -match '\.(bat|cmd)$') {
         # shim .bat, .cmd so they can be used by programs with no awareness of PSH
         warn_on_overwrite "$shim.cmd" $path
-        "@`"$resolved_path`" $arg %*" | out-file "$shim.cmd" -encoding ascii
+        "@rem $resolved_path
+@`"$resolved_path`" $arg %*" | out-file "$shim.cmd" -encoding ascii
 
         warn_on_overwrite $shim $path
-        "#!/bin/sh`nMSYS2_ARG_CONV_EXCL=/C cmd.exe /C `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
+        "#!/bin/sh
+# $resolved_path
+MSYS2_ARG_CONV_EXCL=/C cmd.exe /C `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
     } elseif($path -match '\.ps1$') {
         # if $path points to another drive resolve-path prepends .\ which could break shims
         warn_on_overwrite "$shim.ps1" $path
-        if($relative_path -match "^(.\\[\w]:).*$") {
-            "`$path = `"$path`""                                                                               | out-file "$shim.ps1" -encoding utf8
-            "if(`$myinvocation.expectingInput) { `$input | & `$path $arg @args } else { & `$path $arg @args }" | out-file "$shim.ps1" -encoding utf8 -append
+        $ps1text = if($relative_path -match "^(.\\[\w]:).*$") {
+            "# $resolved_path
+`$path = `"$path`"
+if(`$myinvocation.expectingInput) { `$input | & `$path $arg @args } else { & `$path $arg @args }"
         } else {
             # Setting PSScriptRoot in Shim if it is not defined, so the shim doesn't break in PowerShell 2.0
-            "if (!(Test-Path Variable:PSScriptRoot)) { `$PSScriptRoot = Split-Path `$MyInvocation.MyCommand.Path -Parent }" | out-file "$shim.ps1" -encoding utf8
-            "`$path = join-path `"`$psscriptroot`" `"$relative_path`""                                                      | out-file "$shim.ps1" -encoding utf8 -append
-            "if(`$myinvocation.expectingInput) { `$input | & `$path $arg @args } else { & `$path $arg @args }"              | out-file "$shim.ps1" -encoding utf8 -append
+            "# $resolved_path
+if (!(Test-Path Variable:PSScriptRoot)) { `$PSScriptRoot = Split-Path `$MyInvocation.MyCommand.Path -Parent }
+`$path = join-path `"`$psscriptroot`" `"$relative_path`"
+if(`$myinvocation.expectingInput) { `$input | & `$path $arg @args } else { & `$path $arg @args }"
         }
+        $ps1text | out-file "$shim.ps1" -encoding utf8
 
         # TODO: Find a way to determine if pwsh.exe exists, and use that in the following batch and bash scripts
 
         # make ps1 accessible from cmd.exe
         warn_on_overwrite "$shim.cmd" $path
-        "@echo off
+        "@rem $resolved_path
+@echo off
 setlocal enabledelayedexpansion
 set args=%*
 :: replace problem characters in arguments
@@ -608,19 +615,27 @@ if !args! == !invalid! ( set args= )
 powershell -noprofile -ex unrestricted `"& '$resolved_path' $arg %args%;exit `$lastexitcode`"" | out-file "$shim.cmd" -encoding ascii
 
         warn_on_overwrite $shim $path
-        "#!/bin/sh`npowershell.exe -noprofile -ex unrestricted `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
+        "#!/bin/sh
+# $resolved_path
+powershell.exe -noprofile -ex unrestricted `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
     } elseif($path -match '\.jar$') {
         warn_on_overwrite "$shim.cmd" $path
-        "@java -jar `"$resolved_path`" $arg %*" | out-file "$shim.cmd" -encoding ascii
+        "@rem $resolved_path
+@java -jar `"$resolved_path`" $arg %*" | out-file "$shim.cmd" -encoding ascii
 
         warn_on_overwrite $shim $path
-        "#!/bin/sh`njava -jar `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
+        "#!/bin/sh
+# $resolved_path
+java -jar `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
     } elseif($path -match '\.py$') {
         warn_on_overwrite "$shim.cmd" $path
-        "@python `"$resolved_path`" $arg %*" | out-file "$shim.cmd" -encoding ascii
+        "@rem $resolved_path
+@python `"$resolved_path`" $arg %*" | out-file "$shim.cmd" -encoding ascii
 
         warn_on_overwrite $shim $path
-        "#!/bin/sh`npython `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
+        "#!/bin/sh
+# $resolved_path
+python `"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
     } else {
         warn_on_overwrite "$shim.cmd" $path
         # find path to Git's bash so that batch scripts can run bash scripts
@@ -628,10 +643,13 @@ powershell -noprofile -ex unrestricted `"& '$resolved_path' $arg %args%;exit `$l
         if ($gitdir.FullName -imatch 'mingw') {
           $gitdir = $gitdir.Parent
         }
-        "@`"$(Join-Path (Join-Path $gitdir.FullName 'bin') 'bash.exe')`" `"$resolved_path`" $arg %*" | out-file "$shim.cmd" -encoding ascii
+        "@rem $resolved_path
+@`"$(Join-Path (Join-Path $gitdir.FullName 'bin') 'bash.exe')`" `"$resolved_path`" $arg %*" | out-file "$shim.cmd" -encoding ascii
 
         warn_on_overwrite $shim $path
-        "#!/bin/sh`n`"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
+        "#!/bin/sh
+# $resolved_path
+`"$resolved_path`" $arg `"$@`"" | out-file $shim -encoding ascii
     }
 }
 
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -523,32 +523,24 @@ function movedir($from, $to) {
     }
 }
 
-function get_app_name($path) {
-    if ($path -match '([^/\\]+)[/\\]current[/\\]') {
-        return $matches[1].tolower()
-    }
-    return ''
-}
-
+# deprecated function
 function get_app_name_from_ps1_shim($shim_ps1) {
     if (!(Test-Path($shim_ps1))) {
         return ''
     }
     $content = (Get-Content $shim_ps1 -Encoding utf8) -join ' '
-    return get_app_name $content
+    if ($content -match '([^/\\]+)[/\\]current[/\\]') {
+        return $matches[1].tolower()
+    }
+    return ''
 }
 
-function warn_on_overwrite($shim_ps1, $path) {
-    if (!(Test-Path($shim_ps1))) {
-        return
-    }
-    $shim_app = get_app_name_from_ps1_shim $shim_ps1
-    $path_app = get_app_name $path
-    if ($shim_app -eq $path_app) {
+function warn_on_overwrite($shim_ext, $path) {
+    if (!(Test-Path($shim_ext))) {
         return
     }
     $filename = [System.IO.Path]::GetFileName($path)
-    warn "Overwriting shim to $filename installed from $shim_app"
+    warn "Overwriting already existing shim $filename"
 }
 
 function shim($path, $global, $name, $arg) {
@@ -558,31 +550,15 @@ function shim($path, $global, $name, $arg) {
 
     $shim = "$abs_shimdir\$($name.tolower())"
 
-    warn_on_overwrite "$shim.ps1" $path
-
     # convert to relative path
     Push-Location $abs_shimdir
     $relative_path = resolve-path -relative $path
     Pop-Location
     $resolved_path = resolve-path $path
 
-    # if $path points to another drive resolve-path prepends .\ which could break shims
-    if($relative_path -match "^(.\\[\w]:).*$") {
-        write-output "`$path = `"$path`"" | out-file "$shim.ps1" -encoding utf8
-    } else {
-        # Setting PSScriptRoot in Shim if it is not defined, so the shim doesn't break in PowerShell 2.0
-        Write-Output "if (!(Test-Path Variable:PSScriptRoot)) { `$PSScriptRoot = Split-Path `$MyInvocation.MyCommand.Path -Parent }" | Out-File "$shim.ps1" -Encoding utf8
-        write-output "`$path = join-path `"`$psscriptroot`" `"$relative_path`"" | out-file "$shim.ps1" -Encoding utf8 -Append
-    }
-
-    if($path -match '\.jar$') {
-        "if(`$myinvocation.expectingInput) { `$input | & java -jar `$path $arg @args } else { & java -jar `$path $arg @args }" | out-file "$shim.ps1" -encoding utf8 -append
-    } else {
-        "if(`$myinvocation.expectingInput) { `$input | & `$path $arg @args } else { & `$path $arg @args }" | out-file "$shim.ps1" -encoding utf8 -append
-    }
-
     if($path -match '\.(exe|com)$') {
         # for programs with no awareness of any shell
+        warn_on_overwrite "$shim.exe" $path
         Copy-Item (get_shim_path) "$shim.exe" -force
         write-output "path = $resolved_path" | out-file "$shim.shim" -encoding utf8
         if($arg) {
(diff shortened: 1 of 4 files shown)


lib/core.ps1
```diff
@@ -606,19 +611,40 @@ set invalid=`"='
...
     }
 }
 
 function get_shim_path() {
-    $shim_path = "$(versiondir 'scoop' 'current')\supporting\shimexe\bin\shim.exe"
+    $shim_path = "$(versiondir 'scoop' 'current')\supporting\shims\kiennq\shim.exe"
     $shim_version = get_config 'shim' 'default'
     switch ($shim_version) {
         '71' { $shim_path = "$(versiondir 'scoop' 'current')\supporting\shims\71\shim.exe"; Break }
-        'kiennq' { $shim_path = "$(versiondir 'scoop' 'current')\supporting\shims\kiennq\shim.exe"; Break }
```
rasa, line 621: Need to leave kiennq in, or users who set this, will see error messages, yes?

rashil2000: Fixed

issaclin32 (approved):

rashil2000:
The `scoop which` functionality in this PR is currently broken (it always points to the shim directory for batch and bash scripts), as it always relied on a .ps1 file being present.

@niheaven is there a way to extract the app path from batch and bash scripts (similar to .exe and .ps1 files)?

libexec/scoop-which.ps1
```diff
@@ -19,10 +19,14 @@ $path = "$($gcm.path)"
...
-    $shimtext = Get-Content $path
-
-    $exepath = ($shimtext | Where-Object { $_.startswith('$path') }).split(' ') | Select-Object -Last 1 | Invoke-Expression
+if($path -like "$usershims*" -or $path -like "$globalshims*") {
+    $exepath = if ($path.endswith(".exe")) {
+        (Get-Content "$(join-path ([system.io.fileinfo]$path).DirectoryName ([system.io.fileinfo]$path).BaseName).shim" | Select-Object -First 1).replace('path = ', '')
+    } elseif ($path.endswith(".ps1")) {
+        (Get-Content $path | Where-Object { $_.startswith('$path') }).split(' ') | Select-Object -Last 1 | Invoke-Expression
+    } else {
+        # need to handle .cmd files
```
niheaven, line 28:
```suggestion
        (Select-String -Path $path -Pattern '[''"]([^&=]*?)[''"]').Matches.Groups[1].Value
```

An ugly job but should work.

rashil2000: Hmm, I'm not very strong at Regex but will this work for all types of batch files we create?

We currently have these types:

- https://github.com/ScoopInstaller/Scoop/blob/968e38bad8bd9cf7d1bf33bd0ba4c8b32d3db2f6/lib/core.ps1#L578
- https://github.com/ScoopInstaller/Scoop/blob/968e38bad8bd9cf7d1bf33bd0ba4c8b32d3db2f6/lib/core.ps1#L608
- https://github.com/ScoopInstaller/Scoop/blob/968e38bad8bd9cf7d1bf33bd0ba4c8b32d3db2f6/lib/core.ps1#L614
- https://github.com/ScoopInstaller/Scoop/blob/968e38bad8bd9cf7d1bf33bd0ba4c8b32d3db2f6/lib/core.ps1#L620
- https://github.com/ScoopInstaller/Scoop/blob/968e38bad8bd9cf7d1bf33bd0ba4c8b32d3db2f6/lib/core.ps1#L631

niheaven: I'm not sure if it works for `bash`, but the first four are okay with this regex.

If PATH in `.cmd` is the only one that enclosed in `"` or `'` and doesn't contain `&` and `=`, the regex will extract it.

Examples:

scoop.cmd (excluding `&` and `=` is for such `ps1` script)
```cmd
@echo off
setlocal enabledelayedexpansion
set args=%*
:: replace problem characters in arguments
set args=%args:"='%
set args=%args:(=`(%
set args=%args:)=`)%
set invalid="='
if !args! == !invalid! ( set args= )
powershell -noprofile -ex unrestricted "& 'D:\Scoop\apps\scoop\current\bin\scoop.ps1'  %args%;exit $lastexitcode"
```

idle.cmd
```cmd
@"D:\Scoop\apps\python39\current\Lib\idlelib\idle.bat"  %*
```

shinyproxy.cmd (`java`, `python` is the same, maybe `bash`?)
```cmd
@java -jar "D:\Scoop\apps\shinyproxy\current\shinyproxy.jar"  %*
```

rashil2000: Oops, I forgot to put the bash path in double quotes. Fixed now.

Maybe we can add a negation in the regex so that it doesn't match the path with 'bash.exe' in it?

niheaven: Of course we could, but I think a better way is using `REM xxx` in the first line of every cmd so this function could find it without any additional job.

Since you are modifying shim creation function now, it is just the place to add them.

rashil2000: > a better way is using REM xxx in the first line of every cmd so this function could find it without any additional job

I was thinking of the same thing yesterday!

But I was hesitant to do it because it required modifying all the shims. If this is fine, I will make this change.

niheaven: I think it's fine to add this comment, maybe in `.cmd` first, and then `.ps1`. `.shim` has a simple structure and needn't this.

rashil2000: Done.

rashil2000:
@niheaven this is ready for review

libexec/scoop-which.ps1
niheaven:
```powershell
(Select-String -Path $path -Pattern '^(?:@rem|#)\s*(.*)$').Matches.Groups[1].Value
```

RegEx is `^(?:@rem|#)\s*(.*)$`, use `^` and `$` to find line begin and end. Please try it to see if works.

rashil2000: Working!

rashil2000:
@niheaven this is ready for review.

libexec/scoop-which.ps1
```diff
@@ -19,10 +19,14 @@ $path = "$($gcm.path)"
...
-    $shimtext = Get-Content $path
-
-    $exepath = ($shimtext | Where-Object { $_.startswith('$path') }).split(' ') | Select-Object -Last 1 | Invoke-Expression
+if($path -like "$usershims*" -or $path -like "$globalshims*") {
+    $exepath = if ($path.endswith(".exe") -or $path.endswith(".shim")) {
+        (Get-Content "$(join-path ([system.io.fileinfo]$path).DirectoryName ([system.io.fileinfo]$path).BaseName).shim" | Select-Object -First 1).replace('path = ', '')
+    } elseif ($path.endswith(".ps1")) {
+        (Get-Content $path | Where-Object { $_.startswith('$path') }).split(' ') | Select-Object -Last 1 | Invoke-Expression
+    } else {
+        (Select-String -Path $path -Pattern '^(?:@rem|#)\s*(.*)$').Matches.Groups[1].Value
```
niheaven, line 28:
```suggestion
    } else {
        (Select-String -Path $path -Pattern '^(?:@rem|#)\s*(.*)$').Matches.Groups[1].Value
```

niheaven (commented):
The `ps1` shim also has PATH comment, right?

libexec/scoop-which.ps1
rashil2000: Yes, but the PATH comment won't be there in older existing ps1 shims, so I have kept this statement for backward compatibility

lib/core.ps1
```diff
@@ -604,22 +610,64 @@ set args=%args:(=``(%
...
+        'scoopcs' { $shim_path = "$(versiondir 'scoop' 'current')\supporting\shimexe\bin\shim.exe"; Break }
         'rshim' { $shim_path = "$(versiondir 'scoop' 'current')\supporting\shims\rshim\shim.exe"; Break }
+        'kiennq' { Break } # for backward compatibility
```
rasa, line 670: If the user has kiennq set, he will get scoop's default shim, not kiennq, now?

rashil2000: No, if the user has kiennq set, it will use the path `"$(versiondir 'scoop' 'current')\supporting\shims\kiennq\shim.exe"` (and come out of the switch case)

libexec/scoop-which.ps1
rashil2000: Done

libexec/scoop-which.ps1
```diff
@@ -19,10 +19,15 @@ $path = "$($gcm.path)"
...
-
-    $exepath = ($shimtext | Where-Object { $_.startswith('$path') }).split(' ') | Select-Object -Last 1 | Invoke-Expression
+if($path -like "$usershims*" -or $path -like "$globalshims*") {
+    $exepath = if ($path.endswith(".exe") -or $path.endswith(".shim")) {
+        (Get-Content "$(join-path ([system.io.fileinfo]$path).DirectoryName ([system.io.fileinfo]$path).BaseName).shim" | Select-Object -First 1).replace('path = ', '')
+    } else {
+        ((Select-String -Path $path -Pattern '^(?:@rem|#)\s*(.*)$').Matches.Groups | Select-Object -Index 1).Value
+    }
+    if (!$exepath) {
+        $exepath = (Get-Content $path | Where-Object { $_.startswith('$path') }).split(' ') | Select-Object -Last 1 | Invoke-Expression
```
niheaven, line 29:
```suggestion
        $exepath = (Select-String -Path "shinyproxy.cmd" -Pattern '[''"]([^@&]*?)[''"]' -AllMatches).Matches.Groups[-1].Value | Convert-Path
```

libexec/scoop-which.ps1
```diff
@@ -19,10 +19,15 @@ $path = "$($gcm.path)"
...
 
-if($path.endswith(".ps1") -and ($path -like "$usershims*" -or $path -like "$globalshims*")) {
-    $shimtext = Get-Content $path
-
-    $exepath = ($shimtext | Where-Object { $_.startswith('$path') }).split(' ') | Select-Object -Last 1 | Invoke-Expression
+if($path -like "$usershims*" -or $path -like "$globalshims*") {
+    $exepath = if ($path.endswith(".exe") -or $path.endswith(".shim")) {
+        (Get-Content "$(join-path ([system.io.fileinfo]$path).DirectoryName ([system.io.fileinfo]$path).BaseName).shim" | Select-Object -First 1).replace('path = ', '')
+    } else {
+        ((Select-String -Path $path -Pattern '^(?:@rem|#)\s*(.*)$').Matches.Groups | Select-Object -Index 1).Value
```
niheaven, line 26: Why use `Select-Object` while a simple `[]` could work?

libexec/scoop-which.ps1
```diff
@@ -19,10 +19,15 @@ $path = "$($gcm.path)"
 $usershims = "$(resolve-path $(shimdir $false))"
 $globalshims = fullpath (shimdir $true) # don't resolve: may not exist
 
-if($path.endswith(".ps1") -and ($path -like "$usershims*" -or $path -like "$globalshims*")) {
-    $shimtext = Get-Content $path
-
-    $exepath = ($shimtext | Where-Object { $_.startswith('$path') }).split(' ') | Select-Object -Last 1 | Invoke-Expression
+if($path -like "$usershims*" -or $path -like "$globalshims*") {
+    $exepath = if ($path.endswith(".exe") -or $path.endswith(".shim")) {
+        (Get-Content "$(join-path ([system.io.fileinfo]$path).DirectoryName ([system.io.fileinfo]$path).BaseName).shim" | Select-Object -First 1).replace('path = ', '')
```
niheaven, line 24:
```suggestion
        ([Array](Get-Content ($path -replace '.exe$', '.shim')))[0].replace('path = ', '')
```

niheaven (changes requested):

rashil2000:
`Select-Object` is used because it can fail gracefully.

Using `[]` will throw an error if the index (or the array itself) doesn't exist (which will be the case for older existing ps1 shims), and then we'll have to use a try-catch block or something other to first detect the error and then write a fallback.

In case of `Select-Object`, we can directly use a fallback (as is done here).

niheaven:
I've changed two expression that use regex to clarify what we really want.

rashil2000:
Done

niheaven:
Could I commit to fix test and add some tweaks (mainly cmdlet name)

rashil2000:
Sure!

Which tests are failing?

niheaven:
You have wrong name for `get_app_name_from_ext_shim` 😄 

So change it to `get_app_name_from_shim` will not be ambiguous I thought.

niheaven:
I make some change to `warn_on_overwrite` since it will produce:

image

Now it will be:

image

rashil2000:
Okay, looks good to me!

niheaven (approved):

niheaven:
Let's begin testing it!

chawyehsu:
you shouldn't close those issues when the fix hasn't been shipped to master...

rashil2000:
I closed them with the idea that people can test the develop branch if they want to

chawyehsu:
You can notice them by just leaving a comment instead of directly closing the issue which is not fixed “officially”...

adamency:
How is this related to [#2230](https://github.com/ScoopInstaller/Scoop/issues/2230) ?

rashil2000:
It concerns shims

niheaven:
Does this fix #2230 and #3915? What should we do now?

rashil2000:
> Does this fix #2230 and #3915? What should we do now?

I think that problem can be solved by using relative paths in batch and bash shims too (ps1 shim already uses relative paths). But it will require careful testing, and rework of scoop-which also.

PR #4555, opened 2021-12-02
feat(shim): ps1 shim hotfix

rashil2000:
`pwsh` has different argument parsing rules than `powershell`, so we need to specify the -Command parameter.


rashil2000:
I am merging this because without this, scoop command will fail in cmd shell and anyone using develop branch will face errors.

niheaven:
You should always use `Squash and merge` to simplify git log.

rashil2000:
Will do from next time 😅

PR #4579, opened 2021-12-14
chore(config): Document all configuration options

rashil2000:
#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ You can tweak the following `aria2` settings with the `scoop config` command:
 - [aria2-split](https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-s) (default: 5)
 - [aria2-max-connection-per-server](https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-x) (default: 5)
 - [aria2-min-split-size](https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-k) (default: 5M)
+- [aria2-options](https://aria2.github.io/manual/en/html/aria2c.html#options) (default: )
 
 ## Inspiration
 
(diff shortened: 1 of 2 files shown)


rashil2000:
Let me know if I missed any setting

issaclin32 (approved):

niheaven:
```yml
force-update: Force apps updating to bucket's version
shim: Choose scoop shim builds
rootPath: Root path of scoop, default '$env:USERPROFILE\scoop'
globalPath: Global root path of scoop, default '$env:ProgramData\scoop'
cachePath: Cache path of scoop, default '$rootPath\cache'
```

rashil2000:
Can we actually use variables in the config values? Will scoop do a string substitution on them?

niheaven:
Environment Variables could be used, but other local variables should not.

rashil2000:
Then `$rootPath\cache` shouldn't be used, right?

niheaven:
Yes, it is the default value, but user cannot set as is.

rashil2000:
Right. 
So if we show 

```
# cachePath: $rootPath\cache
```

in the help doc, users will think it's okay to use variables.

So we should either leave it empty (and write default value in description), or write an example valid path here.

niheaven:
Use `'cache' folder under Scoop root directory`?

rasa (approved):

rashil2000:
@niheaven done

rashil2000:
Should we also document environment variables used by scoop? Like SCOOP_HOME and SCOOP_CHECKVER_TOKEN. Are there any more?

niheaven:
In wiki or homepage, IMO.

`SCOOP`, `SCOOP_GLOBAL`, `SCOOP_CACHE`, read before `rootPath`, `globalPath`, `cahePath` config.

rashil2000:
Updated the Readme 👍

niheaven (approved):

Issue #4569, opened 2021-12-09
decompress error : no error messages in the 7zip.log file


ViCrack:
## Bug Report

I have the same problem.  

ScoopInstaller/Extras#7425
ScoopInstaller/Extras#7310
ScoopInstaller/Main#2990
ScoopInstaller/Main#3042

### Scoop output

```
Loading tabby-1.0.166-setup.exe from cache.
Checking hash of tabby-1.0.166-setup.exe ... ok.
Extracting dl.7z ... done.
Running pre-install script...
ERROR Exit code was 2!
Failed to extract files from D:\Programs\Scoop\apps\tabby\1.0.166\$PLUGINSDIR\app-64.7z.
Log file:
  D:\Programs\Scoop\apps\tabby\1.0.166\$PLUGINSDIR\7zip.log

Please try again or create a new issue by using the following link and paste your console output:
https://github.com/ScoopInstaller/Extras/issues/new?title=tabby%401.0.166%3a+decompress+error
```

### 1- 7zip.log

```
7-Zip 21.04 beta (x64) : Copyright (c) 1999-2021 Igor Pavlov : 2021-11-02

Scanning the drive for archives:
1 file, 77917405 bytes (75 MiB)

Extracting archive: D:\Programs\Scoop\apps\tabby\1.0.166\$PLUGINSDIR\app-64.7z
--
Path = D:\Programs\Scoop\apps\tabby\1.0.166\$PLUGINSDIR\app-64.7z
Type = 7z
Physical Size = 77917405
Headers Size = 22337
Method = LZMA2:20 LZMA:20 BCJ2
Solid = -
Blocks = 1591


Sub items Errors: 1

Archives with Errors: 1

Sub items Errors: 1
```

### Possible Solution

I modified the code and tried again.

```diff
diff --git a/lib/core.ps1 b/lib/core.ps1
index c6797025..c0721139 100644
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -434,7 +434,8 @@ function Invoke-ExternalCommand {
         return $false
     }
     if ($LogPath -and ($FilePath -notmatch '(^|\W)msiexec($|\W)')) {
-        Out-File -FilePath $LogPath -Encoding ASCII -Append -InputObject $Process.StandardOutput.ReadToEnd()
+        Out-File -FilePath $LogPath -Encoding UTF8 -Append -InputObject $Process.StandardOutput.ReadToEnd()
[...]

PR #4582, opened 2021-12-14
fix(decompress): Fix 'Expand-7zipArchive()' for nested archive

niheaven:
After we added error log in 7zip.log (#4570), nested archive extraction by 7-zip had been broken... This PR fix it.

Besides, output and error logs have some encoding error as in https://github.com/ScoopInstaller/Scoop/issues/4569, and change encoding to `Default` solve my problem (CP936, in PS 7 and PS 5), please test it in different codepage and PowerShell version.

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -434,8 +434,8 @@ function Invoke-ExternalCommand {
         return $false
     }
     if ($LogPath -and ($FilePath -notmatch '(^|\W)msiexec($|\W)')) {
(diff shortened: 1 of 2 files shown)


rashil2000:
I was facing errors with extracting tar.xz URL yesterday and wasted a lot of time... Had no idea it was related to redirecting stderr

rashil2000 (approved):

PR #4585, opened 2021-12-15
tests: Make test for files to end with a newline compatible with UTF-16

tukanos:
The import file test will fail if the file tested is `UTF-16` (@Windows called `Unicode`).

This test is modified to meet it purpose, testing if the file imported ends with new line, but also to work with `UTF-16`.


rashil2000:
Combine this with #4586

PR #4586, opened 2021-12-15
tests: Registry files should be always UTF-16LE (UCS-2 LE BOM) encoded.

tukanos:
Even thou registry files encoded UTF-8 and UTF-8 BOM works for now, it
may change in the future. To prevent some weird errors when using registry
files enforce the usage of UTF-16LE (UCS-2 LE BOM) encoding.

For more see: https://gist.github.com/SalviaSage/8eba542dc27eea3379a1f7dad3f729a0

Closes https://github.com/ScoopInstaller/Extras/issues/7546


rashil2000:
- Please rebase the target to `develop` branch (same for other PR)
- Won't this cause tests to fail for all the existing reg files in the buckets? (same for other PR)

tukanos:
@rashil2000 ups, the commit was supposed to target `develop`.   That was a mistake, the other PR does correctly target it.

Yes, this will fail all existing reg files if they are not correctly encoded.

rashil2000:
> Yes, this will fail all existing reg files if they are not correctly encoded.

Then this is a problem. Why not accommodate both encodings?

Even if we were to recode all the reg files, we control only the Main, Extras and Versions buckets. What about hundreds of buckets out there?

tukanos:
> > Yes, this will fail all existing reg files if they are not correctly encoded.
> 
> Then this is a problem. Why not accommodate both encodings?
> 
> Even if we were to recode all the reg files, we control only the Main, Extras and Versions buckets. What about hundreds of buckets out there?

True.  What about a solution, where I would accommodate UTF8 with BOM/without BOM, but display a warning that registry files should be converted to UTF-16?

rashil2000:
Sure 👍

tukanos:
I have added a modification where all the unicode files (UTF8, 16, 32) with BOM are recognized.  Tell user if the registry file(s) is correctly encoded, if not which encoding should be used.  In case there is no BOM be permissive and try to continue.

rasa:
Where is it documented .reg files must be UTF16? I have only ever created or exported UTF8, and never experienced problems

tukanos:
> Where is it documented .reg files must be UTF16? I have only ever created or exported UTF8, and never experienced problems

You will never see something like *must* be UTF16, it is "only" preferred and recommended, which usually is good idea to follow.  The UTF-8 with or without BOM work for now due to backward compatibility with older Windows.

For more, please read my post  - [registry UTF16](https://github.com/ScoopInstaller/Extras/issues/7546#issuecomment-996035148)

rashil2000:
@niheaven thoughts on this (and #4585)?

niheaven:
@tukanos Could you please merge #4585 and #4586 into one PR, since they together solve one thing?

niheaven:
> @niheaven thoughts on this (and #4585)?

Sound good, but should be tested thoroughly, maybe in `devlop' branch?

tukanos:
I have closed this PR and reopened PR that merges correctly into develop branch.  Both tests: for endlines and registry are merge too.

[Merged PR(s) for tests Registry files should be always UTF-16LE and newline compatible with UTF-16](https://github.com/ScoopInstaller/Scoop/pull/4654)

Issue #4574, opened 2021-12-10
[Bug] Unhandled Exception while searching 


dakehero:
## Bug Report

**Package Name:** [name of package which has bug(s)]
nanazip

### Current Behaviour

*No matches found* and exception message thrown by `Invoke-WebRequest`

### Expected Behaviour
no exception message

### Additional context/output
![image](https://user-images.githubusercontent.com/23186318/145511332-0d691c0a-26b6-4d8f-ba22-a3f76a321da9.png)

### Possible Solution
use `try-catch-finally` expression to handle it

### System details

**Windows version:** [e.g. 7, 8, 10]
Windows 11 21H2

**OS architecture:** [e.g. 32bit, 64bit]
64bit

**PowerShell version:** [output of `"$($PSVersionTable.PSVersion)"`]
7.2.0
**Additional software:** [(optional) e.g. ConEmu, Git]
Windows Terminal

rashil2000:
Looks like a bug. 

Can you make a PR?

niheaven:
This will be occurred when you cannot access github.com, and existed for a long time.

rashil2000:
I can access github.com, all scoop functions work fine, but I can still reproduce this error

FDUZS:
same here.

<img width="938" alt="8qkvTFektI" src="https://user-images.githubusercontent.com/20297872/145662369-5d2f399b-ff85-4d98-8ec8-283933255b9a.png">

Mouriya-Emma:
same problam .i can access github.com and implement scoop update.

![Snipaste_2021-12-13_10-48-35.png](https://s2.loli.net/2021/12/13/GgYs3MtmCcrn1Ql.png)

PR #4587, opened 2021-12-15
Remove nightlies bucket, the repo no longer exists

Hamled:
#### Description
This removes the "nightlies" bucket from `buckets.json` as the repo associated with it (https://github.com/ScoopInstaller/Nightlies) no longer exists.

#### Motivation and Context
This was resulting in an error message from the `Invoke-WebRequest` cmdlet being displayed to when `scoop search` was used (and possibly other commands).
Relates to #4574

I've verified that when this entry is no longer in `buckets.json` the `scoop search` command does not produce an error related to requesting that repository.

Issue #4574 should still be fixed separately however, as there are other sources of potential failure by the `Invoke-WebRequest` cmdlet that should be handled.

### DIFF
--- a/buckets.json
+++ b/buckets.json
@@ -2,7 +2,6 @@
     "main": "https://github.com/ScoopInstaller/Main",
     "extras": "https://github.com/ScoopInstaller/Extras",
     "versions": "https://github.com/ScoopInstaller/Versions",
-    "nightlies": "https://github.com/ScoopInstaller/Nightlies",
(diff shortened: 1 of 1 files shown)


rashil2000:
Already done https://github.com/ScoopInstaller/Scoop/commit/48b035d7f99baa2e81d87ead4ff03a9594e49c3d

Wait for a few days for new release

Hamled:
My bad, I was looking at the `master` branch instead of `develop` :smile:

PR #4619, opened 2022-01-03
fix(autoupdate): Allow checksum file that contains whitespaces

issaclin32:
This allows checksum file that contains hash and **nothing else but whitespaces** to be recognized.

example: [SHA-256 checksum for KDEConnect](https://binary-factory.kde.org/view/Windows%2064-bit/job/kdeconnect-kde_Release_win64/lastSuccessfulBuild/artifact/kdeconnect-kde-21.12.0-835-windows-msvc2019_64-cl.7z.sha256).

related PR: https://github.com/ScoopInstaller/Extras/pull/7662


lib/autoupdate.ps1
```diff
@@ -51,7 +51,7 @@ function find_hash_in_textfile([String] $url, [Hashtable] $substitutions, [Strin
     }
 
     if ($regex.Length -eq 0) {
-        $regex = '^([a-fA-F0-9]+)$'
+        $regex = '^\s*([a-fA-F0-9]+)\s*$'
```
rashil2000, line 54: Isn't this equivalent to just `([a-fA-F0-9]+)`?

issaclin32: I think `^$` exists for some reason.

for example, if the developer releases the checksum with the archive file name, then, the hash will not be catched by this line, but rather by https://github.com/ScoopInstaller/Scoop/blob/master/lib/autoupdate.ps1#L78

This will help avoid errors in some cases, **e.g.** an outdated checksum file.

rashil2000: I just tried your example but it doesn't seem to work - https://regex101.com/r/FmSsBb/1

Am I testing it wrong?

issaclin32: > I just tried your example but it doesn't seem to work - https://regex101.com/r/FmSsBb/1
> 
> Am I testing it wrong?

Because this is not supposed to be matched by `^\s*([a-fA-F0-9]+)\s*$`, 
but rather by `([a-fA-F0-9]{32,128})[\x20\t]+.*`$basename(?:[\x20\t]+\d+)?` [at line 78](https://github.com/ScoopInstaller/Scoop/blob/master/lib/autoupdate.ps1#L78)

rashil2000: Ah okay, I got confused. 

But I have one more confusion. Without the `^$` too, it seems to match correctly (notice the extra whitespace):

image

issaclin32: Oh. I probably didn't explain it clearly enough.

I think the design philosophy of **autoupdate.ps1** is:

(1) Match "*hash*" if the file only contains hash string
       -> this is what I am trying to fix
(2) Convert if the file only contains **Base64** encoded string
(3) Match "filename: *hash*"
      -> This provides an extras check, thus avoiding some problems like an outdated hash file.
(4) Match "hash: *hash*"

This is why we should not use just `([a-fA-F0-9]+)` in step (1).

rashil2000: Okay, that makes sense. Thanks for the explanation!

rashil2000 (dismissed):

niheaven:
Please update `CHANGELOG.md`

niheaven (approved):

niheaven (approved):

PR #4621, opened 2022-01-05
feat(download): Allow downloading without installing

### LINKED ISSUE #1804: Download command proposal

rashil2000:
#### Description

Useful for:
- **Users:** to pre-populate the cache:
  - in case they need to install something later (i.e. no internet scenario)
  - in case they want to inspect the installer/archive for an application (without installing)
- **Maintainers:** to manually verify:
  - hash check failures
  - 404 errors (e.g. with aria2, sourceforge etc.)

The second scenario is perhaps the most useful personally because many people open issues about download failures, but I don't really want to install the particular app. So I can just do `scoop download <app>` to rectify the issue.

#### Motivation and Context

Closes #1804

Relates to #2293

1. Simple test: 
image

2. Above test with pre-filled cache: 
image

3. Test 1 with 32bit download: 
image

4. Test 3 with skipping hash check: 
image

5. Test 4 with force download (overwrite cache): 
image

6. Test with one app with unsupported architecture: 
image

7. Test 6 with aria2: 
image

8. Test with pre-cached install: 
image

#### Checklist:

### DIFF
--- a/libexec/scoop-download.ps1
+++ b/libexec/scoop-download.ps1
@@ -22,7 +22,7 @@
 reset_aliases
 
 $opt, $apps, $err = getopt $args 'fsa:' 'force', 'skip', 'arch='
-if ($err) { error "scoop install: $err"; exit 1 }
+if ($err) { error "scoop download: $err"; exit 1 }
 
 $check_hash = !($opt.s -or $opt.skip)
 $use_cache = !($opt.f -or $opt.force)
--- a/libexec/scoop-download.ps1
+++ b/libexec/scoop-download.ps1
@@ -1,5 +1,5 @@
 # Usage: scoop download <app> [options]
-# Summary: Download apps in the cache folder
+# Summary: Download apps in the cache folder and verify hashes
 # Help: e.g. The usual way to download an app, without installing it (uses your local 'buckets'):
 #      scoop download git
 #
@@ -11,7 +11,7 @@
 #
 # Options:
 #   -f, --force               Force download (overwrite cache)
-#   -s, --skip                Skip hash validation (use with caution!)
+#   -h, --no-hash-heck        Skip hash verification (use with caution!)
 #   -a, --arch <32bit|64bit>  Use the specified architecture, if the app supports it
 
 . "$psscriptroot\..\lib\manifest.ps1"
@@ -21,10 +21,10 @@
 
 reset_aliases
 
-$opt, $apps, $err = getopt $args 'fsa:' 'force', 'skip', 'arch='
+$opt, $apps, $err = getopt $args 'fha:' 'force', 'no-hash-check', 'arch='
 if ($err) { error "scoop download: $err"; exit 1 }
 
-$check_hash = !($opt.s -or $opt.skip)
+$check_hash = !($opt.h -or $opt.'no-hash-check')
 $use_cache = !($opt.f -or $opt.force)
 $architecture = default_architecture
 try {
@@ -49,7 +49,7 @@ foreach ($curr_app in $apps) {
     $app, $bucket, $version = parse_app $curr_app
     $app, $manifest, $bucket, $url = Find-Manifest $app $bucket
 
-    info "Starting download for $app"
+    info "Starting download for $app..."
 
     # Generate manifest if there is different version in manifest
     if (($null -ne $version) -and ($manifest.version -ne $version)) {
@@ -66,7 +66,7 @@ foreach ($curr_app in $apps) {
         continue
     }
     $version = $manifest.version
-    if(!$version) { 
+    if(!$version) {
         error "Manifest doesn't specify a version."
         continue
     }
--- a/libexec/scoop-download.ps1
+++ b/libexec/scoop-download.ps1
@@ -1,0 +1,127 @@
+# Usage: scoop download <app> [options]
+# Summary: Download apps in the cache folder
+# Help: e.g. The usual way to download an app, without installing it (uses your local 'buckets'):
+#      scoop download git
+#
+# To download an app from a manifest at a URL:
+#      scoop download https://raw.githubusercontent.com/ScoopInstaller/Main/master/bucket/runat.json
+#
+# To download an app from a manifest on your computer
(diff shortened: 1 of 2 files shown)


rashil2000:
Example: https://github.com/ScoopInstaller/Extras/issues/7679

niheaven:
`bin\checkurls.ps1` and `bin\checkhashes.ps1` are helper functions that work for these situation, so why not modify `checkhashes`? Since `checkhashes` will download the file into cache folder with version `HASH_CHECK` likes: `cpu-z#HASH_CHECK#https_download.cpuid.com_cpu-z_cpu-z_1.99-en.zip`, add a switch to let `checkhashes` correctly rename versioned cache may be useful.

BTW, `bin\checkhashes.ps1` is not ported to bucket repos now, it needs to be added.

rashil2000:
But we cannot use `checkhashes` as a sub-command of `scoop`, right?

I wanted to make this utility for both users and maintainers. Users might not be able to/comfortable in running a specific utility.

libexec/scoop-download.ps1
```diff
@@ -0,0 +1,127 @@
+# Usage: scoop download <app> [options]
+# Summary: Download apps in the cache folder
```
niheaven, line 2:
```suggestion
# Summary: Download apps in the cache folder and do hash(es) check.
```

libexec/scoop-download.ps1
```diff
@@ -0,0 +1,127 @@
...
+    $architecture = ensure_architecture ($opt.a + $opt.arch)
+} catch {
+    abort "ERROR: $_"
+}
+
+if (!$apps) { error '<app> missing'; my_usage; exit 1 }
+
+if (is_scoop_outdated) {
+    scoop update
+}
```
niheaven, line 40: Maybe not force scoop update is better?

libexec/scoop-download.ps1
```diff
@@ -0,0 +1,127 @@
...
+if(!$use_cache) { warn "Cache is being ignored." }
+
+foreach ($curr_app in $apps) {
+    # Prevent leaking variables from previous iteration
+    $bucket = $version = $app = $manifest = $url = $null
+
+    $app, $bucket, $version = parse_app $curr_app
+    $app, $manifest, $bucket, $url = Find-Manifest $app $bucket
+
+    info "Starting download for $app"
```
niheaven, line 52:
```suggestion
    info "Starting download for $app..."
```

libexec/scoop-download.ps1
```diff
@@ -0,0 +1,127 @@
...
+                    error (new_issue_msg $app $bucket "hash check failed")
+                    continue
+                }
+            } else {
+                info "Skipping hash verification."
+            }
+        }
+    }
+
+    success "'$app' ($version) was downloaded successfully!"
```
niheaven, line 124:
```suggestion
    success "'$app' ($version) was downloaded and hash checked successfully!"
```

rashil2000: We have a flag for skipping hash check, so this will not be always true.

libexec/scoop-download.ps1
rashil2000: Should I add an additional flag `-u, --update => update scoop before downloading`?

niheaven: Maybe `--update-scoop`?

libexec/scoop-download.ps1
niheaven: Use an `if` to conditionally output different statement? Likes:

```pwsh
if ($curr_check_hash) {
    success "'$app' ($version) was downloaded and hash checked successfully!"
} else {
    success "'$app' ($version) was downloaded successfully!"
}
```

And IMO `--no-hash-check` is a better switch name. (instead of simple `--skip` which may indicate skip downloading.

rashil2000: > Use an `if` to conditionally output different statement? Likes:

The information whether hash is being checked is already given by the `check_hashes` function, see: 
image

We should show it only once.

> And IMO `--no-hash-check` is a better switch name. (instead of simple `--skip` which may indicate skip downloading.

Will change it

rashil2000: Also, the primary function of this command is to just `download` - hash check is a supplementary internal addition, useful for maintainers but not necessarily users.

libexec/scoop-download.ps1
rashil2000: I have gone with the default behaviour of `scoop install` (enforce update). The flag allows to skip it.

libexec/scoop-install.ps1
```diff
@@ -49,7 +50,7 @@ function is_installed($app, $global) {
     return $false
 }
 
-$opt, $apps, $err = getopt $args 'gfiksa:' 'global', 'force', 'independent', 'no-cache', 'skip', 'arch='
```
rashil2000, line 52: `force` flag was unused here, so removed it.

niheaven:
Why not separate these two features into two PR? `scoop-download` is here, while `--no-update-scoop` in another?

rashil2000:
The new flag change was quite small and related to downloading itself, so I put it here.

I can make a new PR if you really insist.

niheaven:
Yes, I think changes in `scoop-install` and `scoop-vitustotal` should be separated, that will make changing more clear, right?

rashil2000:
Sure, will do

rashil2000:
Created #4634

niheaven (approved):
The short params `-h` and `-u` seem to be confused, but I cannot figure out what will be better...

rashil2000:
> The short params `-h` and `-u` seem to be confused, but I cannot figure out what will be better...

Haha yeah, we'll update it if we find more suitable short forms

PR #4643, opened 2022-01-11
feat(ci): Setup github actions as test ci

chawyehsu:
#### Description
Setup Github Actions CI

#### Motivation and Context
Migration from appveyor

~demo: https://github.com/chawyehsu/scoop/actions/runs/1682023136~

https://github.com/ScoopInstaller/Scoop/actions/runs/1682034804

#### Checklist:


chawyehsu:
Might break downstream bucket tests that still use appveyor, investigating...

rashil2000:
> Might break downstream bucket tests that still use appveyor, investigating...

We will migrate all our buckets CI to GH Actions too. 

But for third-party buckets, instead of removing Appveyor-specific code, why not put both? We could detect if the CI is Appveyor or GitHub (or third case, local installation) and run stuff conditionally. This way third-party bucket maintainers won't have to migrate.

rashil2000:
Also I think `.github/workflows/ci.yml` would be a more precise name.

chawyehsu:
Replaced by #4655

PR #4644, opened 2022-01-11
fix(core): Ensure manipulating config with UTF8 encoding

### LINKED ISSUE #4629: Aria2 and git does not support non-ASCII paths
Errors will occur when aria2 is enabled, and `SCOOP_DIR` contains non-ASCII chars (for example, `C:\Users\小李\scoop`). Aria2 will save the file in a **gibberish/[mojibake](https://en.wikipedia.org/wiki/Mojibake)** path such as `C:\Users\�������������\scoop`.

upstream discussion: https://github.com/aria2/aria2/issues/1417

We cannot expect **aria2** to fix this problem in a short time, but there are 2 solutions I can think of:

**(1) Ban aria2 if it is the case**
If `SCOOP_DIR` contains non-ASCII chars. Then:
Set `aria2-enabled` to false, and forbid setting it to True (also give user a message like `aria2 cannot be enabled because scoop path 'C:\Users\小李\scoop' contains non-ASCII characters`

or, **(2) Create junctions** 
If `SCOOP_DIR` contains non-ASCII chars, then:
Convert the path with **URL encode**:
`[System.Web.HttpUtility]::UrlEncode('C:\Users\小李\scoop\cache')` -> This will equal to `C%3a%5cUsers%5c%e5%b0%8f%e6%9d%8e%5cscoop%5ccache`

Create a junction `C:\users\Public\scoop-cache-C%3a%5cUsers%5c%e5%b0%8f%e6%9d%8e%5cscoop%5ccache` that points to `C:\Users\小李\scoop\cache`.

After doing this, we can set Aria2 to save the file in `C:\users\Public\scoop-cache-C%3a%5cUsers%5c%e5%b0%8f%e6%9d%8e%5cscoop%5ccache`, and it should work fine.

Asking for your opinions:
@rasa @niheaven @chawyehsu @rashil2000

chawyehsu:
#### Description
Use utf8 encoding in scoop's and aria2's config

#### Motivation and Context

Fixes #4629

[1]: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/set-content?view=powershell-5.1
[2]: https://docs.microsoft.com/en-us/dotnet/api/system.io.file.writealltext

New unit tests introduced.

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 - **autoupdate:** Allow checksum file that contains whitespaces ([#4619](https://github.com/ScoopInstaller/Scoop/issues/4619))
 - **autoupdate:** Rename $response to $res ([#4706](https://github.com/ScoopInstaller/Scoop/issues/4706))
+- **config:** Ensure manipulating config with UTF8 encoding ([#4644](https://github.com/ScoopInstaller/Scoop/issues/4644))
 - **config:** Allow scoop config use Unicode characters ([#4631](https://github.com/ScoopInstaller/Scoop/issues/4631))
 - **config:** Fix `set_config` bugs ([#3681](https://github.com/ScoopInstaller/Scoop/issues/3681))
 - **current:** Remove 'current' while it's not a junction ([#4687](https://github.com/ScoopInstaller/Scoop/issues/4687))
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -41,9 +41,9 @@ function load_cfg($file) {
     }
 
     try {
-        # ReadAllText will detect the encoding of the file automatically
-        # Ref: https://docs.microsoft.com/en-us/dotnet/api/system.io.file.readalltext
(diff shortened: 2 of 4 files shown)


niheaven (commented):
Maybe `New-Item` is a better choice.

I found in its docs that: ([`-ItemType File`](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/new-item?view=powershell-5.1#parameters))

> When you create a file using this method, the resulting file is encoded as UTF-8 without a byte-order-mark (BOM).

niheaven:
I'll check it tomorrow.

niheaven:
@chawyehsu It is a bug of Pester/PS5 when test file contains non-ASCII characters. Using `[Regex]::Unescape` to avoid it.

Although change `[System.IO.File]::ReadAllText` to `[System.IO.File]::ReadAllLines` to keep consistence.

niheaven (approved):

chawyehsu (commented):
LGTM

PR #4671, opened 2022-01-17
build(ci): Use cache in GitHub Actions

niheaven:
#### Description

Use ~https://github.com/actions/cache~ https://github.com/potatoqualitee/psmodulecache to cache PowerShell modules in GitHub Actions. (Really not happy while working with https://github.com/actions/cache)

#### Motivation and Context

PowerShell modules installation is the most time-consuming step of CI test, this fix tries to skip them.

It also skips modules installation if there're ones.

~I don't find an effective way to test the PR, but it seems to work~

~From CI results below, it found cache, but doesn't use them. Maybe a little more tweak are needed.~

Use https://github.com/potatoqualitee/psmodulecache to cache PowerShell Modules, it works fine as this PR's two commits.

#### Checklist:

- [ ] ~I have updated the tests accordingly.~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: Scoop Core CI Tests
 
 on:
   pull_request:
@@ -14,11 +14,14 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 2
-      - name: Init and Test
+      - name: Init Test Suite
+        uses: potatoqualitee/psmodulecache@v4
+        with:
+          modules-to-cache: PSScriptAnalyzer, BuildHelpers, Pester:4.10.1
+          shell: powershell
+      - name: Test Scoop Core
         shell: powershell
-        run: |
-          .\test\bin\init.ps1
-          .\test\bin\test.ps1
+        run: ./test/bin/test.ps1
   test_pwsh:
     name: PowerShell
     runs-on: windows-latest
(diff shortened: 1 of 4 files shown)


niheaven:
Should we remove CI running trigger by `push` if every change in Scoop Core must submit a PR? For this one, GHA runs four tests of which two are unneeded.

rashil2000:
> For this one, GHA runs four tests of which two are unneeded.

I think this happens only when a PR is opened from Scoop core itself. If someone from outside forks and sends a PR it will run twice.

niheaven:
I mean, since a PR is required, is it needed for other pushed branches to pass CI test? This test should be run only when a PR is submitted.

rashil2000:
Hmm, yeah, makes sense

rashil2000:
Will it be possible to cache stuff for other workflows too, like pull request/issue handler, hash fix/404 error handler, `/verify` etc.? Currently it installs scoop, sets up 7zip etc. on each run and it takes a lot of time.

niheaven:
Caching scoop is not preferred since we need CI to test modified scoop code. For decompressing tools, current CI won't fetch them until needed, see https://github.com/ScoopInstaller/Scoop/blob/5f407ca43452f6025accdfb58490d6ee100bf31b/test/bin/test.ps1#L30-L33 and https://github.com/ScoopInstaller/Scoop/blob/5f407ca43452f6025accdfb58490d6ee100bf31b/test/bin/test.ps1#L42-L45

rashil2000:
> ... we need CI to test modified scoop code.

Why would this be required for issues/PR handlers?

niheaven:
> Why would this be required for issues/PR handlers?

You mean tests in bucket repo? Yes, scoop code should be cached there, and a proper hash should be found to make sure it is in `HEAD` of 'master' branch.

rashil2000:
> You mean tests in bucket repo? 

Yes, I meant for the buckets itself

niheaven:
I'll take some time to check the probability. It should be cacheable, IMO.

chawyehsu (approved):

PR #4683, opened 2022-01-22
Do not try to set read only property values in the normalize_values function

pstaszko:
#### Description
I added to an if statement to exclude properties that are not settable.

#### Motivation and Context
The normalize_values uses reflection to flatten array values on properties into single strings. This fails if the object has read only properties. Under clean circumstances, I guess this maybe doesn't happen, but in my case I add some properties via extensions and the "$_.Value = whatever" lines throw exceptions that display in the console as errors.

I have been using this changed code for myself and things work without errors now. I also ran the 

#### Checklist:

- [ ] I have updated the tests accordingly. <- I could not figure out how to run the tests without errors (even without my change). I installed pester and ran invoke-pester from the scoop source directory

### DIFF
--- a/lib/json.ps1
+++ b/lib/json.ps1
@@ -179,7 +179,7 @@ function normalize_values([psobject] $json) {
             }
         }
 		
-		# Convert single value array into string
+        # Convert single value array into string
         if (($_.Value -is [Array]) -and $_.IsSettable) {
             # Array contains only 1 element String or Array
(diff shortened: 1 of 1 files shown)


niheaven:
Please start from 'develop' branch and attach some tests, situations, or screenshots.

rashil2000:
Closing due to lack of feedback

PR #4687, opened 2022-01-25
fix(current): Remove 'current' while it's not a junction

niheaven:
#### Description

Junction is copied as native folder when we copy `apps` folder, and this may break scoop's `reset`, `uninstall`, `cleanup` actions. Use `Remove-Item -Recure` could fix this bug.

```pwsh
❯ Copy-Item D:\Scoop\apps\ag D:\ -Recurse
❯ scoop uninstall ag
Uninstalling 'ag' (2.2.5).
Removing shim 'ag.exe'.
Removing shim 'ag.shim'.
Unlinking D:\Scoop\apps\ag\current
'ag' was uninstalled.
❯ Move-Item D:\ag D:\Scoop\apps
❯ scoop prefix ag
D:\Scoop\apps\ag\current
❯ (Get-Item D:\Scoop\apps\ag\current).GetType()

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     False    DirectoryInfo                            System.IO.FileSystemInfo

❯ scoop reset ag
```

Before:

```pwsh
❯ (Get-Item D:\Scoop\apps\ag\current).LinkType -eq 'Junction'
False
```

After:

```pwsh
❯ (Get-Item D:\Scoop\apps\ag\current).LinkType -eq 'Junction'
True
```

#### Checklist:

- [ ] ~I have updated the tests accordingly.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - **autoupdate:** Allow checksum file that contains whitespaces ([#4619](https://github.com/ScoopInstaller/Scoop/issues/4619))
 - **config:** Allow scoop config use Unicode characters ([#4631](https://github.com/ScoopInstaller/Scoop/issues/4631))
+- **current:** Remove 'current' while it's not a junction ([#4687](https://github.com/ScoopInstaller/Scoop/issues/4687))
 - **depends:** Prevent error on no URL ([#4595](https://github.com/ScoopInstaller/Scoop/issues/4595))
 - **depends:** Check if extractor is available ([#4042](https://github.com/ScoopInstaller/Scoop/issues/4042))
 - **decompress:** Fix nested Zstd archive extraction ([#4608](https://github.com/ScoopInstaller/Scoop/issues/4608))
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -911,23 +911,23 @@ function current_dir($versiondir) {
 # Returns the 'current' junction directory if in use, otherwise
 # the version directory.
 function link_current($versiondir) {
-    if(get_config NO_JUNCTIONS) { return $versiondir }
+    if (get_config NO_JUNCTIONS) { return $versiondir }
 
     $currentdir = current_dir $versiondir
 
-    write-host "Linking $(friendly_path $currentdir) => $(friendly_path $versiondir)"
+    Write-Host "Linking $(friendly_path $currentdir) => $(friendly_path $versiondir)"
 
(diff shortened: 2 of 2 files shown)


rashil2000:
> BTW, use `New-Item -ItemType Junction` instead of `mklink /j` to make current junction.

Just to confirm, this works in PS 5.1 and Win 7/8 also, right?

There are 3 more places that use `mklink` (see https://cs.github.com/ScoopInstaller/Scoop?q=mklink) - 2 in lib/install.ps1 and 1 in lib/psmodule.ps1. We can replace those too.

niheaven:
Okay, I'll revert the change here (mklink) and change them all in another PR to make clear change.

rashil2000:
Some more places where we can use `Remove-Item` instead of `rmdir`. Would be cool to refactor them too.


https://github.com/ScoopInstaller/Scoop/blob/2644a5f7b1014a3b5843ecddbf8c739beb2ed802/lib/install.ps1#L1195

https://github.com/ScoopInstaller/Scoop/blob/2644a5f7b1014a3b5843ecddbf8c739beb2ed802/lib/install.ps1#L1198

https://github.com/ScoopInstaller/Scoop/blob/2644a5f7b1014a3b5843ecddbf8c739beb2ed802/lib/psmodules.ps1#L26

https://github.com/ScoopInstaller/Scoop/blob/2644a5f7b1014a3b5843ecddbf8c739beb2ed802/lib/psmodules.ps1#L43

niheaven:
See https://github.com/ScoopInstaller/Scoop/pull/3502/files#diff-cfd4446b6fd477007943865889b4cdbf84e1c089d960b82cd3b57c6bf019fce3 for my previous rework of removing directory. I'm separating #3502 into single purpose PRs, and this should be next job soon.

rashil2000:
Hmm. I just wanted to get rid of COMSPEC entirely, because I faced a ton of issues `The system cannot find the path specified.` today after migrating Scoop to a different drive. This was addressed earlier in https://github.com/ScoopInstaller/Scoop/pull/3993 but only in lib/git.ps1 file.

I'm under the impression that anything COMSPEC does can be handled by PS 5.1 or later.

niheaven:
Yes, of course, everything `COMSPEC` (a.k.a. `cmd.exe`) does could be handled by PowerShell Cmdlets.

I'll do a PR and it should be tested to see if there are some disadvantages.

rashil2000 (approved):

PR #4712, opened 2022-02-03
Prompt user to proceed uninstaller script if process is still running

### LINKED ISSUE #4710: [Bug] Unable to run uninstaller script if any app is running
## Bug Report

I'm trying to create a manifest for https://github.com/sandboxie-plus/Sandboxie. This app installs two services when runs.

In order to perform scoop update or uninstall, I put something like `Stop-Process sandboxie` in the scoop uninstaller script. However, the script doesn't even run because of the error message:
```
$ scoop uninstall sandboxie
Uninstalling 'sandboxie' (1.0.9).
ERROR Application is still running. Close all instances and try again.
```
I think this is originally due to https://github.com/ScoopInstaller/Scoop/pull/3608. I understand the scoop wants user to terminate the process explicitly to avoid data loss, but in this case simply terminating the main process is not enough, because the services also needs stopping (fortunately Sandboxie provides an utility to easily do that).

#### Current Behavior

Scoop refuses to uninstall if any process from the app directory is running, regardless what happens in uninstaller script.

#### Expected Behavior

The manifest should has some power to control how to perform uninstallation when the processes are running.

See above.

#### Possible Solution

Ideally the manifest author can change the existing behavior by specifying some option. With that, when scoop detects process running, instead of immediately error out, it prompts user "Do you want to proceed uninstallation of <app_name>? All unsaved data will be lost [y/N]" When user confirms, it performs the uninstaller script as usual. On the other hand, if scoop does not detects process running, nothing will change from existing logic.

By implementing this, manifest authors can streamlining the update/uninstall procedure by properly implementing uninstaller script. Most of the cases a simple `Stop-Process app_name` is needed.

Basically, instead of the current
```
if (Get-Process | Where-Object { $_.Path -like "$processdir\*" }) {
    error "Application is still running. Close all instances and try again."
    continue
}
```
do something like this
```
if ((Get-Process | Where-Object { $_.Path -like "$processdir\*" }) -and
    (Read-Host "Do you want to proceed uninstallation of '$app'? All unsaved data will be lost [y/N]") -ne 'y') {
    error "Application is still running. Close all instances and try again."
    continue
}
```

### System details

**Windows version:** Windows 11

**OS architecture:** 64bit

**PowerShell version:** 7.2.1

#### Scoop Configuration

```json
{
  "SCOOP_REPO": "https://github.com/ScoopInstaller/Scoop",
  "SCOOP_BRANCH": "master",
  "shim": "kiennq",
  "aria2-warning-enabled": false
}
```

CrendKing:
#### Description

When any process of an application is still running, currently during uninstallation (`uninstall` or `update`), it unconditionally prints error message "Application is still running. Close all instances and try again" and exits. While this approach minimizes risk of data loss, manifest author has no control over the procedure.

Some apps consist of constantly running services and drivers, which has no GUI. In such case, user may have no idea how to "close all instances", and get stuck. Manifest author might be able to `Stop-Service` or call certain external command to stop those processes, but the current logic prevents this.

One simple way to fix this is adding a prompt during uninstallation, as this PR demonstrates. If Scoop detects process running, it calls `Read-Host` to ask user permission to proceed the "uninstaller" script, assuming which will take care of the process termination. In case user answers "N", there will be no change.

Currently there is no flag to guard this change, so all users will be immediately affected by this change. I have no previous experience on how Scoop usually handle these kind of changes. We could create a new property in manifest structure to activate this.

Furthermore, if this is approved, we could even always attempt to call `Stop-Process -Name $bin` after prompt, where "$bin" is the string provided in the manifest's "bin" property, if exist. This way we could streamline the app update even more.

#### Motivation and Context

Closes #4710

1. Install an app.
2. Launch the app without terminating it.
3. `scoop uninstall <app_name>`.
4. Observe a prompt shows during uninstallation.
4.1. Press any key other than "y" results in the "Application is still running. Close all instances and try again" message.
4.2. Press "y", the uninstallation continues.
5. If <app> manifest has uninstaller script that `Stop-Process -Name <app_name>`, app is closed and unininstalled successfully.
5.1. If not, uninstallation fails with error "Couldn't remove <app_dir>; it may be in use."

#### Checklist:

### DIFF
--- a/libexec/scoop-reset.ps1
+++ b/libexec/scoop-reset.ps1
@@ -69,7 +69,8 @@ $apps | ForEach-Object {
 
     #region Workaround for #2952
     $processdir = $dir | Select-Object -ExpandProperty Path
-    if (Get-Process | Where-Object { $_.Path -like "$processdir\*" }) {
+    if ((Get-Process | Where-Object { $_.Path -like "$processdir\*" }) -and
+        (Read-Host "Application '$app' is still running. Do you want to proceed uninstallation? All unsaved data will be lost [y/N]") -ne 'y') {
         error "Application is still running. Close all instances and try again."
         continue
     }
--- a/libexec/scoop-uninstall.ps1
+++ b/libexec/scoop-uninstall.ps1
@@ -58,7 +58,8 @@ if (!$apps) { exit 0 }
 
     #region Workaround for #2952
     $processdir = appdir $app $global | Resolve-Path | Select-Object -ExpandProperty Path
-    if (Get-Process | Where-Object { $_.Path -like "$processdir\*" }) {
+    if ((Get-Process | Where-Object { $_.Path -like "$processdir\*" }) -and
+        (Read-Host "Application '$app' is still running. Do you want to proceed uninstallation? All unsaved data will be lost [y/N]") -ne 'y') {
         error "Application is still running. Close all instances and try again."
         continue
     }
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -238,7 +238,8 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
 
     #region Workaround for #2952
     $processdir = appdir $app $global | Resolve-Path | Select-Object -ExpandProperty Path
-    if (Get-Process | Where-Object { $_.Path -like "$processdir\*" }) {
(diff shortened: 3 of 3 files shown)


CrendKing:
Sorry about that. New PR: https://github.com/ScoopInstaller/Scoop/pull/4713

Issue #4310, opened 2021-04-01
`scoop reset` app running check


FriendlyNeighborhoodShane:
`scoop reset` should have the check for whether an app is running (like in the `uninstall` and `update` routines) as well before doing a reset.

Everything stopped working briefly when I naively ran `scoop reset` on `git` while using the bundled `bash` in a terminal. The only solution was to reboot, completely uninstall `git` through a different terminal, and install again.

https://github.com/lukesampson/scoop/blob/3d67b7d37c4a2714e8942724d4677dcbd3678f01/libexec/scoop-uninstall.ps1#L59-L65

https://github.com/lukesampson/scoop/blob/3d67b7d37c4a2714e8942724d4677dcbd3678f01/libexec/scoop-update.ps1#L237-L243

FriendlyNeighborhoodShane:
Opened PR #4359 with the fix.

PR #4713, opened 2022-02-03
feat(config): Allow Scoop to ignore running processes during reset/uninstall/update

### LINKED ISSUE #4710: [Bug] Unable to run uninstaller script if any app is running

CrendKing:
#### Description

When any process of an application is still running, currently during uninstallation (`uninstall` or `update`), it unconditionally prints error message "Application is still running. Close all instances and try again" and exits. While this approach minimizes risk of data loss, manifest author has no control over the procedure.

Some apps consist of constantly running services and drivers, which has no GUI. In such case, user may have no idea how to "close all instances", and get stuck. Manifest author might be able to `Stop-Service` or call certain external command to stop those processes, but the current logic prevents this.

One simple way to fix this is adding a prompt during uninstallation, as this PR demonstrates. If Scoop detects process running, it calls `Read-Host` to ask user permission to proceed the "uninstaller" script, assuming which will take care of the process termination. In case user answers "N", there will be no change.

Currently there is no flag to guard this change, so all users will be immediately affected by this change. I have no previous experience on how Scoop usually handle these kind of changes. We could create a new property in manifest structure to activate this.

Furthermore, if this is approved, we could even always attempt to call `Stop-Process -Name $bin` after prompt, where "$bin" is the string provided in the manifest's "bin" property, if exist. This way we could streamline the app update even more.

#### Motivation and Context

Closes #4710

1. Install an app.
2. Launch the app without terminating it.
3. `scoop uninstall <app_name>`.
4. Observe a prompt shows during uninstallation.
4.1. Press any key other than "y" results in the "Application is still running. Close all instances and try again" message.
4.2. Press "y", the uninstallation continues.
5. If <app> manifest has uninstaller script that `Stop-Process -Name <app_name>`, app is closed and unininstalled successfully.
5.1. If not, uninstallation fails with error "Couldn't remove <app_dir>; it may be in use."

#### Checklist:


libexec/scoop-reset.ps1
```diff
@@ -69,7 +69,8 @@ $apps | ForEach-Object {
 
     #region Workaround for #2952
     $processdir = $dir | Select-Object -ExpandProperty Path
-    if (Get-Process | Where-Object { $_.Path -like "$processdir\*" }) {
+    if ((Get-Process | Where-Object { $_.Path -like "$processdir\*" }) -and
+        (Read-Host "Application '$app' is still running. Do you want to proceed uninstallation? All unsaved data will be lost [y/N]") -ne 'y') {
```
tech189, line 73:
```suggestion
        (Read-Host "Application '$app' is still running. Do you want to proceed with the reset? All unsaved data will be lost [y/N]") -ne 'y') {
```
Not sure if this prompt is necessary, will any data actually be lost - doesn't reset just recreate shims and shortcuts?

libexec/scoop-uninstall.ps1
```diff
@@ -60,7 +60,8 @@ if (!$apps) { exit 0 }
 
         #region Workaround for #2952
         $processdir = $appDir | Resolve-Path | Select-Object -ExpandProperty Path
-        if (Get-Process | Where-Object { $_.Path -like "$processdir\*" }) {
+        if ((Get-Process | Where-Object { $_.Path -like "$processdir\*" }) -and
+            (Read-Host "Application '$app' is still running. Do you want to proceed uninstallation? All unsaved data will be lost [y/N]") -ne 'y') {
```
tech189, line 64:
```suggestion
            (Read-Host "Application '$app' is still running. Do you want to proceed with uninstallation? All unsaved data will be lost [y/N]") -ne 'y') {
```

libexec/scoop-update.ps1
```diff
@@ -232,7 +232,8 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
 
     #region Workaround for #2952
     $processdir = appdir $app $global | Resolve-Path | Select-Object -ExpandProperty Path
-    if (Get-Process | Where-Object { $_.Path -like "$processdir\*" }) {
+    if ((Get-Process | Where-Object { $_.Path -like "$processdir\*" }) -and
+        (Read-Host "Application '$app' is still running. Do you want to proceed uninstallation? All unsaved data will be lost [y/N]") -ne 'y') {
```
tech189, line 236:
```suggestion
        (Read-Host "Application '$app' is still running. Do you want to proceed with the update? All unsaved data will be lost [y/N]") -ne 'y') {
```

tech189 (commented):
Just some suggestions for improving the wording, I haven't actually tested the functionality - I'll leave that to the maintainers.

libexec/scoop-reset.ps1
CrendKing: The "data loss" I meant is in-memory progress, like if you are editing a file without saving to disk. Nothing already on disk is not lost.

What's your suggestion?

tech189: I don't think scoop stops the process when resetting. Have a look at lines 78-90 of `scoop-reset.ps1`, it's just recreating shims, start menu shortcuts, re-adding environment variables and any directories to the PATH variable, plus recreating the junctions for the `current` folder and the persisted folders. You can try for yourself, I just reset audacity while it was running and it wasn't stopped. It still remembers my preferences so persisted data was not lost either.

So if that's true my suggestion is remove this prompt for reset but keep it for update/uninstallation. (see my other suggestions below for better wording in update and uninstallation)

CrendKing: Oh, you are talking specifically for the "reset" case. Sorry I misunderstood.

So the reason why you don't see the "stop process" message and the new prompt, I think, is because the current `scoop-reset.ps1` has a problem. Instead of getting the `$processdir = appdir $app $global | Resolve-Path | Select-Object -ExpandProperty Path` like `scoop-uninstall.ps1`, it's doing `$processdir = $dir | Select-Object -ExpandProperty Path`. This means the variable points to the versioned directory instead of the `current` junction. Since when you launch an app the current directory is most likely the `current` junction, that process running check in `scoop-reset.ps1` never worked. I just noticed this bug.

Once you change line 71 to the same as `scoop-uninstall.ps1`, you will see the prompt. Can you confirm?

tech189: I had a look at #4310, yes it looks like this prompt is necessary. 

>Once you change line 71 to the same as scoop-uninstall.ps1, you will see the prompt. Can you confirm?

Yes I see the prompt.

In that case, my only suggestion for this file is change the wording from 'proceed uninstallation?' to 'proceed with the reset?'

niheaven:
I don't like this idea, a conclusion notes after all uninstallation to remind user that some apps are still running and not uninstalled is more better, since this one will interrupt the unattended uninstalling process.

Or add a new config option to let scoop force exit all running processes and uninstall/update the apps.

CrendKing:
I agree that this change is backward incompatible and breaks unattended process. As I mentioned in the opening, if the community would come up with a "force exit running process" option, I'm OK with that too. Maybe provide that as a `scoop config` option (i.e. when true, always force exit instead of always error and fail uninstall/reset).

@rashil2000, what do you think?

libexec/scoop-reset.ps1
CrendKing: Yes. I'll change the wording if this approach is approved.

rashil2000:
Hmm, the config option is a much better UX. It does not break any previous user workflows, and offers customizability without needing to ask the user for a prompt each time. 

For implementation of the config option, you can refer to this PR - #4155 

You'll also need to update the libexec/scoop-config.ps1 file, for documentation.

CrendKing:
OK. Just to recap the new requirements:

1. Add a new boolean config to the `scoop config` command, which enables the "force kill processes" function. This defaults to false for backward compatibility.
2. For uninstall, update and reset, if Scoop detects any process running from the app directory, AND the config above is enabled by user, instead of print error message and exit, forcefully kill all running process by calling `Stop-Process`. Print relevant message for this action.
3. Afterwards, proceed the uninstall/update/reset procedure as usual.

Is this good?

CrendKing:
Updated the PR. Moved the handling logic to core.ps1 for reuse between all the 3 use cases. If this function is better placed in a different file, please advise.

rashil2000:
Do you think Scoop should stop the process if enabled? It will most likely fail (and print a wall of errors) - unless they are stopped in the proper order and with the proper flags (like `-Force` if necessary). I think we should just let the user stop the processes themselves. 

The config option should be boolean - false (by default) to retain current behaviour, and true to just print the warning and proceed with uninstaller script.

niheaven:
For the code, put the function under `install.ps1` IMO.

CrendKing:
> It will most likely fail (and print a wall of errors) - unless they are stopped in the proper order and with the proper flags (like -Force if necessary)

According to the [PowerShell doc](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/stop-process?view=powershell-7.2) without `-Force` the cmdlet doesn't stop process spawned from different user account. So I suppose since the function could be thrown with a global app, which could be started by other user, we should use `-Force` here. Gotcha.

I tried several apps locally, and I don't get any problem. If you have see errors, could you provide the repro steps?

> true to just print the warning and proceed with uninstaller script.

To clarify, I think we are discussing about two use cases:

1. Normal everyday app is running when doing update/uninstall, where a generic `Stop-Process` is sufficient to handle.
2. Special app (such as Everything driver/service from https://github.com/ScoopInstaller/Scoop/issues/4710), which no GUI app is running. Stopping those processes require special tools/steps.

For 2, I completely agree we should leave to the uninstaller script.
For 1, if we just display a warning, since most manifest today does not have `Stop-Process`, it's not gonna change anything. Besides, what other special ways can the manifest authors do better than a generic `Stop-Process`?

> For the code, put the function under install.ps1 IMO.

Gotcha.

rashil2000:
> If you have see errors with this impl, could you provide the repro steps?

I've encountered problems outside of Scoop, where a generic Stop-Process isn't enough.

> For 1, if we just display a warning, since most manifest today does not have Stop-Process, it's not gonna change anything.

That's preferable, to be honest. Scoop is a pretty non-invasive manager, and doing something outside of its scope will deviate from that. As I said, if any apps need special-casing, manifest authors will include a Stop-Process in their uninstaller scripts, and I think that's sufficient.

CrendKing:
Understood. I'll remove the `Stop-Process` and update.

CrendKing:
Manual test steps I did:

1. Install an app A, which has GUI. Config JSON is empty.
2. Launch A. Leave it running. Do `scoop uninstall A`.
3. Observe error message `ERROR Application "A" is still running. Close all instances and try again.` and Scoop terminates there.
4. Add config `"ignoreRunning": false,` and repeat steps 2-3. Observe exactly the same result.
5. Change `ignoreRunning` to `true`. Observe warn message `WARN  Application "A" is still running. Scoop is configured to ignore this condition.` instead and other messages such as "Removing shim". Scoop stops with error `ERROR Couldn't remove 'D:\Scoop\apps\A\0.0.1'; it may be in use.` This is because A manifest does not handle running process in uninstaller.
6. Reinstall A. Add the following script to A\0.0.1\manifest.json:
```
    "uninstaller": {
        "script": "Stop-Process -Name A"
    },
```
7. Do uninstallation. Observe the warning message, then `Running uninstaller script...` , A's process gets killed, and eventually `'A' was uninstalled.`

=====

Is there any test file need update?

rashil2000:
Looks great. I'll do a proper review later today.

Hmm, I don't think there are tests for this. @niheaven can you confirm?

rashil2000:
Vague idea: instead of a config option, how about we detect the presence of `uninstaller` field in the manifest, to automate this process? If `uninstaller` exists, Scoop will warn the user and proceed. If it doesn't, it pops up an error and stops.

niheaven:
Nope, this process related to specific app installation/update cannot be tested. Just run `.\bin\test.ps1` to see if all passes.

BTW, you should check config before running the new function to avoid `Get-Process` for every uninstallation/updating.

niheaven:
Aha sorry, I misunderstood the logic of the function 🤣 Please ignore it.

rashil2000 (changes requested):
We should call it `ignore_running_processes` to be more clear.

rashil2000 (dismissed):

rashil2000:
Please add an entry in the changelog

niheaven:
Call the function `ignore_running_process` and flip the output (`$true` for ignore/`$false` for break) to make this more readable.

CrendKing:
Oh, you mean I should make the function "ignore_running_process", not "handle_running_processes"? What about the config? Any suggestion?

Sorry for misunderstanding.

rashil2000:
This should be `feat` instead of `fix`

niheaven:
It should be `stop_running_process` 🤣 

> Or you can change the PR yourself, feel free to just modify it.

Okay, I'll change it a little, feel free to reverse my commit if it got something wrong

niheaven:
image

niheaven (approved):

niheaven:
@CrendKing If this is fine, I'll merge it.

CrendKing:
Of course. Go ahead.

rashil2000:
Thanks for the contribution @CrendKing!

CrendKing:
@niheaven I think https://github.com/ScoopInstaller/Scoop/pull/4713/commits/e55639d2d7913ef5e5d8082ef976825e08aea094 introduced a bug. Previously either the error or warn message would print only if there is some running process. After that commit, the warn message would print as long as that config is set.

Should I make a PR to correct it or you can do it?

niheaven:
It's been fixed in #4731, you could try it 🤣

CrendKing:
That's exactly what I'm looking for. Sorry didn't notice that commit. Thank you!

Issue #339, opened 2015-04-17
All cmd shims created for Invoke-Build have a mistake


nightroman:
The problem may be common for all shims created for PowerShell scripts. But I
have noticed it for Invoke-Build, so I explain it just for Invoke-Build.

Let's take a look at *invoke-build.cmd*

    @powershell -noprofile -ex unrestricted "& '...\scoop\apps\invoke-build\2.9.14\Invoke-Build.ps1' %*;exit $lastexitcode"

For Invoke-Build the correct shim would be

    @powershell -noprofile -ex unrestricted "& '...\scoop\apps\invoke-build\2.9.14\Invoke-Build.ps1' %*"

Because when IB fails it throws a terminating error and PowerShell in this case
exists with code 1. If IB works then PowerShell exits with code 0 itself. This
is a normal scenario.

But the extra part `exit $lastexitcode` is a mistake and it may break things.

Same as above, if IB fails it throws a terminating error. But this means that
the next statement `exit $lastexitcode` is not called at all. So it is not
needed.

The case when IB succeeds is more interesting. The automatic variable
`$lastexitcode` is changed when a native application is called.

If no native command is called this variable is not even defined. In some cases
this will be treated as null and converted to 0. In other cases it may cause a
failure. Try to run this script in a fresh PowerShell session:

    Set-StrictMode -Version 2
    exit $LASTEXITCODE

It fails: *The variable '$LASTEXITCODE' cannot be retrieved because it has not been set.*
So that if a build script actually sets `Set-StrictMode -Version 2` and calls
no native commands then it always fails due to the shim `exit $LASTEXITCODE`.

Then, if a native command was called then its exit code tells nothing about IB
success or failure. Suppose a build script invoked by IB is

    task copy {
        exec {robocopy source destination /mir} (0..3)
    }

`robocopy` codes 0..3 are for success. Note that the function `exec` is told to
take this into account and to not fail on 0..3 codes. But PowerShell still sets
`$lastexitcode` to one of this values when `robocopy` is called. And then the
shim code uses it for exit. So in spite of the fact that IB succeeds the shim
exists with non zero code. This is the mistake.

As the author of Invoke-Build, I am glad and proud it is noticed by *scoop*.
Of course I would like Invoke-Build to be installed properly.

nightroman:
[...]

PR #4718, opened 2022-02-07
refactor(scoop-list): Allow list manipulation

### LINKED ISSUE #3162: [Proposal] Homebrew style scoop-list output option
capture

What is your feeling about this?

### LINKED ISSUE #4138: How to write search query in list/search? Is there any docs?
I want to find all apps in a bucket installed by me. What is the search query? I try to find docs telling me How to write search query but I failed.

### LINKED ISSUE #4179: Held apps color is not visible in Scoop List
There was an asterisk `*` indicating apps on hold and now it is gone.

image

rashil2000:
#### Description
Return PSObject instead of printing to terminal.

#### Motivation and Context

Allows manipulation, sorting etc. of the list with the help of PowerShell's pipelines.

Closes #3162
Closes #4138
Closes #4179

##### Before 

image

image

##### After

image

image

##### More features (not possible before)

Homebrew-style output:

image

Filter apps from a bucket:

image

#### Checklist:

### DIFF
--- a/libexec/scoop-list.ps1
+++ b/libexec/scoop-list.ps1
@@ -3,10 +3,10 @@
 # Help: Lists all installed apps, or the apps matching the supplied query.
 param($query)
 
-. "$psscriptroot\..\lib\core.ps1"
-. "$psscriptroot\..\lib\versions.ps1"
-. "$psscriptroot\..\lib\manifest.ps1"
-. "$psscriptroot\..\lib\buckets.ps1"
+. "$PSScriptRoot\..\lib\core.ps1"
+. "$PSScriptRoot\..\lib\versions.ps1"
+. "$PSScriptRoot\..\lib\manifest.ps1"
+. "$PSScriptRoot\..\lib\buckets.ps1"
 
 reset_aliases
 $def_arch = default_architecture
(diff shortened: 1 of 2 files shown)


rashil2000:
In the next PR I will add lastwritetime field so that we can sort the list by last updated/installed time

niheaven:
Source or Bucket? I think `Bucket` is better for Scoop.

rashil2000:
`Source` because it can be a local file path or a remote URL also, so `Bucket` won't fit there

rashil2000:
I just realized that this PR (and the previous one for buckets by @HUMORCE) does not work in Cmd and Bash. No output is shown, unless we remove the `exit $lastexitcode` part from these lines:

https://github.com/ScoopInstaller/Scoop/blob/7967905980702688f876225035146a1bb1e2bfb8/lib/core.ps1#L645

https://github.com/ScoopInstaller/Scoop/blob/7967905980702688f876225035146a1bb1e2bfb8/lib/core.ps1#L656

@niheaven any idea why these are essential, and whether we can remove them?

niheaven:
This line has been added here for a long time and I must check original codes.

niheaven:
Found an old explanation of this `exit $LastExitCode` (https://github.com/ScoopInstaller/Scoop/issues/339#issuecomment-94091121), Maybe it's better to find out why there is no output...

libexec/scoop-list.ps1
```diff
@@ -15,37 +15,44 @@ $local = installed_apps $false | ForEach-Object { @{ name = $_ } }
...
-    exit 0
-} else {
-    write-host "There aren't any apps installed."
-    exit 1
+    $item.Info = $info -join ', '
+
+    $list += $item
 }
+
+$list.ForEach({[PSCustomObject]$_})
```
niheaven, line 57:
```suggestion
$list.ForEach({ [PSCustomObject]$_ }) | Out-Host
```

This could solve the problem.

niheaven:
image

Some suggestions: For this generated manifest, tell user this is `Generated` and a manifest location is more better. This needs further parsing for manifest location with `workspace`.

libexec/scoop-list.ps1
rashil2000: Putting `Out-Host` disables the manipulation of objects in the pipeline, so this will just make it same as `Write-Host`

rashil2000:
> Found an old explanation of this `exit $LastExitCode` ([#339 (comment)](https://github.com/ScoopInstaller/Scoop/issues/339#issuecomment-94091121)), Maybe it's better to find out why there is no output...

I kinda agree with nightroman there. I didn't fully understand the explanation luke gave (also there are no examples in his comment to reproduce the problem if we remove it). 

Maybe we just put `exit %errorlevel%` at the end of the .cmd file?

rashil2000:
I just checked. If you run the command `powershell -nop -c "exit 1"` in Cmd, it will show the correct exitcode. The `exit $LastExitCode` in the line is redundant IMO.

libexec/scoop-list.ps1
niheaven: `Write-Host` cannot produce correct output...

You could modify cmd shim as:

```cmd
if %errorlevel% equ 0 (
    pwsh -noprofile -ex unrestricted -command "(& 'D:\Scoop\apps\scoop\current\bin\scoop.ps1'  %args%) | Out-Host;exit $Global:LASTEXITCODE"
) else (
    powershell -noprofile -ex unrestricted -command "(& 'D:\Scoop\apps\scoop\current\bin\scoop.ps1'  %args%) | Out-Host;exit $Global:LASTEXITCODE"
)
```

Pipeline is fine in PS and bash:

image

image

I'm not familiar with CMD, you could test it yourself.

niheaven:
> Instead of using `-Command` parameter, we can use `-File`, which does not suffer from the exitcode limitation described in the Microsoft Docs. Any exit code is returned to the Cmd shell as is.
> 
> ```batchfile
> powershell -noprofile -ex unrestricted -file "D:\Data\Scoop\apps\scoop\current\bin\scoop.ps1"  %args%
> ```
> 
> (I'm not even sure why we use `-Command` in the first place, when `-File` parameter exists exactly for this purpose).

PS5 doesn't document exit code for `-File` clearly, and this code block is written about 8 years ago while PS's version is 2. So I'm not sure if it will output correct exit code in current PowerShell edition (5.1 and 7).

If `-File` works, it is better, IMO.

rashil2000:
I tested in PS 5.1 before.

Just now, I tested in PS 7.1.2.

Both work nicely 👍

niheaven:
Not only scoop itself, but also other apps with `ps1` shim? If so, let's use `-File` and test in `develop` branch.

HUMORCE:
> I just realized that this PR (and the previous one for buckets by @HUMORCE) does not work in Cmd and Bash. No output is shown, unless we remove the exit $lastexitcode part from these lines

switch parameter for return? nothing else comes to mind at this time.

rashil2000:
> > I just realized that this PR (and the previous one for buckets by @HUMORCE) does not work in Cmd and Bash. No output is shown, unless we remove the exit $lastexitcode part from these lines
> 
> switch parameter for return? nothing else comes to mind at this time.

It will work once #4721 is merged

HUMORCE:
> It will work once https://github.com/ScoopInstaller/Scoop/pull/4721 is merged

i seem to have missed some comments in this thread on gh mobile.

libexec/scoop-list.ps1
niheaven: No, fish shell and starship ^_^

niheaven:
> image
> 
> Some suggestions: For this generated manifest, tell user this is `Generated` and a manifest location is more better. This needs further parsing for manifest location with `workspace`.

What about this? Or maybe another PR?

rashil2000:
I have not added any new features in this PR. Scoop list shows the same information as it did before.

Each new feature should be a new PR IMO

niheaven (approved):

libexec/scoop-list.ps1
rashil2000: Oh, Fish shell is quite nice (I saw syntax highlighting, and you said bash, so I got confused 😅)

niheaven: Hmm, strictly, `sh`? Haha

Now `Zsh` is the default shell of Ubuntu, but it needs too many custom configurations, so I prefer Fish shell with a more simpler `oh-my-fish` (`oh-my-zsh` is too complex)

rashil2000: > Zsh is the default shell of Ubuntu

Oh, I didn't know that! When did they change it?

I don't use WSL much (I use Zsh, Bash, Fish all on MSYS2 and stay on Windows :)).

niheaven: Forgotten... When I install Ubuntu using WSL2 in Windows 11 last time (mid-2021), it prompts that zsh is the default shell now...

Thanks to Windows Terminal and starship, I often forget what I'm using since basic commands (`cd`, `ls`, `rm`, etc.) and the prompt have similar interface.

PR #4723, opened 2022-02-08
feat(scoop-list): Show last-updated time

### LINKED ISSUE #3777: [feature] `scoop list` sort by created time, update time
Sometimes, I forget which apps have been installed recently, and using the `scoop list` command is just alphabetical, which is not convenient. Add a sort parameter would be great.

rashil2000:
#### Description

Include an "Updated" field in the list, so users can sort it by most-recently updated apps.

**Note:** By default, if an object has more than 4 properties, PowerShell shows it as a 'List' instead of 'Table'. So a custom format file is needed to retain the Table here.

#### Motivation and Context

Closes #3777

Relates to #4718 

image

image

#### Checklist:


rashil2000:
@niheaven how to detect if a manifest is user-generated? If it's a one-line change I can include it in this PR (and update the title).

niheaven:
Likes in my screenshot, the manifest is installed from a folder called `workspace` which contains all user generated manifests.

You could try it by `scoop install ag@2.2.4` and see `scoop list ag`. This should work, but I'm not near a Windows machine to test it.

This type of installation is always hold, since the generated manifest will never be updated and `scoop update ag` doesn't work here 'cos its source never change.

rashil2000:
I added the `Auto-generated` part, but I think the Source should still point to the bucket from where the manifest was generated. The workspace path is an internal Scoop detail and shouldn't be exposed to users.

Any way to know from where the custom manifest was generated?

niheaven:
Nope, scoop cannot know the original bucket except for an app search (just like the way scoop find app, by searching all local buckets).

You could mask the manifest by using `Generated` as `Source` instead of add a info tag.

rashil2000:
Done

niheaven (approved):
LGTM

PR #4727, opened 2022-02-09
feat(shim): Allow switching shim's target

niheaven:
**A WONDERFUL FEATURE OF SCOOP** 😆 🎉 

#### Description

Now Scoop could switch apps using `scoop alter <app>` just as `(update-)alternatives` in Linux. Meanwhile, uninstalling some apps which overwrite other's shims will restore previous ones (instead of removing commands).

- Instead of overwriting shims, Scoop renames original ones.
- During the uninstalling process, if there're renamed shims, Scoop will rename them back.
- Add new `scoop alter` command to choose default of command (a.k.a., shim)

#### Motivation and Context

There are many issues requesting this feature, will be added soon.

Tested in PowerShell 5.1/7.2.1, CMD, and WSL/Ubuntu. (Not tested in MSYS2 and CygWin)

Use `7zip`/`7zip-zstd` and `python`/`python3` as examples:

image

image

For command without alternatives:

image

For command without shim:

image

#### Checklist:

### DIFF
--- a/libexec/scoop-alter.ps1
+++ b/libexec/scoop-alter.ps1
@@ -1,0 +1,60 @@
+# Usage: scoop alter <command>
+# Summary: Maintain shims determining default commands (similar to '(update-)alternatives' on Linux)
+# Help: Display and maintain alternatives of a shimmed command installed with Scoop (similar to '(update-)alternatives' on Linux)
+
+param($command)
+. "$psscriptroot\..\lib\core.ps1"
+. "$psscriptroot\..\lib\help.ps1"
+
+reset_aliases
+
+if (!$command) { 'ERROR: <command> missing'; my_usage; exit 1 }
+
+try {
+    $gcm = (Get-Command "$command" -ErrorAction Stop).Path.ToString()
+} catch {
+    abort "'$command' not found" 3
+}
+
+$path = $gcm -replace '\.exe$', '.shim'
+$userShimPath = "$(Resolve-Path $(shimdir $false))"
+$globalShimPath = fullpath (shimdir $true) # don't resolve: may not exist
+
+if ($path -like "$userShimPath*" -or $path -like "$globalShimPath*") {
+    $altShims = Get-Item -Path "$path.*" -Exclude '*.shim', '*.cmd', '*.ps1'
+    if ($null -eq $altShims) {
+        Write-Host "No alternatives of '$command' found."
+        exit 0
+    }
+    $app = get_app_name_from_shim $path
+    $apps = @($app) + ($altShims | ForEach-Object { $_.Extension.Remove(0, 1) } | Select-Object -Unique)
+    [System.Management.Automation.Host.ChoiceDescription[]]$altApps = 1..$apps.Length | ForEach-Object {
+        New-Object System.Management.Automation.Host.ChoiceDescription "&$($_)`b$($apps[$_ - 1])", "Sets '$command' shim from $($apps[$_ - 1])."
+    }
+    $selected = $Host.UI.PromptForChoice("Alternatives of '$command' command", "Please choose one that provides '$command' as default:", $altApps, 0)
+    if ($selected -eq 0) {
+        Write-Host "'$command' is already from '$app', nothing changed."
+        exit 0
+    } else {
+        $newApp = $apps[$selected]
+        Write-Host "Use '$command' from '$newApp' as default..." -NoNewline
+        $pathNoExt = strip_ext $path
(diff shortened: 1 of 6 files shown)


rashil2000:
Wow, very interesting!

Some notes:
- Since we are changing/refactoring function signatures, would it be good to follow PS conventions and call it `Remove-Shim` instead of `rm_shim`?
- In a subsequent PR, I might add a functionality to create user-defined shims, to solve https://github.com/ScoopInstaller/Scoop/issues/3303. Since this is related to shims itself, should we call the file something other than `scoop-alter.ps1`?

rashil2000:
> I'll make another PR that moves all shim-related functions to a separate file and refactors all the functions using PowerShell conventions, it's better IMO, right?

Sure.

> The 'user-defined shims' in #3303 is more likely `links`, it links a virtual app folder to some existing one and make new shims that points to them (likes `scoop reset`), have I misunderstood?

No no, the idea is to allow creation of custom shims. For example, I want to create a shim like this:

```
scoop shim nvimdiff D:\Data\Scoop\apps\neovim\bin\nvim.exe -d
```

The `nvimdiff` shim is not created by the neovim manifest, but we can allow users to create custom shims like this. The format would be:

```
scoop shim <shim_name> <full_path_to_program> <arg1> <arg2> <arg3> ...
```

niheaven:
Okay, got it.

`scoop alter` is more user-friendly IMO, end users needn't know what is 'shim' and they just `alter` the `command`. A specific command `scoop shim` for advanced users could add a subcommand that directly call it. `scoop shim` might have following subcommand:

- `scoop shim <command>` shows where the command comes from
  - likes `scoop which`, could call it directly or modify some output likes '<command> is from <app>'
- `scoop shim alter <command>` calls `scoop alter` to change command's source
- `scoop shim add <shim_name> <full_path_to_program> <arg1> <arg2> <arg3> ...`
  - this is what you want
- `scoop shim rm/del <command>` remove shim
  - maybe give some caution notes?

niheaven:
Any other suggestions? @rashil2000 @issaclin32

rashil2000:
Yes, plan looks awesome!

But why create both `scoop-alter` and `scoop-shim` if `scoop shim alter` exist?

niheaven:
> Yes, plan looks awesome!
> 
> But why create both `scoop-alter` and `scoop-shim` if `scoop shim alter` exist?

For convenience 😆 . For example, in my use cases, I may never use `scoop shim` except for `scoop shim alter` if there is only `scoop shim`, and this is the general usage for most users IMO.

rashil2000:
To be honest, the command `scoop shim alter` is more clear as to what exactly it is doing. `scoop alter` could mean anything (and not just shims). @issaclin32 thoughts?

issaclin32:
I would vote for `scoop shim alter`. It looks more intuitive.

niheaven:
This is not a binary choice; they are mutually inclusive. If only `scoop shim alter` is kept, I'll do some additional coding that adds basic functions of `scoop shim`.

niheaven:
@rashil2000 I've done `scoop shim` command, but I prefer merge this PR and I'll submit a new one that accomplish the feature.

libexec/scoop-alter.ps1
```diff
@@ -0,0 +1,60 @@
...
+                if (Test-Path -Path $newShimPath -PathType Leaf) {
+                    Rename-Item -Path $newShimPath -NewName $shimPath -Force
+                }
+            }
+        }
+        Write-Host 'done.'
+    }
+} else {
+    [Console]::Error.WriteLine('Not a scoop shim.')
+    Write-Output "Command path is: $gcm"
```
rashil2000, line 56:
```suggestion
    Write-Host "Command path is: $gcm"
```

Why use `[Console]::Error.WriteLine` instead of `Write-Error` or `Write-Host`?

rashil2000, line 56:
```suggestion
    [Console]::Error.WriteLine('Not a scoop shim.')
    Write-Host "Command path is: $gcm"
```

Why use `[Console]::Error.WriteLine` instead of `Write-Error` or `Write-Host`?

niheaven: Copied from `scoop-which.ps1`, maybe `Write-Host` is enough.

niheaven: Copied from `scoop-which`, maybe `Write-Host` is enough.

rashil2000 (approved):

PR #4734, opened 2022-02-11
fix(shim): Fix shim when app path has white spaces

niheaven:
#### Description

Enclose app path with double quotes to fix bug while app path has white spaces.

#### Motivation and Context

Closes #4728

#### How Has This Been Tested?

Only occurred in exe shims, hard to test.

#### Checklist:

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).
- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #4736, opened 2022-02-13
feat(scoop-shim): Add `scoop shim` to manipulate shims

### LINKED ISSUE #3303: Adding shims to an installation

niheaven:
#### Description

Add a new `scoop shim` command, with following subcommand:

- `scoop shim add <shim_name> <command_path>`
  - Create shim in shim dir, `command_path` should be full path or shim name or external program
  - If `command_path` has whitespaces or has arguments, use quotes
- `scoop shim remove <shim_names>`
  - Remove shims, custom or predefined, use whitespaces to separate
- `scoop shim list [<shim_name>]`
  - List all shims or specific ones (regex search)
- `scoop shim info <shim_name>`
  - Shows shim information
- `scoop shim alter <shim_name>`
  - Change shim's source

I borrowed some ideas from #4718 (thanks to @rashil2000) to show shims list table view and make some tweaks to `scoop list` command.

#### Motivation and Context

Closes #3303

Relates to #4727

image
image
image

#### Checklist:

- [x] ~I have updated the tests accordingly.~


niheaven:
> For the list command, there are 7 columns. When there are that many columns in an object, the List view is more comprehensible (instead of Table view).

For so many shim items, Table View is more compact. 

> For the info command, we can reuse the ScoopShimType and return the object instead of putting Write-Host.

I can do this, and it is my first thought when writing the function. But I wonder who needs such an object even if we return it? For clear result display, a List View of ScoopShim should be added.

> So can we merge the code for scoop-alter here? (instead of calling the script explicitly)

Sure, will do it.

rashil2000:
> For so many shim items, Table View is more compact.

I wonder how many results it would typically return. Won't it be 2 or 3 at max?

> But I wonder who needs such an object even if we return it?

Right now we don't need, but in future someone might xD. Also, it looks a lot nicer. 

image

> For clear result display, a List View of ScoopShim should be added.

Yes, that's what I had in mind.

niheaven:
> I wonder how many results it would typically return. Won't it be 2 or 3 at max?

If you just want a full 'list' (a.k.a. `scoop shim list`), maybe hundreds? Of cause, if you filter it by app names, 2 or 3.

rashil2000:
Hmm, understood.

niheaven:
`scoop shim list` and `scoop shim info`:

image

`scoop shim alter` and `scoop shim rm` (alias to `scoop shim remove`):

image

Since `scoop shim list` may produce non array output, I cannot distinct the output of `scoop shim list` and `scoop shim info`, so using different type in `ScoopTypes.Format.ps1xml` to create different views.

niheaven:
After this is merged, Scoop Core could be bumped to version 0.9.0.0. This is a wonderful version IMO, isn't it?

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,211 @@
...
+    $info = @{ PSTypeName = 'ScoopShim' }
+    $info.Name = strip_ext (fname $ShimPath)
+    $info.Path = $ShimPath -replace 'shim$', 'exe'
+    $info.Source = get_app_name_from_shim $ShimPath
+    $info.Type = if ($ShimPath.EndsWith('.shim')) { 'Application' } elseif ($ShimPath.EndsWith('.cmd')) { 'Script' } else { 'Unknown' }
+    $altShims = Get-Item -Path "$ShimPath.*" -Exclude '*.shim', '*.cmd', '*.ps1'
+    if ($altShims) {
+        $info.Alternatives = @($info.Source) + ($altShims | ForEach-Object { $_.Extension.Remove(0, 1) } | Select-Object -Unique)
+    }
+    $info.IsGlobal = if ($ShimPath.StartsWith("$globalShimDir")) { $true } else { $false }
```
rashil2000, line 75:
```suggestion
    $info.IsGlobal = $ShimPath.StartsWith("$globalShimDir")
```

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,211 @@
...
+        } else {
+            $commandPath = Get-ShimTarget (Get-ShimPath $addArgs[0] $global)
+        }
+        if ($commandPath -and (Test-Path $commandPath)) {
+            Write-Host "Adding $globalTag shim " -NoNewline
+            Write-Host $ShimName -ForegroundColor Cyan -NoNewline
+            Write-Host ' ... ' -NoNewline
+            shim $commandPath $global $ShimName $(if ($addArgs.Length -gt 1) { $addArgs[1..($addArgs.Length - 1)] })
+            Write-Host 'Done'
+        } else {
```
rashil2000, line 114: Suggestion: there can be a third case here - in which the application exists in PATH - which we can accommodate. For example, the following fails:

image

But the app does exist in PATH, see:

image

We can get the logic for target from here: https://github.com/ScoopInstaller/Scoop/blob/21c5c6e775f51ea84eafd436b55078d43490a7e6/libexec/scoop-which.ps1#L38-L39

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,211 @@
...
+            abort "ERROR: '$($addArgs[0])' does not exist" 3
+        }
+    }
+    { $_ -in @('remove', 'rm') } {
```
rashil2000, line 118:
```suggestion
    { $_ -in 'remove', 'rm' } {
```

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,211 @@
...
+        if ($addArgs[0] -match '[\\/]') {
+            $commandPath = $addArgs[0]
+        } else {
+            $commandPath = Get-ShimTarget (Get-ShimPath $addArgs[0] $global)
+        }
+        if ($commandPath -and (Test-Path $commandPath)) {
+            Write-Host "Adding $globalTag shim " -NoNewline
+            Write-Host $ShimName -ForegroundColor Cyan -NoNewline
+            Write-Host ' ... ' -NoNewline
+            shim $commandPath $global $ShimName $(if ($addArgs.Length -gt 1) { $addArgs[1..($addArgs.Length - 1)] })
```
rashil2000, line 112: It will be a very common use case where users can specify shim args. But there is an error. See this example:

image

Here the scoop-shim.ps1 script is swallowing up the args needed to pass into the `add` subcommand.

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,211 @@
...
+# To alternate a shim's target source, use the 'alter' subcommand:
+#
+#     scoop shim alter <shim_name>
+#
+# Options:
+#   -g, --global       Add/Remove/Info/Alter global shim(s)
+
+param($SubCommand, $ShimName)
+
+. "$PSScriptRoot\..\lib\core.ps1"
```
rashil2000, line 30:
```suggestion
```

`core` is not required because it is imported already in https://github.com/ScoopInstaller/Scoop/blob/21c5c6e775f51ea84eafd436b55078d43490a7e6/bin/scoop.ps1#L6

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,211 @@
...
+    }
+    $info.IsGlobal = if ($ShimPath.StartsWith("$globalShimDir")) { $true } else { $false }
+    $info.IsHidden = if ((Get-Command -Name $info.Name).Path -eq $info.Path) { $false } else { $true }
+    [PSCustomObject]$info
+}
+
+function Get-ShimPath($ShimName, $Global) {
+    '.shim', '.cmd' | ForEach-Object {
+        $shimPath = Join-Path (shimdir $Global) "$ShimName$_"
+        if (Test-Path -Path $shimPath) {
```
rashil2000, line 83:
```suggestion
        if (Test-Path $shimPath) {
```

rashil2000:
I have made a commit for minor styling stuff, please pull.

Doubt: when will a shim be hidden? Do you have an example?

rashil2000:
Another suggestion: it would be useful to show if a shim is custom-made while doing `scoop shim info <shim>`. But I guess this is not possible due to current implementation of shims Scoop core.

niheaven:
> Doubt: when will a shim be hidden? Do you have an example?

For some internal alias or some command from other apps that has high priority

```console
~ took 22m12s
❯ scoop shim list | ? { $_.IsHidden }

Name     Source           Alternatives IsGlobal IsHidden
----     ------           ------------ -------- --------
ar       busybox                       False    True
bash     busybox                       False    True
cat      coreutils-uutils              False    True
clear    busybox                       False    True
cp       coreutils-uutils              False    True
diff     diffutils                     False    True
dos2unix busybox                       False    True
dotnet   dotnet-sdk                    False    True
echo     coreutils-uutils              False    True
env      coreutils-uutils              False    True
expand   coreutils-uutils              False    True
find     busybox                       False    True
getopt   busybox                       False    True
hostname coreutils-uutils              False    True
iconv    busybox                       False    True
```

libexec/scoop-shim.ps1
niheaven: Hmm, you should put arguments insides <command_path>, I'll change the code

rashil2000: Could you give an example (specifically when the app binary path has spaces in it)?

rasa:
> For some internal alias or some command from other apps that has high priority

So a hidden shim is one app's shim, that's been overwritten by another app's shim?

niheaven:
> So a hidden shim is one app's shim, that's been overwritten by another app's shim?

A hidden shim is an inaccessible one, for example I've `curl` installed from Scoop, but since Windows put a `curl.exe` under `System32`, I couldn't use scooped `curl` unless by its full path.

```console
~
❯ scoop which curl
C:\WINDOWS\system32\curl.exe

~
❯ Get-Command curl -All

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     curl.exe                                           7.79.1.0   C:\WINDOWS\system32\curl.exe
Application     curl.exe                                           0.0.0.0    D:\Scoop\shims\curl.exe

~
❯ scoop shim info curl

Name     : curl
Path     : D:\Scoop\shims\curl.exe
Source   : curl
Type     : Application
IsGlobal : False
IsHidden : True
```

niheaven:
```console
~
❯ scoop shim add my-shim 'curl --version'
Adding local shim my-shim...

~
❯ my-shim
curl 7.81.0 (x86_64-pc-win32) libcurl/7.81.0 OpenSSL/3.0.1 (Schannel) zlib/1.2.11 brotli/1.0.9 libidn2/2.3.2 libssh2/1.10.0 nghttp2/1.46.0 libgsasl/1.10.0
Release-Date: 2022-01-05
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli gsasl HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL SSPI TLS-SRP UnixSockets

~
❯ scoop shim info my-shim

Name     : my-shim
Path     : D:\Scoop\shims\my-shim.exe
Source   : curl
Type     : Application
IsGlobal : False
IsHidden : False

~
❯ Get-Command curl

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     curl.exe                                           7.79.1.0   C:\WINDOWS\system32\curl.exe

~
❯ scoop shim add my-shim '"C:\Program Files\WindowsApps\Microsoft.PowerShell_7.2.1.0_x64__8wekyb3d8bbwe\pwsh.exe" -v'
Adding local shim my-shim...
WARN  Overwriting shim ('my-shim.exe' -> 'pwsh.exe') installed from curl

~
❯ my-shim
PowerShell 7.2.1

~
❯ scoop shim add my-shim '7z l'
Adding local shim my-shim...
WARN  Overwriting shim ('my-shim.exe' -> '7z.exe') installed from External

~
❯ my-shim D:\Scoop\cache\ag#2.2.5#https_github.com_JFLarvoire_the_silver_searcher_releases_download_2.2.4-Windows_ag.zip

7-Zip 21.07 (x64) : Copyright (c) 1999-2021 Igor Pavlov : 2021-12-26

Scanning the drive for archives:
1 file, 479251 bytes (469 KiB)

Listing archive: D:\Scoop\cache\ag#2.2.5#https_github.com_JFLarvoire_the_silver_searcher_releases_download_2.2.4-Windows_ag.zip

--
Path = D:\Scoop\cache\ag#2.2.5#https_github.com_JFLarvoire_the_silver_searcher_releases_download_2.2.4-Windows_ag.zip
Type = zip
Physical Size = 479251

   Date      Time    Attr         Size   Compressed  Name
------------------- ----- ------------ ------------  ------------------------
2021-06-04 18:29:01 ....A        11560         3875  LICENSE.txt
2021-06-04 18:29:01 ....A        12729         5009  README.txt
2021-06-04 18:15:26 ....A       592384       221168  WIN32\ag.exe
2021-06-04 18:15:28 ....A       748544       248639  WIN64\ag.exe
------------------- ----- ------------ ------------  ------------------------
2021-06-04 18:29:01            1365217       478691  4 files
```

libexec/scoop-shim.ps1
niheaven: See https://github.com/ScoopInstaller/Scoop/pull/4736#issuecomment-1041571435

niheaven:
Ready to merge now.

libexec/scoop-list.ps1
```diff
@@ -28,38 +28,38 @@ Write-Host "Installed apps$(if($query) { `" matching '$query'`"}):"
...
         $info += $install_info.architecture
     }
     $item.Info = $info -join ', '
 
-    $list += $item
+    $list += [PSCustomObject]$item
 }
 
-$list.ForEach({[PSCustomObject]$_})
+$list | Add-Member -TypeName 'ScoopApps' -PassThru
```
rashil2000, line 64: Won't this create a new object every time (inefficient)? What's the advantage of doing this vs. specifying the type name on Line 31?

niheaven: No, this just `Add-Member`, i.e., affects only `$list`.

This is more clear, and create normal PSObject unless we need a Table View for the array of ScoopApp item, i.e., we just modify the final output.

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,227 @@
...
+switch ($SubCommand) {
+    'add' {
+        $commandPath = $addArgs[1]
+        if ($commandPath -match "^[`"'](.*?)[`"']\s*(.*)$") {
+            $commandPath = $Matches[1]
+            $commandArgs = $Matches[2]
+        } elseif ($commandPath -match '^([^\s]*)\s*(.*)$') {
+            $commandPath = $Matches[1]
+            $commandArgs = $Matches[2]
+        }
```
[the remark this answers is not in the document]
niheaven: I'll try to handle `-g` by the script itself.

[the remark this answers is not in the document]
rashil2000: Yes that'll be much better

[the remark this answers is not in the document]
niheaven: It is done, but still has bug in CMD. I'll try to fix it in this weekend.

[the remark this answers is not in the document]
niheaven: Yes it does, but what if there's another `-g/--global` switch in your shimmed program? I forced `-g` to be latest one, otherwise it will be used as shimmed command line.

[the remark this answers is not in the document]
rashil2000: 1. If the shimmed program needs `--global`, it can be passed as is because PS does not recognize flags with 2 hyphens.
2. If the shimmed program needs `-g` or `-global`, just put those in quotes. Then they will be passed to the shim.

See this:

image

image


The value of $global is false in all cases, which is what we need.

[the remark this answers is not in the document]
niheaven: Use a direct approach to handle `-g/--global` flag. In syntax, explicitly use global flag in last position, and the script check `$Args[-1]` to see if manipulate global shim.

Test passed in PowerShell and CMD, but for bash, whitespaces in path could not be figured out while `$Args` are separated by whitespaces and no quote is added in `$Args`'s items.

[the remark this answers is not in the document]
rashil2000: Do CMD and Bash behave the same with my method (in comment above) also?

Also could you also confirm for Git Bash (it is slightly different than WSL Bash)?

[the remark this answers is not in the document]
niheaven: It's not related to the arguments parsing, it's COMMAND_PATH parsing... If there are whitespaces in command path, WSL got error.

For arguments, both methods work, and is equivalent IMO.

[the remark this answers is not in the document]
niheaven: Tested in MSYS2 and Cygwin, both got errors as in WSL.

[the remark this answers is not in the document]
rashil2000: > Yes it does, but what if there's another `-g/--global` switch in your shimmed program? I forced `-g` to be latest one, otherwise it will be used as shimmed command line. And if you use `-g` in command path (as arg) and want to make the shim global, `-g -g` is needed.

I did not fully understand this part. Could you give examples of the different cases?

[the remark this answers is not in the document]
niheaven: e.g., assume `test` has a `-g` flag, and if you want to shim `test -g` as global one, you could use `scoop shim add my-test test -g -g` (first `-g` is for `test`, latest one is for scoop shim).

A more complicated case: if `test`'s `-g` flag could accept argument (`test -g arg`), if you use `scoop shim add my-test test -g arg` to add local shim, the `-g` could not be marked as scoop shim's global flag, and you must use `scoop shim add my-test test -g arg -g` to add global one.

So I force `scoop shim`'s global flag (`-g`/`--global`) to be latest one to differ it from command path's `-g` (if existed).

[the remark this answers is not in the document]
rashil2000: > But this could lead to an unexpected case that the only param in command is `-g` and you want a local shim, since `scoop shim add my-test test -g` will make a global shim, and in such situation, scoop could not distinguish command param with scoop shim param... This is a rare use case, IMO.

Yes, this was exactly what I was confused about. Even if it is rare, current implementation will always create a global shim irrespectively.

That is why I suggest using the method I described above. It has the following advantages:
- it is unambiguous. If shim needs `-g`, just pass it in quotes, like `'g'`.
- it does not put restriction that the global flag needs to be the last argument.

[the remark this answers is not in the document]
niheaven: The param syntax is not consistent with other scoop command if using `[Switch]`, just as I commented under `scoop info` patch.

A trick to set `test -g` as local shim is `scoop shim add my-shim test -g ''`, i.e., pass an empty param as latest param.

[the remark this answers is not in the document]
rashil2000: > The param syntax is not consistent with other scoop command if using `[Switch]`, just as I commented under `scoop info` patch.

This is just an implementation detail; users do not need to know how params are implemented internally.

> A trick to set `test -g` as local shim is `scoop shim add my-shim test -g ''`, i.e., pass an empty param as latest param.

I'm personally against tricks/hacks like these, and this behaviour is not user friendly. `scoop shim add my-shim test -g` for global shim, and `scoop shim add my-shim test '-g'` for a local shim with g flag is much more intuitive for a user.

We are implementing this functionality from scratch, so why not do it in a proper way instead of adding workarounds?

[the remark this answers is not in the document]
niheaven: Personally, my only consideration is param syntax. If we use PS switch `-Global`, the `-g` is just a short form of it, and PS thinks `-g`, `-gl`, `-glo`, etc. all as `-global`, should this cause some problems? And PS doesn't support `--global` which needs a special treatment.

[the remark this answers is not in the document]
rashil2000: > Personally, my only consideration is param syntax. If we use PS switch `-Global`, the `-g` is just a short form of it, and PS thinks `-g`, `-gl`, `-glo`, etc. all as `-global`, should this cause some problems? 

All apps other than PS use either `-g` or `--g[full name]`, so I don't think `-gl` or `-glo` should be a problem.

> And PS doesn't support `--global` which needs a special treatment.

Hmm, yes. But we can just mention:
```
#     -g, -global           Create a global shim
```

instead of 
```
#     -g, --global          Create a global shim
```

in the help text, which the users can see and use.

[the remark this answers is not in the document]
niheaven: I'll try to find a way to handle such param, or maybe mix the two methods.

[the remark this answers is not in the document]
niheaven: Some minor tweaks, and result is fine:

image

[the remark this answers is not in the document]
niheaven: Yes, I only added `[switch]$g` to avoid `-g*` mismatch, and handle `--global` in the script.

rashil2000:
Will review later today - I have my exams going on 😅

niheaven:
> Will review later today - I have my exams going on 😅

Aha, good luck!

niheaven:
I'll merge this one and #4734 and plan to next Core release. If there's some bug or improvement, comment here and I'll make hotfix.

rashil2000:
Reviewing now

rashil2000:
I think we can just avoid all the manual handling hassle by doing `[Switch]$global` instead of `[Switch]$g`. Since this is a special case, we will just mention that "Use single hyphen instead of double hyphens". Otherwise trying to mix and match unix arguments in PowerShell will create a lot of pain and edge cases.

niheaven:
PowerShell handle `--arg` with or without quotes, and remove quotes in the args... So,

1. As your comment, using `$global`
2. Add a virtual param, `scoop shim add bash bash --global ''`

rashil2000:
> 2. Add a virtual param, `scoop shim add bash bash --global ''`

How will we tell users about the reasoning/semantics behind this? This would be confusing IMO.

niheaven:
> How will we tell users about the reasoning/semantics behind this? This would be confusing IMO.

Yes, it is. I'll change the syntax and hope there are no bugs.

niheaven:
Check the last commit.

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,241 @@
+# Usage: scoop shim <subcommand> [<shim_names>] [<command_path> [<args>...]] [-g(lobal)]
+# Summary: Manipulate Scoop shims
+# Help: Manipulate Scoop shims: add, remove/rm, list, info, alter, etc.
```
rashil2000, line 3:
```suggestion
# Help: Manipulate Scoop shims: add, rm, list, info, alter, etc.
```

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,241 @@
...
+#
+# To add a costom shim, use the 'add' subcommand:
+#
+#     scoop shim add <shim_name> <command_path> [<args>...]
+#
+# To remove a shim, use the 'remove' or 'rm' subcommand (CAUTION: this could remove shims added by app manifest):
```
rashil2000, line 9:
```suggestion
# To remove a shim, use the 'rm' subcommand (CAUTION: this could remove shims added by an app manifest):
```

For consistency with other scoop commands

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,241 @@
...
+#
+#     scoop shim remove <shim_names>
```
rashil2000, line 11:
```suggestion
#     scoop shim rm <shim_names>
```

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,241 @@
...
+#     scoop shim info <shim_name>
+#
+# To alternate a shim's target source, use the 'alter' subcommand:
+#
+#     scoop shim alter <shim_name>
+#
+# Options:
+#   -g(lobal)       Add/Remove/Info/Alter global shim(s)
+#                   (NOTICE: USING SINGLE DASH)
+#                   (HINT: To pass argument likes '-g' to local shim, using quotes)
```
rashil2000, line 28:
```suggestion
#                   (HINT: To pass arguments like '-g' or '-global' to the shim, use quotes)
```

It can be global shim also

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,241 @@
...
+
+param($SubCommand, $ShimName, [Switch]$global)
+
+. "$PSScriptRoot\..\lib\help.ps1"
+. "$PSScriptRoot\..\lib\install.ps1" # for rm_shim
+
+if ($SubCommand -notin @('add', 'remove', 'rm', 'list', 'info', 'alter')) {
```
rashil2000, line 35:
```suggestion
if ($SubCommand -notin @('add', 'rm', 'list', 'info', 'alter')) {
```

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,241 @@
...
+    'ERROR: <subcommand> must be one of: add, remove/rm, list, info, alter'
```
rashil2000, line 36:
```suggestion
    'ERROR: <subcommand> must be one of: add, rm, list, info, alter'
```

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,241 @@
...
+                    $commandArgs = $Matches[2]
+                } else {
+                    $commandPath = $Args[0]
+                    if ($Args.Length -gt 1) {
+                        $commandArgs = $Args[1..($Args.Length - 1)]
+                    }
+                }
+            }
+        }
+        { $_ -in 'remove', 'rm' } {
```
rashil2000, line 66:
```suggestion
        'rm' {
```

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,241 @@
...
+        if ($commandPath -and (Test-Path $commandPath)) {
+            Write-Host "Adding $(if ($global) { 'global' } else { 'local' }) shim " -NoNewline
+            Write-Host $shimName -ForegroundColor Cyan -NoNewline
+            Write-Host '...'
+            shim $commandPath $global $shimName $commandArgs
+        } else {
+            abort "ERROR: '$($Args[0])' does not exist" 3
+        }
+    }
+    { $_ -in 'remove', 'rm' } {
```
rashil2000, line 147:
```suggestion
    'rm' {
```

rashil2000 (changes requested):
It works now! 

(Tested in PS 7.1.2).

Just some minor comments

libexec/scoop-shim.ps1
niheaven: Why not add `remove` and use `rm` as alias?

rashil2000: I commented below. Basically:

> For consistency with other scoop commands

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,241 @@
...
+    'info' {
+        $shimPath = Get-ShimPath $ShimName $global
+        if ($shimPath) {
+            Get-ShimInfo $shimPath
+        } else {
+            Write-Host "$(if ($global) { 'Global' } else { 'Local' }) shim not found: " -NoNewline
+            Write-Host $ShimName -ForegroundColor Cyan
+            if (Get-ShimPath $ShimName (!$global)) {
+                Write-Host "But a $(if ($global) { 'local' } else {'global' }) shim exists, " -NoNewline
+                Write-Host "run 'scoop shim info $ShimName$(if (!$global) { ' --global' })' to show its info"
```
rashil2000, line 182:
```suggestion
                Write-Host "run 'scoop shim info $ShimName$(if (!$global) { ' -global' })' to show its info"
```

libexec/scoop-shim.ps1
```diff
@@ -0,0 +1,241 @@
...
+                    }
+                }
+                Write-Host 'done.'
+            }
+        } else {
+            Write-Host "$(if ($global) { 'Global' } else { 'Local' }) shim not found: " -NoNewline
+            Write-Host $ShimName -ForegroundColor Cyan
+            if (Get-ShimPath $ShimName (!$global)) {
+                Write-Host "But a $(if ($global) { 'local' } else {'global' }) shim exists, " -NoNewline
+                Write-Host "run 'scoop shim alter $ShimName$(if (!$global) { ' --global' })' to alternate its source"
```
rashil2000, line 233:
```suggestion
                Write-Host "run 'scoop shim alter $ShimName$(if (!$global) { ' -global' })' to alternate its source"
```

libexec/scoop-shim.ps1
niheaven: Oh, I've not noticed. And, why not add `remove` to all other command? It's strange that only `rm` make sense.

Alias could be added in a smarter way, so I'll remove it first and add `remove` to all scoop command in another PR.

rashil2000: `remove` is longer to type 🤣

`rm` has been used since long time because of unix-like tools, it is synonymous with remove

niheaven: Yes, even myself never type `remove` 😄

rashil2000 (approved):

PR #4741, opened 2022-02-16
refactor(scoop-info): Use List View for output

### LINKED ISSUE #4475: [Feature Request] scoop info --version <app>
Currently the command `scoop info <app>` shows a lot of useful information. However, sometimes it would be useful to have a command to show only the version(s).

Examples:
- `scoop info --version <app>`
- `scoop info --version --latest <app>`

Current workaround:
```
# Git for Windows
scoop info <app> | grep "Version: " | sed 's/Version: //'

# PowerShell
((scoop info <app> | Out-String) -split "`r`n" | Select-String "Version: .*").ToString().Replace("Version: ", "")
```

Relates #3969

rashil2000:
#### Description
`scoop-info` was already using Write-Output. This PR just wraps that output in a PSCustomObject.

#### Motivation and Context

Closes #4475

##### Before
image

##### After
image

##### Pipeline
image

#### Checklist:

### DIFF
--- a/libexec/scoop-info.ps1
+++ b/libexec/scoop-info.ps1
@@ -82,53 +82,53 @@ $item.Manifest = $manifest_file
 
 if ($status.installed) {
     # Show installed versions
-    $installed_output = ""
+    $installed_output = @()
     Get-InstalledVersion -AppName $app -Global:$global | ForEach-Object {
-        $installed_output += "`n$(versiondir $app $_ $global)"
+        $installed_output += versiondir $app $_ $global
     }
-    $item.Installed = $installed_output
+    $item.Installed = $installed_output -join "`n"
 }
 
 $binaries = @(arch_specific 'bin' $manifest $install.architecture)
 if ($binaries) {
-    $binary_output = ""
+    $binary_output = @()
     $binaries | ForEach-Object {
         if ($_ -is [System.Array]) {
-            $binary_output += "`n$($_[1]).exe"
+            $binary_output += "$($_[1]).exe"
         } else {
-            $binary_output += "`n$_"
+            $binary_output += $_
         }
     }
-    $item.Binaries = $binary_output
+    $item.Binaries = $binary_output -join "`n"
 }
 $env_set = (arch_specific 'env_set' $manifest $install.architecture)
 $env_add_path = (arch_specific 'env_add_path' $manifest $install.architecture)
 if ($env_set) {
-    $env_vars = ""
+    $env_vars = @()
     $env_set | Get-Member -member noteproperty | ForEach-Object {
         $value = env $_.name $global
         if (!$value) {
             $value = format $env_set.$($_.name) @{ "dir" = $dir }
         }
-        $env_vars += "`n$($_.name) = $value"
+        $env_vars += "$($_.name) = $value"
     }
-    $item.'Environment Variables' = $env_vars
+    $item.'Environment Variables' = $env_vars -join "`n"
 }
 if ($env_add_path) {
-    $env_path = ""
+    $env_path = @()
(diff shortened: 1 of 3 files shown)


rashil2000:
I will add lots of new details to scoop-info - last updated date, remote URL, dependencies, suggestions, support for local filepaths - so this PR is essential to be merged before.

libexec/scoop-info.ps1
```diff
@@ -76,66 +74,63 @@ if ($manifest.license) {
...
-        if($_ -eq '.') {
-            Write-Output "  PATH=%PATH%;$dir"
+        if ($_ -eq '.') {
+            $env_path += "`n$dir"
         } else {
-            Write-Output "  PATH=%PATH%;$dir\$_"
+            $env_path += "`n$dir\$_"
         }
     }
+    $item.'Path Additions' = $env_path
```
niheaven, line 126:
```suggestion
    $item.'Path Appended' = $env_path
```

libexec/scoop-info.ps1
```diff
@@ -76,66 +74,63 @@ if ($manifest.license) {
...
-        $dir = versiondir $app $_ $global
-        if($global) { $dir += " *global*" }
-        Write-Output "  $dir"
+    $installed_output = ""
+    Get-InstalledVersion -AppName $app -Global:$global | ForEach-Object {
+        $installed_output += "`n$(versiondir $app $_ $global)"
     }
-} else {
-    Write-Output "Installed: No"
+    $item.Installed = $installed_output
```
niheaven, line 89: Why add so many blank lines?

niheaven: This should be better:

image

libexec/scoop-info.ps1
rashil2000: Append means adding at the end, which is not true. Scoop actually "prepends" the PATH variable. But in the end, it becomes somewhere in between because the machine wide PATH takes a priority. So, it is ambiguous and hence "Addition".

libexec/scoop-info.ps1
rashil2000: It will become more clear when you have more than one versions installed. I was trying to be similar to how it was before (which had a newline in beginning).

libexec/scoop-info.ps1
niheaven: So why not 'Path Added'? Path Addition sounds strange.

rashil2000: The property name should be a noun, right? "Added" is verb, "Addition" is noun.

libexec/scoop-info.ps1
niheaven: Since you use List View now, the output with blank line looks like a misformatted table. For old string-like output, indentation is suitable to clarify long sentence, but for the list (you've used a list with wider first column), line width is not worth considering.

libexec/scoop-info.ps1
niheaven: 😂 Path added means 'Path to be added', it's noun.

libexec/scoop-info.ps1
rashil2000: Okay, will change it. But we need to note that if I remove the blank line in beginning, it will be a breaking change for users who relied on non-PowerShell tools (such as `grep` and `findstr`) to gather information from the scoop-info command.

niheaven: This command is used mostly in console or single line command IMO, and is a easy-to-found format change, not breaking. We could change it, the sooner, the better.

rashil2000:
Done.

niheaven (approved):

PR #4742, opened 2022-02-16
feat(scoop-cat): Use `bat` to pretty-print JSON

rashil2000:
#### Description

Silly new feature of scoop-cat 🤣

#### Motivation and Context

Your manifest looks cute now xD

##### Before

image

##### After

image

#### Checklist:

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -34,10 +34,11 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
         return
     }
 
-    if ((get_config 'manifest-review' $false) -and ($MyInvocation.ScriptName -notlike '*scoop-update*')) {
+    if ((get_config 'manifest_review' $false) -and ($MyInvocation.ScriptName -notlike '*scoop-update*')) {
         "Manifest: $app.json"
-        if (Get-Command bat -CommandType Application -ErrorAction Ignore) {
-            $manifest | ConvertToPrettyJson | bat -pp -l json
+        $style = get_config cat_style ''
+        if (![String]::IsNullOrWhiteSpace($style)) {
+            $manifest | ConvertToPrettyJson | bat --no-paging --style $style --language json
         } else {
             $manifest | ConvertToPrettyJson
         }
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -37,7 +37,7 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
     if ((get_config 'manifest-review' $false) -and ($MyInvocation.ScriptName -notlike '*scoop-update*')) {
         "Manifest: $app.json"
         if (Get-Command bat -CommandType Application -ErrorAction Ignore) {
-            $manifest | ConvertToPrettyJson | bat --no-paging --language json
+            $manifest | ConvertToPrettyJson | bat -pp -l json
         } else {
             $manifest | ConvertToPrettyJson
         }
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -35,9 +35,9 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
     }
 
     if ((get_config 'manifest_review' $false) -and ($MyInvocation.ScriptName -notlike '*scoop-update*')) {
-        "Manifest: $app.json"
-        $style = get_config cat_style ''
-        if (![String]::IsNullOrWhiteSpace($style)) {
+        Write-Host "Manifest: $app.json"
+        $style = get_config cat_style
+        if ($style) {
             $manifest | ConvertToPrettyJson | bat --no-paging --style $style --language json
         } else {
             $manifest | ConvertToPrettyJson
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -1,5 +1,5 @@
-. "$psscriptroot/autoupdate.ps1"
-. "$psscriptroot/buckets.ps1"
+. "$PSScriptRoot/autoupdate.ps1"
+. "$PSScriptRoot/buckets.ps1"
 
 function nightly_version($date, $quiet = $false) {
     $date_str = $date.tostring("yyyyMMdd")
@@ -35,8 +35,12 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
     }
 
     if ((get_config 'manifest-review' $false) -and ($MyInvocation.ScriptName -notlike '*scoop-update*')) {
-        Write-Output "Manifest: $app.json"
-        Write-Output $manifest | ConvertToPrettyJson
+        "Manifest: $app.json"
+        if (Get-Command bat -CommandType Application -ErrorAction Ignore) {
+            $manifest | ConvertToPrettyJson | bat --no-paging --language json
+        } else {
+            $manifest | ConvertToPrettyJson
+        }
         $answer = Read-Host -Prompt "Continue installation? [Y/n]"
         if (($answer -eq 'n') -or ($answer -eq 'N')) {
             return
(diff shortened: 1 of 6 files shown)


niheaven:
Since 'bat' is somehow slower, use config option such as `CAT_USE_BAT` or `VIEW_USE_BAT` to determine whether using bat even if existed.

Add bat option `--style=grid,numbers` to not show `STDIN` file name. (Have you set `BAT_STYLE`?)

image

rashil2000:
> Since 'bat' is somehow slower,

Will it really matter? We just show one json file at a time, which is hardly 2-3KB.

> use config option such as `CAT_USE_BAT` or `VIEW_USE_BAT` to determine whether using bat even if existed.

I can't find any docs on them. How to use them?

> Add bat option `--style=grid,numbers` to not show `STDIN` file name. (Have you set `BAT_STYLE`?)

Haha yes, I forgot that I had a bat config file. 
Fixed. 
I have used the `-pp` option to keep it visually similar to how it was before.

niheaven:
If it looks the same, why use `bat`? line numbers and grid line are pretty 😄 , but yes, highlighting is the most wonderful one and existed.

For the speed, in my PC (i5-8500/32G), `bat` is slower 20x-30x, i.e., direct output take milliseconds while `bat` take seconds.

> I can't find any docs on them. How to use them?

I mean add new config option to switch the background `scoop cat` uses, i.e., check the config is TRUE and `bat` existed, while the config's default is FALSE.

rashil2000:
> If it looks the same, why use `bat`? line numbers and grid line are pretty 😄 , but yes, highlighting is the most wonderful one and existed. (Hmm, why your terminal doesn't show color? `BAT_STYLE`?)

Grid lines and splits are useful for git output and multiple files. I suppose line numbers are useful, but I don't like them personally. 

My terminal is showing colors (notice the escaped backslashes). I'm using --theme=ansi (because it works in both light and dark terminals).

> For the speed, in my PC (i5-8500/32G), `bat` is slower 20x-30x, i.e., direct output take milliseconds while `bat` take seconds.

> I mean add new config option to switch the background `scoop cat` uses, i.e., check the config is TRUE and `bat` existed, while the config's default is FALSE.

I think I have a genius solution to both these problems: a config variable called `cat_style`. The value of this variable can be:
- **empty**: in this case bat will not be used.
- **not empty**: if set to any of the supported bat styles (see below), it will use bat and pass this value as `--style`.

image

How does that sound?

niheaven:
And `bat` should be installed if `cat_style` is set.

BTW, it should be named `CAT_STYLE`, `cat_style` or `cat-style`? The scoop option names now are a bit messy.

rashil2000:
> And `bat` should be installed if `cat_style` is set.

I'm not sure if we should do this automatically. What if the user doesn't have internet, and just wants to see local manifests? Downloading something for a local functionality seems bad. We can just detect if doesn't exist and use the normal Write-Output in that case. 

> BTW, it should be named `CAT_STYLE`, `cat_style` or `cat-style`? The scoop option names now are a bit messy.

To stick with a convention, I think we should use small case and underscores from now on.

niheaven:
Just like `aria2`, there're two choices. One for automatic enable `bat` unless explicit disabled, while another disable `bat` by default and user could enable it.

For the ones that don't have Internet access, why they want to set `cat_style`? Make notes for this option that it needs Internet.

Treat it just as existed ones, e.g., `aria2` and `lessmsi`. The former is enabled by default, while the latter is enabled by option.

rashil2000:
aria2 and lessmsi are used for downloading and installing apps, which automatically needs internet connection.

I'm against the idea of making an offline functionality require being "online" (even if it's just one time setup).

In the config, I will mention that bat is mandatory if the bat_style option is set (I will also mention in the docs that `This option requires bat. Do 'scoop install bat' to install it.` We also show a similar message for 'sudo' during scoop-checkup.)

niheaven:
Okay, that works fine. Besides, if user has `bat` installed, it should be enabled by default or manually? i.e., the option should be an enable one (`lessmsi`) or disable one (`aria2`)?

rashil2000:
I was initially going for enabled by default, but as you mentioned that bat might be slow on some PCs, we should make it manually enabled.

niheaven:
Okay, please make changes and I'll review again.

rashil2000:
Done

niheaven:
Hmm, could you please separate config underscores to another PR?

rashil2000:
Okay sure

lib/install.ps1
```diff
@@ -34,9 +34,14 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
         return
     }
 
-    if ((get_config 'manifest-review' $false) -and ($MyInvocation.ScriptName -notlike '*scoop-update*')) {
-        Write-Output "Manifest: $app.json"
-        Write-Output $manifest | ConvertToPrettyJson
+    if ((get_config 'manifest_review' $false) -and ($MyInvocation.ScriptName -notlike '*scoop-update*')) {
```
niheaven, line 37:
```suggestion
    if ((get_config 'manifest-review' $false) -and ($MyInvocation.ScriptName -notlike '*scoop-update*')) {
```

lib/install.ps1
```diff
@@ -34,9 +34,14 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
...
+        "Manifest: $app.json"
```
niheaven, line 38:
```suggestion
        Write-Host "Manifest: $app.json"
```

Do we need this line for pipeline? If not, use `Write-Host`

lib/install.ps1
```diff
@@ -34,9 +34,14 @@ function install_app($app, $architecture, $global, $suggested, $use_cache = $tru
...
+        $style = get_config cat_style ''
+        if (![String]::IsNullOrWhiteSpace($style)) {
```
niheaven, line 40:
```suggestion
        if (get_config 'cat_style') {
```

image

libexec/scoop-cat.ps1
```diff
@@ -15,7 +15,12 @@ $app, $bucket, $null = parse_app $app
 $app, $manifest, $bucket, $url = Find-Manifest $app $bucket
 
 if ($manifest) {
-        $manifest | ConvertToPrettyJson | Write-Host
+        $style = get_config cat_style ''
+        if (![String]::IsNullOrWhiteSpace($style)) {
```
niheaven, line 19:
```suggestion
        if (get_config 'cat_style') {
```

See above comment

lib/install.ps1
rashil2000: This is a bug. The value should be `manifest_review`. See https://github.com/ScoopInstaller/Scoop/blob/d1f828c94275641b86838e90a041af3315fb7d0f/libexec/scoop-config.ps1#L59

lib/install.ps1
rashil2000: Sure

lib/install.ps1
rashil2000: $style variable needs to be defined. It is used in the pipeline

lib/install.ps1
niheaven: Aha, yes

lib/install.ps1
niheaven: Okay, I'll edit the commit.

lib/install.ps1
niheaven: See edited comment.

rashil2000: Write-Host is better, since we don't need to pass it to pipeline

niheaven: Hmm, I thought you'll use this line... 🤣 

Then `Write-Host`

PR #4744, opened 2022-02-17
fix(scoop-info): Use pipe to separate bins

rashil2000:
Hotfix for #4741

### Before

image

### After

image


niheaven:
Maybe `,` is better? `|` makes me dizzy 🤣

rashil2000:
`,` is a valid character in filename, so it will cause problems. Actually, the only usable characters were `:` and `|`. But since `:` is already used for drive letters, I went with `|`.

PR #4756, opened 2022-02-24
fix(scoop-bucket): Check for git before running command

rashil2000:
Hotfix for #4704

PR #4759, opened 2022-02-24
fix(scoop-bucket): Fix date formatting issues with git bucket(s)

HUMORCE:
#### Motivation and Context
Relates to https://github.com/ScoopInstaller/Scoop/pull/4756

#### Checklist:


rashil2000 (dismissed):

rashil2000:
One more thing. I don't think we need quotes for 
```
"`"$source`"" 
```
variable. Meaning we can just say 
```
$source
```

while calling git. Can you check once?

HUMORCE:
you are right, the git command still works fine after removing them, but many code still uses quotes like this.

so, i will not change them in this PR.

rashil2000:
I actually wanted to remind you in your original buckets PR, but I forgot 😅

The only places where double quotes are needed are those where we use `git_cmd` function, because that function is run by cmd.exe. For generic `git` calls, parameters are handled by PowerShell so quoting is not needed.

HUMORCE:
okay, i did not notice that.

also removed them from L18 of `scoop-status.ps1`.

rashil2000 (dismissed):

niheaven:
Please update CHANGELOG, thanks

PR #4765, opened 2022-02-27
feat(scoop-config): Show all settings

### LINKED ISSUE #4154: How to export/backup list of config/settings variables?
```powershell
scoop config MSIEXTRACT_USE_LESSMSI true
scoop config aria2-enabled false
```

I want to backup changes, how to list all configurations with value?

rashil2000:
#### Description

Show all configuration options at once.

#### Motivation and Context

Closes #4154

image

#### Checklist:

### DIFF
--- a/libexec/scoop-config.ps1
+++ b/libexec/scoop-config.ps1
@@ -2,6 +2,10 @@
 # Summary: Get or set configuration values
 # Help: The scoop configuration file is saved at ~/.config/scoop/config.json.
 #
+# To get all configuration settings:
+#
+#     scoop config
+#
 # To get a configuration setting:
 #
 #     scoop config <name>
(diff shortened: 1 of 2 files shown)


libexec/scoop-config.ps1
```diff
@@ -123,25 +127,27 @@
...
 reset_aliases
 
-if(!$name) { my_usage; exit 1 }
-
-if($name -like 'rm') {
+if (!$name) {
+    $scoopConfig
+} elseif ($name -like '--help') {
+    my_usage
+} elseif ($name -like 'rm') {
```
niheaven, line 139:
```suggestion
} elseif ($name -like 'rm' -or $null -eq $value) {
```

What do you think of letting `scoop config <config> $null` to remove the config too?

rashil2000: Hmm, I think it would get confusing. Both `scoop config rm hello` and `scoop config hello $null` would be redundant. Also $null is not a recognized symbol in CMD/Bash?

niheaven: Okay, yes, in BASH/CMD `$null` may not be recognized.

niheaven (approved):

PR #4773, opened 2022-03-03
fix(scoop-search): Remove redundant 'bucket/' in search result

### LINKED ISSUE #4772: [Bug] Remote Search Result Display Redundant
When searching from added buckets fails to generate results, scoop will search known but not added buckets online (in this case, games). However:
pic
"bucket" here is redundant.

zStruCat:
Closes #4772 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 - **scoop-bucket:** Fix date formatting issues with git bucket(s) ([#4759](https://github.com/ScoopInstaller/Scoop/issues/4759))
 - **scoop-checkup:** Skip 'check_windows_defender' when have not admin privileges ([#4699](https://github.com/ScoopInstaller/Scoop/issues/4699))
 - **scoop-cleanup:** Remove apps other than current version ([#4665](https://github.com/ScoopInstaller/Scoop/issues/4665))
+- **scoop-search:** Remove redundant 'bucket/' in search result ([#4773](https://github.com/ScoopInstaller/Scoop/issues/4773))
 - **scoop-update:** Skip updating non git buckets ([#4670](https://github.com/ScoopInstaller/Scoop/issues/4670), [#4672](https://github.com/ScoopInstaller/Scoop/issues/4672))
 
 ### Performance Improvements
--- a/libexec/scoop-search.ps1
+++ b/libexec/scoop-search.ps1
@@ -48,28 +48,26 @@ function search_bucket($bucket, $query) {
 }
 
 function download_json($url) {
-    $progressPreference = 'silentlycontinue'
-    $result = invoke-webrequest $url -UseBasicParsing | Select-Object -exp content | convertfrom-json
-    $progressPreference = 'continue'
+    $ProgressPreference = 'SilentlyContinue'
+    $result = Invoke-WebRequest $url -UseBasicParsing | Select-Object -ExpandProperty content | ConvertFrom-Json
+    $ProgressPreference = 'Continue'
     $result
 }
 
 function github_ratelimit_reached {
-    $api_link = "https://api.github.com/rate_limit"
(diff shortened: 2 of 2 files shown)


zStruCat:
> Actually, the `bucket/` is not redundant. Sometimes multiple buckets have same package, so we specify `scoop install <bucket>/<app>` to install a particular app.
> 
> It would be better to replace the `bucket/` word by the actual name of the bucket, like `games/`.

The display of searched manifests is divided into several section,  each belonging to a particular bucket with its name displayed at the begining of the section, so perhaps there's no need for it to appear again in the manifest.

pid

rashil2000:
Then we should leave it as it is, because the instruction `scoop install <bucket>/<app>` is still important

zStruCat:
Also, it's consistent with the display format of local bucket
图片

Off the topic, i do think `scoop install bucket/app` is important. To be more precise, i feel that the form of bucket/app@version as a consistent input query is vital. Unfortunately, different command of scoop has acted inconsistently. For instance, search and list doesn't support bucket/app yet......

rashil2000:
I think a common middle ground would be to make this result:
```
Results from other known buckets...
(add them using 'scoop bucket add <name>')

'nonportable' bucket:
    bucket/qttabbar-np
```

into
```
Results from other known buckets...
(add them using 'scoop bucket add <name>')

'nonportable' bucket (install using 'scoop install nonportable/<app>'):
    qttabbar-np
```

This will make it consistent with local bucket results.

zStruCat:
> I think a common middle ground would be to make this result:
> 
> ```
> Results from other known buckets...
> (add them using 'scoop bucket add <name>')
> 
> 'nonportable' bucket:
>     bucket/qttabbar-np
> ```
> 
> into
> 
> ```
> Results from other known buckets...
> (add them using 'scoop bucket add <name>')
> 
> 'nonportable' bucket (install using 'scoop install nonportable/<app>'):
>     qttabbar-np
> ```
> 
> This will make it consistent with local bucket results.

This is exactly what my commit do.
pic

zStruCat:
> Notice the additional text beside the bucket header, that is not present in your commit

But `scoop install` cannot install app from a known but not added bucket. It results in the following:
`
WARN  Bucket 'games' not installed. Add it with 'scoop bucket add games' or 'scoop bucket add games <repo>'.
Couldn't find manifest for '0ad' from 'games' bucket.
`

rashil2000:
> But scoop install cannot install app from a known but not added bucket. It results in the following:

Instruction for adding other known buckets is already given at the beginning of results, right

zStruCat:
changed accordingly

niheaven:
You should make `bucket/` optional since it is not required in bucket repo.

rashil2000:
> You should make `bucket/` optional since it is not required in bucket repo.

Do you mean the hint should only be shown for not added buckets?

niheaven:
> Do you mean the hint should only be shown for not added buckets? 

No, I means the regex. Not all bucket repo has manifest files in 'bucket' subfolder, they may be put in root folder and that's scoop's original bucket structure.

zStruCat:
> > Do you mean the hint should only be shown for not added buckets?
> 
> No, the regex. Not all bucket repo has manifest files in 'bucket' subfolder, they may be put in root folder and that's scoop's original bucket structure.

First, currently the remote search only search the remote 'known' bucket, i.e. main, extras, nirsoft, etc, all of which put manifests under bucket directory.
Secondly, if scaning all files, we won't be able to distinguish between a manifest and a normal json file without downloading and check them with the schema. Is the time-cost here acceptable?

zStruCat:
Any other issues?

niheaven:
I've tweaked some code and now LGTM.

niheaven (approved):

Issue #2123, opened 2018-03-14
Disallow adding the same bucket more than once


rasa:
This is allowed, currently:
```
scoop bucket add comp500 https://github.com/comp500/scoop-comp500.git
scoop bucket add anothername https://github.com/comp500/scoop-comp500.git
```
We should fail with a message, like:
```
> scoop bucket add anothername https://github.com/comp500/scoop-comp500.git
ERROR: Bucket comp500 already exists for https://github.com/comp500/scoop-comp500.git
```

PR #4814, opened 2022-03-15
refactor(bucket): Move 'Find-Manifest' and 'list_buckets' to 'buckets'

niheaven:
#### Description

`find_manifest()` and `Find-Manifest()` (former `locate()`) have similar functionality and now be merged into `Find-Manifest()`.

Also move `list_buckets()` from `libexec\scoop-bucket.ps1` to `lib\buckets.ps1`.

#### Motivation and Context

Discussed with @rashil2000 

`scoop bucket` and `scoop 

#### Checklist:

### DIFF
--- a/lib/buckets.ps1
+++ b/lib/buckets.ps1
@@ -118,7 +118,7 @@ function Convert-RepositoryUri {
         } else {
             error "$Uri is not a valid Git URL!"
             error "Please see https://git-scm.com/docs/git-clone#_git_urls for valid ones."
-            exit 1
+            return $null
         }
     }
 }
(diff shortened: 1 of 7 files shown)


rashil2000:
Nice change!

Since we're refactoring bucket methods, can we also incorporate https://github.com/ScoopInstaller/Scoop/pull/3348? This will allow us to close #2123

niheaven:
Okay, wait a moment.

niheaven:
Done. Now check for duplicated buckets.

lib/buckets.ps1
```diff
@@ -103,6 +103,24 @@ function Find-Manifest($app, $bucket) {
...
 }
 
+function Convert-RepositoryUri {
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory, Position = 0, ValueFromPipeline = $true)]
+        [String] $Uri
+    )
+
+    process {
```
rashil2000, line 113: We can omit the `process` block for such a small function

niheaven: Enabling pipeline needs explicit `process`, and in fact, if not specified, the script block is put under `begin`.

rashil2000: Oh, didn't know that 😅

rashil2000:
image

The bucket-add function should stop at the first error (`ERROR github.com/ScoopInstaller/Extras is not a valid Git URL!`), no?

niheaven:
So you mean we should assume default protocol? The error you met is no protocol provided (it is http? https? ftp? ftps? ssh?)

rashil2000:
I meant that we're showing the error twice. When we get an error (any error) from the `Convert-RepositoryUri` function, we should stop the `add_bucket` function (instead of proceeding to the `Checking repo...` step)

niheaven:
Oh, sorry, haven't noticed that, will change it later.

lib/buckets.ps1
```diff
@@ -117,6 +117,8 @@ function Convert-RepositoryUri {
             $Matches.provider, $Matches.user, $Matches.repo -join '/'
         } else {
             error "$Uri is not a valid Git URL!"
+            error "Please see https://git-scm.com/docs/git-clone#_git_urls for valid ones."
+            exit 1
```
rashil2000, line 121: Exiting from script might not be desired if we use this function somewhere else. Instead we can check something like last error code or return value of function.

rashil2000:
Since the location of `Find-Manifest` has changed, has the import of this function in all other source files been taken care of?

niheaven:
> Since the location of Find-Manifest has changed, has the import of this function in all other source files been taken care of?

`buckets.ps1` is imported in every `lib-exec` scripts, so it's not a problem.

lorengordon:
Traced this patch as the cause of this issue #4917

PR #4842, opened 2022-03-25
feat(config): Try SCOOP_GH_TOKEN value first before gh_token value from config

astelmachonak-nydig:
#### Description
There are multiple usages of GitHub token value in the scoop (in checkver.ps1, core.ps1, install.ps1), but only in once place (in checkver.ps1), there is a logic to try using token value from `SCOOP_GH_TOKEN` environment variable first. This PR unifies the token value obtain logic and puts it into the function, which is used everywhere token value is needed.

#### Motivation
Using token in environment variable is needed for GitHub Actions setup (I am setting up GitHub Actions for private bucket right now and token is required for accessing artifacts located in private repository).

Tested with private GitHub repositories using both `gh_token` and `SCOOP_GH_TOKEN` values.

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 ### Features
 
 - **install:** Allow downloading from private repositories ([#4254](https://github.com/ScoopInstaller/Scoop/issues/4243))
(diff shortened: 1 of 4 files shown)


rashil2000 (approved):

rasa (approved):
Untested but LGTM

niheaven (approved):

PR #4869, opened 2022-04-13
Dual license between Unlicense and MIT

### LINKED ISSUE #4868: [Feature] Do not use the Unlicense for scoop
## Feature Request

**Note: This is not legal advice. I am not a lawyer.**
#### Is your feature request related to a problem? Please describe.

Scoop is currently using the Unlicense and was using it since development of it began (that means Scoop is released to the public domain). I have an issue with Scoop's license - the Unlicense is **invalid** in some countries (e.g. Germany), and does not seem to be clear in a few countries (e.g. Australia). This is a problem with the Unlicense, so we should be moving to MIT or Apache because "public domain" does not exist everywhere. Also, [the Unlicense is not allowed to be used inside Google](https://opensource.google/docs/patching/#forbidden), so Google would not be allowed to make contributions to Scoop or fork it to @google. However, Google employees are important contributors to open-source, so always stay away from the Unlicense. There are also rumours going on in github/choosealicense.com#805.

This should apply to all Scoop repositories using the Unlicense (yes, that means most of the repositories).

Note that [0BSD](https://choosealicense.com/licenses/0bsd) is an exception.
#### Describe the solution you'd like

I would edit the `LICENSE` file to use a permissive license instead of the Unlicense or any other public domain dedication.
#### Describe alternatives you've considered

Using Chocolatey instead. I am also writing [CPPM](https://github.com/Tyler887/CPPM), a package manager not only for Windows, but for _any operating system that supports Python_. It's not released into the public domain but is under a permissive license.

Tyler887:
#### Description
Patches the discussion at #4868.

#### Motivation and Context

Closes #4868

Only modifications to non-source files, not a code update, so I can't test it.

#### Checklist:


Tyler887:
I recently updated the branch name so I will try opening the PR again.

PR #4915, opened 2022-05-10
fix(depends): Avoid digits in archive file extension

issaclin32:
* This avoids extraction error for 7-zip split archives like `xxx.7z.001`, `xxx.7z.002`, ....

for example:

```
"url": [
    "https://someapp.com/files/some-app.7z.001",
    "https://someapp.com/files/some-app.7z.002"
]
```

In this case. `some-app.7z.001` extracts properly but `some-app.7z.002` is treated as another independent archive, resulting in extraction error.

* This fix would be useful when we need to split files in ScoopInstaller/Binary (into multiple 7z archives) to avoid using Git LFS.


issaclin32:
By the way, this fix would be useful when we need to split files in `ScoopInstaller/Binary` (into multiple 7z archives) to avoid using Git LFS.

niheaven:
After this is applied, are `.00x` removed after 7z extracted the files?

issaclin32:
> After this is applied, are `.00x` removed after 7z extracted the files?

Missed that. Thanks for reminding me.

lib/decompress.ps1
```diff
@@ -65,6 +65,11 @@ function Expand-7zipArchive {
     if ($Removal) {
         # Remove original archive file
         Remove-Item $Path -Force
+        if ([System.IO.Path]::GetExtension($Path) -eq '.001') {
+            # Remove remaining 7-zip archive parts
+            $FileNameWithoutExt = [System.IO.Path]::GetFileNameWithoutExtension($Path)
+            Get-ChildItem $([System.IO.Path]::GetDirectoryName($Path)) | Where {$_ -match ([Regex]::Escape($FileNameWithoutExt)+'\.\d{3}$')} | Remove-Item
+        }
```
niheaven, line 72:
```suggestion
        if (($Path -replace '.*\.([^\.]*)$', '$1') -eq '001') {
            # Remove remaining 7-zip archive parts
            Get-ChildItem "$($Path -replace '\.[^\.]*$', '').???" | Remove-Item -Force
        } else {
            Remove-Item $Path -Force
        }
```

niheaven (commented):
1. Using `-replace` may be more readable
2. I've updated the test scripts and test cases, please review them

issaclin32:
> 2. I've updated the test scripts and test cases, please review them

The test scripts & cases works fine for me. Thank you.

PR #4941, opened 2022-05-20
fix(core): Use Invoke-Command instead of Invoke-Expression

niheaven:
#### Description

Get rid of `Invoke-Expression` that may cause problems and may leak some code.

Ref: https://github.com/MicrosoftDocs/PowerShell-Docs/wiki/The-case-for-and-against-Invoke-Expression

#### Motivation and Context

- Closes https://github.com/ScoopInstaller/Extras/issues/8494
- Closes https://github.com/ScoopInstaller/Extras/issues/8498
- Closes https://github.com/ScoopInstaller/GithubActions/pull/8

- `scoop -v`
- `scoop update`
- `scoop install gnupg`
- `.\bin\checkver i*` in 'extras' bucket
- `.\test.ps1`

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Bug Fixes
 
+- **core:** Using `Invoke-Command` instead of `Invoke-Expression` ([#4941](https://github.com/ScoopInstaller/Scoop/issues/4941))
 - **core:** Load config file before initialization ([#4932](https://github.com/ScoopInstaller/Scoop/issues/4932))
 - **depends:** Avoid digits in archive file extension (except for .7z and .001) ([#4915](https://github.com/ScoopInstaller/Scoop/issues/4915))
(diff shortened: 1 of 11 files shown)


rashil2000:
There are multiple instances of such commands:
```pwsh
$currentBranch = Invoke-Command { git -C "$currentdir" branch }
```

Why not just replace them with:
```pwsh
$currentBranch = git -C "$currentdir" branch
```

niheaven:
`Invoke-Command` run the script block in a child scope, so it should be safer and could improve readability.

rashil2000:
Scope doesn't matter for git (and other similar external commands) right? On the other hand, I feel it hampers readability because it makes it seem like git commands need special handling. 

For strings containing PowerShell code, it makes sense.

niheaven:
Great, let's try and see if there are unexpected bugs.

niheaven:
Removed `Invoke-Command` before `git`, and tested okay here.

rashil2000 (approved):

PR #4950, opened 2022-05-24
chore(core): Deprecate tls1 and tls1.1

neheb:
Both of these are deprecated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Bug Fixes
 
+- **chore:** Deprecate tls1 and tls1.1 ([#4950](https://github.com/ScoopInstaller/Scoop/pull/4950))
 - **chore:** Update Nonportable bucket URL ([#4955](https://github.com/ScoopInstaller/Scoop/issues/4955))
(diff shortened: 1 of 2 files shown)


neheb:
The interesting thing on my Windows 8.1 machine is that TLS1.3 is available. So .Net 4.8 is installed?

neheb:
Interestingly enough,

 ```
if (!($isNewerNetFramework -and $isSystemDefault)) {
```

evaluates to true on my machine. The weird thing is, [System.Net.ServicePointManager]::SecurityProtocol is set to SystemDefault. It shouldn't be evaluating to true and overwriting SystemDefault.

neheb:
Done.

niheaven:
I'm not familiar with TLS and .net, but for Win 7 and .net 4.5, should this cause errors?

And if we update minimal requirement to .net 4.7, this TLS sector could be removed IMO.

neheb:
.net 4.5 supports tls 1.2. There's no change here other than removing 1.0 and 1.1.

I will say that even this is not enough to fix https://github.com/ScoopInstaller/Main/issues/3593 . Maybe the issue needs to be reported to curl developers.

chawyehsu (approved):

neheb:
Done.

chawyehsu (approved):

niheaven (approved):

Issue #4324, opened 2021-04-17
checkver tool encoding problem


everyx:
I need extract version number from a page with CJK characters, and it does not work correctly, because the `webclient.Encoding` is not set.

should we add a `checkver.encoding` option in App-Manifests file?

```json
"checkver": {
    "url": "https://www.todesk.com/download.html",
    "re": "当前版本:(\\d+.\\d+.\\d+)",
    "encoding": "UTF8"
},
```

and set `webclient` encoding in `checkver.ps1`

https://github.com/lukesampson/scoop/blob/643cfb0da8369dbbe2964511698763c5ff179492/bin/checkver.ps1#L99

```powershell
    $encoding = 'utf-8'
    if ($json.checkver.encoding) {
        $encoding = $json.checkver.encoding
    }
    $wc.Encoding = [System.Text.Encoding]::GetEncoding($json.checkver.encoding)
```

HUMORCE:
```pwsh
$ cat .\bucket\todesk.json      
{
    "version": "0",
    "checkver": {
        "url": "https://www.todesk.com/download.html",
        "regex": "当前版本:([\\d.]+)"
    }
}
$ .\bin\checkver.ps1 -App todesk
todesk: 3.0.0 (scoop version is 0)
DONE
```

works fine.

everyx:
> ```powershell
> $ cat .\bucket\todesk.json      
> {
>     "version": "0",
>     "checkver": {
>         "url": "https://www.todesk.com/download.html",
>         "regex": "当前版本:([\\d.]+)"
>     }
> }
> $ .\bin\checkver.ps1 -App todesk
> todesk: 3.0.0 (scoop version is 0)
> DONE
> ```
> 
> works fine.

@HUMORCE Hi, what's your default Active code page or language? I use scoop in my win10 build 19042 simplified chinese version, Maybe different language versions have different default encodings?

HUMORCE:
same build in english but **unicode utf-8 for worldwide language support** enabled.

```
$ [System.Console]::InputEncoding

Preamble          :
BodyName          : utf-8
EncodingName      : Unicode (UTF-8)
HeaderName        : utf-8
WebName           : utf-8
WindowsCodePage   : 1200
IsBrowserDisplay  : True
IsBrowserSave     : True
IsMailNewsDisplay : True
IsMailNewsSave    : True
IsSingleByte      : False
EncoderFallback   : System.Text.EncoderReplacementFallback
[...]

PR #4956, opened 2022-05-28
feat(core): Add 'Get-Encoding()' function to fix missing webClient encoding

### LINKED ISSUE #4911: [Bug] Can not install with url encoded in `utf-8`
#### Current Behavior

图片

#### Expected Behavior

The file is encoding with utf-8, but the `Webclient` use `default` encoding, and it is not recommended.

See more in [system.net.webclient.encoding](https://docs.microsoft.com/en-us/dotnet/api/system.net.webclient.encoding?view=net-6.0#system-net-webclient-encoding) and [system.text.encoding.default](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.default?view=net-6.0#system-text-encoding-default).

#### Possible Solution

We can solve this problem by set the `Encoding` for `Webclient`.

### System details

**Windows version:** [e.g. 7, 8, 10]

win10 

**OS architecture:** [e.g. 32bit, 64bit]

64bit

5.1.19041.1645

#### Scoop Configuration

default setting.

yi-Xu-0100:
Add Get-Encoding function in core.ps1

#### Motivation and Context

Closes #4911 #4324

Relates to #4911 #4324

#### Checklist:


yi-Xu-0100:
@rashil2000 Any review here will be helpful~ 😁

lib/core.ps1
```diff
@@ -15,6 +15,15 @@ function Optimize-SecurityProtocol {
     }
 }
 
+function Get-Encoding($url) {
```
niheaven, line 18: Pass `$wc` in the function

lib/manifest.ps1
```diff
@@ -1,3 +1,4 @@
+. "$PSScriptRoot\..\lib\core.ps1"
 function manifest_path($app, $bucket) {
```
niheaven, line 2:
```suggestion
function manifest_path($app, $bucket) {
```

libexec/scoop-virustotal.ps1
```diff
@@ -34,6 +34,7 @@
 #   -n, --no-depends          By default, all dependencies are checked too. This flag avoids it.
 #   -u, --no-update-scoop     Don't update Scoop before checking if it's outdated
 
+. "$PSScriptRoot\..\lib\core.ps1"
 . "$PSScriptRoot\..\lib\getopt.ps1"
```
niheaven, line 38:
```suggestion
. "$PSScriptRoot\..\lib\getopt.ps1"
```

niheaven (changes requested):

niheaven:
Do you have some testing manifests?

yi-Xu-0100:
@niheaven ,Thanks for your reply. 😁

In https://github.com/ScoopInstaller/Scoop/issues/4911#issue-1229175577 , I use 

```powershell
scoop install "https://raw.githubusercontent.com/ygguorun/scoop-bucket/master/bucket/PDFPatcher.json"
```

, and [this manifest - DoveBoy/Apps/pdfpatcher](https://raw.githubusercontent.com/DoveBoy/Apps/master/bucket/pdfpatcher.json) has the `checkver`, and they all work well.

niheaven:
`checkver` has some error, could you please check it again?

yi-Xu-0100:
@niheaven  I used it for `scoop update` and `scoop install`, they all work well. 

Could you please provider the way to test the `checkver`, I have no idea for it.🤔

niheaven:
Checkout this branch and cd in any bucket, run `.\bin\checkver.ps1 xxx`, there are some bugs :)

yi-Xu-0100:
@niheaven Thanks for pointing out the problem, I think it's fixed now. 😁

niheaven (approved):

yi-Xu-0100:
@niheaven Thanks for adding the judgment for null values, but there is something we need to revert.

If we use `(Get-Encoding($wc)).GetString($wc.DownloadData($url))`, then `(Get-Encoding($wc))` can only get the return result in UTF-8, because The `ResponseHeaders` of `$wc` will be `$null` before the data is downloaded. 

So I re-split the expression to handle the case where `charset` is something else.

niheaven:
Aha, right?

Okay, I'll test it tomorrow and it seems stable enough now.

niheaven (approved):

PR #4994, opened 2022-06-18
fix(decompress): Handle split RAR archives

issaclin32:
* This is a follow-up of #4915.

* **Splitted RAR archives** is rare for normal apps, but it is still possible for **games** or something with a large file size.

* File name for splitted RAR archives can be: something.**part1**.rar, something.**part01**.rar, something.**part001**.rar, ..., depending on total number of splits.

* For `TestCases.zip`:  I tried to compress `empty` into 3 RAR archives (30 Bytes each), but it didn't work. *WinRAR* rather creates a single file. The minimum size for RAR split archive is 925 Bytes according to my tests. (not sure why though)


lib/decompress.ps1
```diff
@@ -67,6 +67,9 @@ function Expand-7zipArchive {
         if (($Path -replace '.*\.([^\.]*)$', '$1') -eq '001') {
             # Remove splited 7-zip archive parts
             Get-ChildItem "$($Path -replace '\.[^\.]*$', '').???" | Remove-Item -Force
+        } elseif ($($Path -replace '.*\.part(\d+)\.rar$', '$1')[-1] -eq '1') {
```
rashil2000, line 70:
```suggestion
        } elseif (($Path -replace '.*\.part(\d+)\.rar$', '$1')[-1] -eq '1') {
```
I think this should work too

rashil2000 (approved):

niheaven (approved):

PR #5004, opened 2022-06-21
feat(chore): Add missing -a/--all param to all commands

rashil2000:
#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## [Unreleased](https://github.com/ScoopInstaller/Scoop/compare/master...develop)
 
+### Features
+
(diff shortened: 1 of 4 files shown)


niheaven:
LGTM for 'cache' and 'reset', but for 'virustotal', the `-e` sounds strange, and why not remove `arch` and let `-a` stands for `--all`?

Users check not-installed app's URLs for both 32bit and 64bit, right?

rashil2000:
> Users check not-installed app's URLs for both 32bit and 64bit, right?

I'm not really sure about this.

For consistency I can remove the `arch` flag, and use the system's default architecture only. What should we do?

niheaven:
> For consistency I can remove the `arch` flag, and use the system's default architecture only. What should we do?

I prefer to check both for not installed apps, and only check used architecture if installed.

niheaven (dismissed):
LGTM

PR #3481, opened 2019-05-28
Allow passing JSON file to checkver

r15ch13:
- Moved JSON parsing into the queue
  - Removes some loops
- Allow passing a JSON file path to `checkver.ps1`
  - Example: `> .\bin\checkver.ps1 -App ..\somewhere\bucket\test.json -Force`
- Use proper parameters values for `next()`
- Improve event cleanup

PR #5019, opened 2022-06-24
feat(checkver,auto-pr): Allow passing file path

### LINKED ISSUE #4606: [Feature] Allow passing array of app names to `checkver` utility
#### Is your feature request related to a problem? Please describe.

Original discussion here - https://github.com/ScoopInstaller/Main/pull/2871#issuecomment-987558132

When using authenticated requests, GitHub poses a limitation of 1000 API calls per hour. For buckets that have many GitHub-dependent manifests, this will be a problem.

#### Describe the solution you'd like
If the `checkver` utility is allowed to work on a set of app names, we can split the update process by passing each subset at a different hour so that we remain under the limit. 

#### Describe alternatives you've considered
- `checkver` can currently work on a single app, or an entire folder. Neither is feasible because passing individual app names at each workflow run would be too wasteful.
- We can't pause based on the `X-RateLimit-*` HTTP headers, because we will have pause for an hour every time, which is not very ideal. It's better to just have separate workflow runs each time.

rashil2000:
#### Description
Allow passing local file paths to checkver utility.

#### Motivation and Context

Closes #4606

Relates to #3481

image

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 - **scoop-status:** Check bucket status, improve output ([#5011](https://github.com/ScoopInstaller/Scoop/issues/5011))
 - **scoop-info:** Show app installed/download size ([#4886](https://github.com/ScoopInstaller/Scoop/issues/4886))
 - **scoop-import:** Import a Scoop installation from JSON ([#5014](https://github.com/ScoopInstaller/Scoop/issues/5014))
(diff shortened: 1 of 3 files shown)


niheaven:
Please use ParameterSet to choose between `Dir` and `File`, or rename them to `Path` and check if it is directory or file later. The latter approach is better, IMO.

niheaven (commented):
LGTM, just one remark, this should go to `Builds` in CHANGELOG

PR #5032, opened 2022-07-03
refactor(shim, scoop-cache) use -join instead of '+' to concatenate strings

zStruCat:
use "-join" to improve performance

#### Checklist:


rashil2000:
Do you have some benchmarks?

zStruCat:
No

niheaven:
Then this should be kept as is

rashil2000:
Hmm yes, unless there's a problem or bottleneck, we can skip on such optimizations for keeping readability

PR #5034, opened 2022-07-03
fix(scoop-import): Use `foreach` instead of `ForEach-Object` for nullity check

### LINKED ISSUE #5033: [Bug] Import Command Display Abnormality
When trying to import a exported json which doesn't have a config property:
图片
I Suppose there are two issues here:
1. Nullity check
2. Trying to add buckets that already exist (exist in both the import file and scoop)

zStruCat:
## Description
Check whether the wanted property of bucket is null. 
Check if the to-be-imported bucket is already added locally

## Context
Closes #5033


rashil2000:
I've made some tweaks:
- Using `foreach` loop in place of `ForEach-Object` automatically takes care of null values
- The `add_bucket` function has a robust check for duplicate bucket names and origin URIs, so we don't need to add another check. Moreover, if a bucket already exists, we _want_ to show an error while importing, so that users know that bucket is there.

niheaven (dismissed):
LGTM

zStruCat:
>     * The `add_bucket` function has a robust check for duplicate bucket names and origin URIs, so we don't need to add another check. Moreover, if a bucket already exists, we _want_ to show an error while importing, so that users know that bucket is there.

I agree that "if a bucket already exists, we _want_ to show an error while importing", but currently the error hint is misleading: 
"use "scoop bucket rm main" to remove it".
From my perspectivie, "'$($_.Name)' is already in your local bucket. It won't be added again." is much meaningful and user-friendly.

libexec/scoop-import.ps1
```diff
@@ -21,18 +24,25 @@ if (Test-Path $scoopfile) {
...
+    Write-Host "'$($item.Name)' has been set to '$($item.Value)'"
 }
 
-$import.buckets | ForEach-Object {
-    add_bucket $_.Name $_.Source | Out-Null
-    $bucket_names += $_.Name
+
+$local_buckets = Get-LocalBucket
+foreach($item in $import.buckets){
+    if($item.Name -in $local_buckets){
```
rashil2000, line 35: As I said, this check is not sufficient, it only checks the names. The proper method would be to revert the commit and remove the message part `Use 'scoop bucket rm $name' to remove it.` from the `add_bucket` function in lib/buckets.ps1

zStruCat: You're right. I will revert the commit.

zStruCat:
I've changed accordingly
图片

lib/buckets.ps1
```diff
@@ -120,7 +120,8 @@ function add_bucket($name, $repo) {
 
     $dir = Find-BucketDirectory $name -Root
     if (Test-Path $dir) {
-        warn "The '$name' bucket already exists. Use 'scoop bucket rm $name' to remove it."
+        warn "The '$name' bucket already exists locally. It won't be added again."
+        info "If you want to add the '$name' bucket. You must first use 'scoop bucket rm $name' to remove the already-locally-added namesake."
```
rashil2000, line 124:
```suggestion
        warn "The '$name' bucket already exists. To add this bucket again, first remove it by running 'scoop bucket rm $name'."
```

This is a more clear message

PR #5061, opened 2022-07-23
feat(core): Improve function `Invoke-ExternalCommand`

lewis-yeung:
### Description

This PR mainly made 3 changes to the function `Invoke-ExternalCommand`:

1. Use `ArgumentList` in PowerShell 6.1 and later (built on .NET Core 2.1+), otherwise escape arguments using [the general MSVC Runtime's rules](https://docs.microsoft.com/en-us/previous-versions/17w5ykft(v=vs.85)) in older versions.

	In the past, due to the arbitraryness of arguments, we are not able to handle all cases where arguments may contain spaces, quotes or other special characters for every automatic installer script that uses `Invoke-ExternalCommand`. With this change, we don't have to escape arguments manually for most external commands any more. But for the following exceptions who basically don't follow the general MSVC Runtime's rules of parsing command line, just simply concatenate arguments with spaces to build the argument list string, which means we have to escape arguments manually for them: (case insensitive)

	- Well-known executables: `cmd.exe`, `cscript.exe`, `wscript.exe` and `msiexec.exe`.
	- Files with these suffixes: `.bat`, `.cmd`, `.js`, `.vbs` and `.wsf`.

	Ref: [MS docs](https://docs.microsoft.com/en-us/powershell/scripting/learn/experimental-features?view=powershell-7.2#psnativecommandargumentpassing)

2. Support a new `-NoWindow` switch which indicates the process should be started without creating a new window to contain it. This switch cannot go with `-RunAs`.

3. Remove a deadlock potential. See [MS docs](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput?view=net-6.0#remarks) for details.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## [Unreleased](https://github.com/ScoopInstaller/Scoop/compare/master...develop)
 
+### Features
+- **core:** Improve function `Invoke-ExternalCommand` ([#5061](https://github.com/ScoopInstaller/Scoop/issues/5061))
+
 ### Bug Fixes
 
 - **install:** Move from cache when `--no-cache` is specified ([#5039](https://github.com/ScoopInstaller/Scoop/issues/5039))
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -488,6 +488,9 @@ function Invoke-ExternalCommand {
         [Alias("Args")]
         [String[]]
         $ArgumentList,
+        [Parameter(ParameterSetName = "Default")]
+        [Switch]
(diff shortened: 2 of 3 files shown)


niheaven:
I'm sorry, but could you please separate this PR into three so I could review them one by one?

lewis-yeung:
> I'm sorry, but could you please separate this PR into three so I could review them one by one?

OK. I'll do it later. Now I'm closing this first.

lewis-yeung:
@niheaven Please see #5064, #5065 and #5066. Thanks.

PR #5065, opened 2022-07-25
feat(core): Improve argument concatenation in `Invoke-ExternalCommand`

lewis-yeung:
### Description

Use `ArgumentList` in PowerShell 6.1 and later (built on .NET Core 2.1+), otherwise escape arguments using [the general MSVC Runtime's rules](https://docs.microsoft.com/en-us/previous-versions/17w5ykft(v=vs.85)) in older versions.

In the past, due to the arbitraryness of arguments, we are not able to handle all cases where arguments may contain spaces, quotes or other special characters for every automatic installer script that uses `Invoke-ExternalCommand`. With this change, we don't have to escape arguments manually for most external commands any more. But for the following exceptions who basically don't follow the general MSVC Runtime's rules of parsing command line, just simply concatenate arguments with spaces to build the argument list string, which means we have to escape arguments manually for them: (case insensitive)

- Well-known executables: `cmd.exe`, `cscript.exe`, `wscript.exe` and `msiexec.exe`.
- Files with these suffixes: `.bat`, `.cmd`, `.js`, `.vbs` and `.wsf`.

Ref: [MS docs](https://docs.microsoft.com/en-us/powershell/scripting/learn/experimental-features?view=powershell-7.2#psnativecommandargumentpassing)

### Checklist:

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -526,7 +526,7 @@ function Invoke-ExternalCommand {
         # ArgumentList is supported in PowerShell 6.1 and later (built on .NET Core 2.1+)
         # ref-1: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.argumentlist?view=net-6.0
         # ref-2: https://docs.microsoft.com/en-us/powershell/scripting/whats-new/differences-from-windows-powershell?view=powershell-7.2#net-framework-vs-net-core
-        $ArgumentList | ForEach-Object -Process { $Process.StartInfo.ArgumentList.Add($_) }
+        $ArgumentList | ForEach-Object { $Process.StartInfo.ArgumentList.Add($_) }
     } else {
         # escape arguments manually in lower versions, refer to https://docs.microsoft.com/en-us/previous-versions/17w5ykft(v=vs.85)
         $escapedArgs = $ArgumentList | ForEach-Object {
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -507,11 +507,10 @@ function Invoke-ExternalCommand {
     }
     $Process = New-Object System.Diagnostics.Process
     $Process.StartInfo.FileName = $FilePath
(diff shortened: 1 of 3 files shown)


lib/core.ps1
```diff
@@ -521,6 +520,27 @@ function Invoke-ExternalCommand {
         $Process.StartInfo.UseShellExecute = $true
         $Process.StartInfo.Verb = 'RunAs'
     }
+    if ($FilePath -match '^((cmd|cscript|wscript|msiexec)(\.exe)?|.*\.(bat|cmd|js|vbs|wsf))$') {
+        $Process.StartInfo.Arguments = $ArgumentList -join ' '
+    } elseif ($Process.StartInfo.ArgumentList.Add) {
+        # ArgumentList is supported in PowerShell 6.1 and later (built on .NET Core 2.1+)
+        # ref-1: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.argumentlist?view=net-6.0
+        # ref-2: https://docs.microsoft.com/en-us/powershell/scripting/whats-new/differences-from-windows-powershell?view=powershell-7.2#net-framework-vs-net-core
+        $ArgumentList | ForEach-Object -Process { $Process.StartInfo.ArgumentList.Add($_) }
```
niheaven, line 529:
```suggestion
        $ArgumentList | ForEach-Object { $Process.StartInfo.ArgumentList.Add($_) }
```

`Process` is the default.

niheaven (approved):
LGTM, but may need more tests in `develop`

niheaven:
And thank @issaclin32 for the fixes of incompatibility. 😭

chawyehsu:
Can we revert this? It's a broken implementation.

niheaven:
Revert it in `develop`, do a release, and then try to make it more robust. Is it okay? @ScoopInstaller/maintainers @lewis-yeung

lewis-yeung:
@niheaven Seems to make sense. But what about those manifests that have already been adjusted for this feature? Or if you already have an idea on how to make a bidirectionally compatible implementation?

PR #5066, opened 2022-07-25
feat(core): Create no window by default in `Invoke-ExternalCommand`

lewis-yeung:
### Description

Support a new `-NoWindow` switch which indicates the process should be started without creating a new window to contain it. **This switch cannot go with `-RunAs`.**

### Checklist:

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -520,6 +520,8 @@ function Invoke-ExternalCommand {
     if ($RunAs) {
         $Process.StartInfo.UseShellExecute = $true
         $Process.StartInfo.Verb = 'RunAs'
+    } else {
+        $Process.StartInfo.CreateNoWindow = $true
     }
     if ($FilePath -match '^((cmd|cscript|wscript|msiexec)(\.exe)?|.*\.(bat|cmd|js|vbs|wsf))$') {
         $Process.StartInfo.Arguments = $ArgumentList -join ' '
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Features
 
+- **core:** Create no window by default in `Invoke-ExternalCommand` ([#5066](https://github.com/ScoopInstaller/Scoop/issues/5066))
 - **core:** Improve argument concatenation in `Invoke-ExternalCommand` ([#5065](https://github.com/ScoopInstaller/Scoop/issues/5065))
 
 ### Bug Fixes


lib/core.ps1
```diff
@@ -507,6 +510,7 @@ function Invoke-ExternalCommand {
     }
     $Process = New-Object System.Diagnostics.Process
     $Process.StartInfo.FileName = $FilePath
+    $Process.StartInfo.CreateNoWindow = $NoWindow
```
niheaven, line 513: Move this line to L523/L527's `if`, and should this flag be set to true by default? Since normally we don't need a new window here.

lewis-yeung: > should this flag be set to true by default? Since normally we don't need a new window here.

No, it doesn't have to be. Mostly we use this function to invoke CLI programs, so `-NoWindow` is just for any other that has a GUI.

niheaven:
```pwsh
    if ($RunAs) {
        $Process.StartInfo.UseShellExecute = $true
        $Process.StartInfo.Verb = 'RunAs'
    } else {
        $Process.StartInfo.CreateNoWindow = $true
    }
```

Only if `Verb = 'RunAs'` `CreateNoWindow` will not be used, and is this a better solution?

lewis-yeung: Oh, yes. Now I got what you meant. ✅

niheaven: > No, it doesn't have to be. Mostly we use this function to invoke CLI programs, so -NoWindow is just for any other that has a GUI.

Now this one... Just in my code above, set `CreateNoWindow` to true will not add costs to CLI process and for GUI one, we always don't want a new window if run in `Invoke-ExternalCommand`, right?

lewis-yeung: > for GUI one, we always don't want a new window if run in `Invoke-ExternalCommand`, right?

Should we keep a way for possible user interaction during the installation? Although this may be rare. I definitely understand and agree that "no GUI" is part of Scoop's philosophy.

niheaven: No, we don't need to keep such a way, so do installation arguments, IMO.

For different installation types, using customized manifest should be better.

lewis-yeung: Fine. Then we should remove this switch and set `CreateNoWindow` to `$true`.

niheaven (approved):

chawyehsu:
This breaks my install/uninstall scripts for my [dorado/msys2-cn](https://github.com/chawyehsu/dorado/blob/master/bucket/msys2-cn.json#L13). With this forced `CreateNoWindow` I keep getting error code 3840 while triggering `msys2-cn` installation. 2 hours were wasted for me finding out this cause. Can we have a switch to stop the forced NoWindow or how could I fix this on my side?

PR #4798, opened 2022-03-10
fix(installed): If no `$global`, check both local and global installed

niheaven:
#### Description

Unintentional deletion in #4650 that breaks `installed <app>` (w/o `$global`) which check both locally and globally installed app.

#### Motivation and Context

Closes #4796

Relates to #4650

#### How Has This Been Tested?

Install `lua` as global app, before and after this fix:

image

#### Checklist:

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).
- [x] I have updated the documentation accordingly.
- [x] I have updated/passed the tests accordingly.

PR #5088, opened 2022-08-06
 fix(installed): Use 'System.Nullable<bool>' for param 'global'

niheaven:
#### Description

Use a native way to pass `$null` to `installed`, see [ref](https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/strongly-encouraged-development-guidelines?view=powershell-5.1#parameters-that-take-true-and-false)

#### Motivation and Context

- Relates to #4798

image

#### Checklist:


rashil2000 (approved):

candrapersada:
will there be support for installing, updating and removing packages as admin+global?

PR #5089, opened 2022-08-07
feat(scoop-(un)hold): Support `scoop (un)hold scoop`

yi-Xu-0100:
#### Description

Support `scoop (un)hold scoop`.
Allow to disable scoop itself updates for one day.

#### Motivation and Context

The `scoop` is a app in scoop, but it can not to be hold.
Now it works. 😁

#### Checklist:

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -998,7 +998,7 @@ function show_app($app, $bucket, $version) {
 
 function last_scoop_update() {
     # PowerShell 6 returns an DateTime Object
-    $last_update = (get_config lastupdate)
+    $last_update = get_config lastUpdate
 
     if ($null -ne $last_update -and $last_update.GetType() -eq [System.String]) {
         try {
@@ -1011,14 +1011,13 @@ function last_scoop_update() {
 }
 
 function is_scoop_outdated() {
-    $last_update = $(last_scoop_update)
-    $now = [System.DateTime]::Now
+    $last_update = last_scoop_update
     if($null -eq $last_update) {
-        set_config lastupdate $now.ToString('o')
+        set_config lastUpdate ([System.DateTime]::Now.ToString('o')) | Out-Null
         # enforce an update for the first time
         return $true
     }
-    return $last_update.AddHours(3) -lt $now.ToLocalTime()
+    return $last_update.AddHours(3) -lt [System.DateTime]::Now.ToLocalTime()
 }
 
 function substitute($entity, [Hashtable] $params, [Bool]$regexEscape = $false) {
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -997,12 +997,13 @@ function show_app($app, $bucket, $version) {
 }
 
 function is_scoop_outdated() {
+    $now = [System.DateTime]::Now
     try{
-        $expireHour = (New-TimeSpan $(get_config lastUpdate)).TotalHours
+        $expireHour = (New-TimeSpan (get_config lastUpdate) $now).TotalHours
         return ($expireHour -ge 3)
     } catch {
         # If not System.DateTime
-        set_config lastUpdate ([System.DateTime]::Now.ToString('o')) | Out-Null
+        set_config lastUpdate ($now.ToString('o')) | Out-Null
         return $true
     }
 }
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -998,7 +998,7 @@ function show_app($app, $bucket, $version) {
 
 function is_scoop_outdated() {
     $now = [System.DateTime]::Now
-    try{
+    try {
         $expireHour = (New-TimeSpan (get_config lastUpdate) $now).TotalHours
         return ($expireHour -ge 3)
     } catch {
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -996,28 +996,15 @@ function show_app($app, $bucket, $version) {
     return $app
 }
 
-function last_scoop_update() {
-    # PowerShell 6 returns an DateTime Object
-    $last_update = get_config lastUpdate
-
-    if ($null -ne $last_update -and $last_update.GetType() -eq [System.String]) {
-        try {
-            $last_update = [System.DateTime]::Parse($last_update)
-        } catch {
-            $last_update = $null
-        }
-    }
-    return $last_update
-}
-
 function is_scoop_outdated() {
-    $last_update = last_scoop_update
-    if($null -eq $last_update) {
+    try{
+        $expireHour = (New-TimeSpan $(get_config lastUpdate)).TotalHours
+        return ($expireHour -ge 3)
+    } catch {
+        # If not System.DateTime
         set_config lastUpdate ([System.DateTime]::Now.ToString('o')) | Out-Null
-        # enforce an update for the first time
         return $true
     }
-    return $last_update.AddHours(3) -lt [System.DateTime]::Now.ToLocalTime()
 }
 
 function substitute($entity, [Hashtable] $params, [Bool]$regexEscape = $false) {
--- a/libexec/scoop-config.ps1
+++ b/libexec/scoop-config.ps1
@@ -39,8 +39,8 @@
 #       Could be used for testing specific functionalities before released into all users.
 #       If you want to receive updates earlier to test new functionalities use develop (see: 'https://github.com/ScoopInstaller/Scoop/issues/2939')
 #
-# SCOOP_HOLD_DAYS: 1
-#       Allow to disable Scoop itself updates for 'SCOOP_HOLD_DAYS' days.
+# SCOOP_HOLD: $true|$false|<n>
+#       Allow to disable Scoop itself updates, '$true' for 1 day, '$false' for reenable, '<n>' for n days.
 #       This configuration havs the same effect with 'scoop (un)hold scoop'.
 #
 # proxy: [username:password@]host:port
--- a/libexec/scoop-config.ps1
+++ b/libexec/scoop-config.ps1
@@ -40,8 +40,8 @@
 #       If you want to receive updates earlier to test new functionalities use develop (see: 'https://github.com/ScoopInstaller/Scoop/issues/2939')
 #
 # SCOOP_HOLD: $true|$false|<n>
-#       Allow to disable Scoop itself updates, '$true' for 1 day, '$false' for reenable, '<n>' for n days.
-#       This configuration havs the same effect with 'scoop (un)hold scoop'.
+#       Disables Scoop self-updates: '$true' for 1 day, '$false' for reenable, '<n>' for n days.
+#       This configuration has the same effect as 'scoop (un)hold scoop'.
 #
 # proxy: [username:password@]host:port
 #       By default, Scoop will use the proxy settings from Internet Options, but with anonymous authentication.
--- a/libexec/scoop-config.ps1
+++ b/libexec/scoop-config.ps1
@@ -39,9 +39,9 @@
 #       Could be used for testing specific functionalities before released into all users.
 #       If you want to receive updates earlier to test new functionalities use develop (see: 'https://github.com/ScoopInstaller/Scoop/issues/2939')
 #
-# SCOOP_HOLD: $true|$false
-#       Allow to disable Scoop itself updates.
-#       This configuration have the same function with 'scoop (un)hold scoop'.
+# SCOOP_HOLD_DAYS: 30
+#       Allow to disable Scoop itself updates for 'SCOOP_HOLD_DAYS' days.
+#       This configuration havs the same effect with 'scoop (un)hold scoop'.
 #
 # proxy: [username:password@]host:port
 #       By default, Scoop will use the proxy settings from Internet Options, but with anonymous authentication.
--- a/libexec/scoop-config.ps1
+++ b/libexec/scoop-config.ps1
@@ -39,6 +39,10 @@
 #       Could be used for testing specific functionalities before released into all users.
 #       If you want to receive updates earlier to test new functionalities use develop (see: 'https://github.com/ScoopInstaller/Scoop/issues/2939')
 #
+# SCOOP_HOLD: $true|$false
+#       Allow to disable Scoop itself updates.
+#       This configuration have the same function with 'scoop (un)hold scoop'.
+#
 # proxy: [username:password@]host:port
 #       By default, Scoop will use the proxy settings from Internet Options, but with anonymous authentication.
 #
--- a/libexec/scoop-config.ps1
+++ b/libexec/scoop-config.ps1
@@ -39,10 +39,6 @@
 #       Could be used for testing specific functionalities before released into all users.
 #       If you want to receive updates earlier to test new functionalities use develop (see: 'https://github.com/ScoopInstaller/Scoop/issues/2939')
 #
-# SCOOP_HOLD: $true|$false|<n>
-#       Disables Scoop self-updates: '$true' for 1 day, '$false' for reenable, '<n>' for n days.
-#       This configuration has the same effect as 'scoop (un)hold scoop'.
-#
 # proxy: [username:password@]host:port
 #       By default, Scoop will use the proxy settings from Internet Options, but with anonymous authentication.
 #
--- a/libexec/scoop-config.ps1
+++ b/libexec/scoop-config.ps1
@@ -39,7 +39,7 @@
 #       Could be used for testing specific functionalities before released into all users.
 #       If you want to receive updates earlier to test new functionalities use develop (see: 'https://github.com/ScoopInstaller/Scoop/issues/2939')
 #
-# SCOOP_HOLD_DAYS: 30
+# SCOOP_HOLD_DAYS: 1
 #       Allow to disable Scoop itself updates for 'SCOOP_HOLD_DAYS' days.
 #       This configuration havs the same effect with 'scoop (un)hold scoop'.
 #
--- a/libexec/scoop-hold.ps1
+++ b/libexec/scoop-hold.ps1
@@ -33,8 +33,9 @@ $apps | ForEach-Object {
     $app = $_
 
     if ($app -eq 'scoop') {
-        set_config 'SCOOP_HOLD' $true | Out-Null
-        success "$app is now held and might not be updated for one day."
+        $update_until = [System.DateTime]::Now.AddDays(1)
+        set_config 'update_until' $update_until.ToString('o') | Out-Null
+        success "$app is now held and might not be updated until $($update_until.ToLocalTime())."
         return
     }
     if (!(installed $app $global)) {
--- a/libexec/scoop-hold.ps1
+++ b/libexec/scoop-hold.ps1
@@ -32,26 +32,30 @@ if ($global -and !(is_admin)) {
 $apps | ForEach-Object {
     $app = $_
 
-    if (!(installed $app $global)) {
-        if ($global) {
-            error "'$app' is not installed globally."
-        } else {
-            error "'$app' is not installed."
-        }
-        return
-    }
-
-    if (get_config NO_JUNCTIONS) {
-        $version = Select-CurrentVersion -App $app -Global:$global
+    if ($app -eq 'scoop') {
+        set_config 'SCOOP_HOLD' $true | Out-Null
     } else {
-        $version = 'current'
+        if (!(installed $app $global)) {
+            if ($global) {
+                error "'$app' is not installed globally."
+            } else {
+                error "'$app' is not installed."
+            }
+            return
+        }
+
+        if (get_config NO_JUNCTIONS) {
+            $version = Select-CurrentVersion -App $app -Global:$global
+        } else {
+            $version = 'current'
+        }
+        $dir = versiondir $app $version $global
+        $json = install_info $app $version $global
+        $install = @{}
+        $json | Get-Member -MemberType Properties | ForEach-Object { $install.Add($_.Name, $json.($_.Name)) }
+        $install.hold = $true
+        save_install_info $install $dir
     }
-    $dir = versiondir $app $version $global
-    $json = install_info $app $version $global
-    $install = @{}
-    $json | Get-Member -MemberType Properties | ForEach-Object { $install.Add($_.Name, $json.($_.Name)) }
-    $install.hold = $true
-    save_install_info $install $dir
     success "$app is now held and can not be updated anymore."
 }
 
--- a/libexec/scoop-hold.ps1
+++ b/libexec/scoop-hold.ps1
@@ -33,8 +33,8 @@ $apps | ForEach-Object {
     $app = $_
 
     if ($app -eq 'scoop') {
-        set_config 'SCOOP_HOLD_DAYS' 1 | Out-Null
-        success "$app is now held and can not be updated for one day."
+        set_config 'SCOOP_HOLD' 1 | Out-Null
+        success "$app is now held and might not be updated for one day."
         return
     }
     if (!(installed $app $global)) {
--- a/libexec/scoop-hold.ps1
+++ b/libexec/scoop-hold.ps1
@@ -33,7 +33,7 @@ $apps | ForEach-Object {
     $app = $_
 
     if ($app -eq 'scoop') {
-        set_config 'SCOOP_HOLD' 1 | Out-Null
+        set_config 'SCOOP_HOLD' $true | Out-Null
         success "$app is now held and might not be updated for one day."
         return
     }
--- a/libexec/scoop-hold.ps1
+++ b/libexec/scoop-hold.ps1
@@ -34,28 +34,29 @@ $apps | ForEach-Object {
 
     if ($app -eq 'scoop') {
         set_config 'SCOOP_HOLD' $true | Out-Null
-    } else {
-        if (!(installed $app $global)) {
-            if ($global) {
-                error "'$app' is not installed globally."
-            } else {
-                error "'$app' is not installed."
-            }
-            return
-        }
-
-        if (get_config NO_JUNCTIONS) {
-            $version = Select-CurrentVersion -App $app -Global:$global
-        } else {
-            $version = 'current'
-        }
-        $dir = versiondir $app $version $global
-        $json = install_info $app $version $global
-        $install = @{}
-        $json | Get-Member -MemberType Properties | ForEach-Object { $install.Add($_.Name, $json.($_.Name)) }
-        $install.hold = $true
-        save_install_info $install $dir
+        success "$app is now held and can not be updated anymore."
+        return
     }
+    if (!(installed $app $global)) {
+        if ($global) {
+            error "'$app' is not installed globally."
+        } else {
+            error "'$app' is not installed."
+        }
+        return
+    }
+
+    if (get_config NO_JUNCTIONS) {
+        $version = Select-CurrentVersion -App $app -Global:$global
+    } else {
+        $version = 'current'
+    }
+    $dir = versiondir $app $version $global
+    $json = install_info $app $version $global
+    $install = @{}
+    $json | Get-Member -MemberType Properties | ForEach-Object { $install.Add($_.Name, $json.($_.Name)) }
+    $install.hold = $true
+    save_install_info $install $dir
     success "$app is now held and can not be updated anymore."
 }
 
--- a/libexec/scoop-hold.ps1
+++ b/libexec/scoop-hold.ps1
@@ -33,8 +33,8 @@ $apps | ForEach-Object {
     $app = $_
 
     if ($app -eq 'scoop') {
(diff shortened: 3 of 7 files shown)


niheaven:
I don't think scoop itself should be hold (not updated) since we keep fixing bugs in it, and this shouldn't be fixed IMO.

yi-Xu-0100:
@niheaven It is only an option both for user and contributor. Scoop always updates with remote without asking, and it may lose the local changes.🤔

For me, I change the content of scoop for doing some test locally, but the scoop always updates as the remote and loss my change content, I think I can hold the scoop to test the local feature, and enable update quickly if new version published. 😥

niheaven:
Yes, I've the same trouble as yours, and I think a better way is auto stashing the uncommited changes and noticing users that they have these changes and scoop has stashed them.

Would you like to make this feature and submit PR about that?

yi-Xu-0100:
https://github.com/ScoopInstaller/Scoop/blob/66296dcad228198d1d8628007dad7c63b5850574/libexec/scoop-update.ps1#L62-L66

I use a warning notification when scoop used to update itself.  It effects every time when users want to update scoop. 😁

The stash operation may helpful, but I may not deal with it good. If I make this feature, I will submit a PR to here. 🙌

But I also think the hold should be a choice for uses of scoop. The scoop is an app in `scoop/apps`, so I think it needs to obey the rule of an app. An app can be hold to disable update, not likes Windows. 😂

r15ch13:
Wouldn't it be better to move the bucket update code out of `update_scoop()` so the there is no need for a big overarching if-statement?

niheaven:
Just add autostash in #5091, and this could meet your need @yi-Xu-0100

niheaven:
> Wouldn't it be better to move the bucket update code out of `update_scoop()` so there is no need for a big overarching if-statement? The function can then be returned early.

Good idea for the function separation and may be done later.

yi-Xu-0100:
@niheaven

#5091 is good for users who change a little and do things in the current branch. 👍

But for contributors, It always set the scoop to remote, we still need to check out our own branch and stash pop changes. It will make many steps for test. 🤔

niheaven:
No, you don't. It will stash uncommitted change on your working branch and then checkout 'develop' branch (the one in config file), all your works will not be lost.

yi-Xu-0100:
@niheaven I know that the #5091 will stash uncommitted change and will not lose my work. That is a good job! 👍

But we do not always work on the branch in which we config `SCOOP_REPO` and `SCOOP_BRANCH` in the config file.
New PR may create a new branch, and I do not want to set the config every time. 😥

It will update when I test in another branch, and I need to check out and get my work back every time. It can be simply used `scoop hold scoop` to avoid this happening. 🤔

niheaven:
Hmm, you may not understand my point. Which I mean is, you create a new branch and do some change, commit them, then do another changes, then you want to test it but a `scoop update` happened, the #5091 will stash all your uncommitted changes on the new branch and then the following code of `scoop update` will checkout `develop`.

All you do after such update are checkout your working branch and `git stash pop`, everything is back: your new branch, your new commit and your uncommitted changes.

r15ch13:
Or develop stuff in a separate repo and use it from the editor and create a custom shim that points to the develop repo. :smile:

yi-Xu-0100:
@niheaven @r15ch13 

I know what all you say, I just think it needs to do many steps to test little things about update.

The another reason is that I think the user can choose to hold the scoop if the user did not have a fork repo. It could be an option, right?

niheaven:
> The another reason is that I think the user can choose to hold the scoop if the user did not have a fork repo.
> It could be an option, right?

That's what I'm against, that one should never hold scoop core since there're ongoing bugfixes and one may omit them if scoop is held. It's something dangerous even if there may be enough notices or warnings.

yi-Xu-0100:
@niheaven All right, if you resolutely retain this opinion. Could we add a configuration of `SCOOP_HOLD_DAYS` instead of it? 

Just like windows, it could hold update operation for days. It could be a choice for the minority.

niheaven:
> @niheaven All right, if you resolutely retain this opinion. Could we add a configuration of `SCOOP_HOLD_DAYS` instead of it?
> 
> Just like windows, it could hold update operation for days. It could be a choice for the minority.

That's great, and please make it!

yi-Xu-0100:
@niheaven The new configuration of `SCOOP_HOLD_DAYS` has been done. 😁

libexec/scoop-config.ps1
```diff
@@ -39,6 +39,10 @@
 #       Could be used for testing specific functionalities before released into all users.
 #       If you want to receive updates earlier to test new functionalities use develop (see: 'https://github.com/ScoopInstaller/Scoop/issues/2939')
 #
+# SCOOP_HOLD_DAYS: 30
```
niheaven, line 42: Default should be 1 day

libexec/scoop-hold.ps1
```diff
@@ -32,6 +32,11 @@ if ($global -and !(is_admin)) {
 $apps | ForEach-Object {
     $app = $_
 
+    if ($app -eq 'scoop') {
+        set_config 'SCOOP_HOLD_DAYS' 30 | Out-Null
```
niheaven, line 36: See above

libexec/scoop-update.ps1
```diff
@@ -136,6 +144,14 @@ function update_scoop() {
 
     shim "$currentdir\bin\scoop.ps1" $false
 
+    set_config lastupdate ([System.DateTime]::Now.ToString('o')) | Out-Null
```
niheaven, line 147: Run this `set_config` after scoop itself and buckets are all updated, just as before.

libexec/scoop-config.ps1
```diff
@@ -39,6 +39,10 @@
 #       Could be used for testing specific functionalities before released into all users.
 #       If you want to receive updates earlier to test new functionalities use develop (see: 'https://github.com/ScoopInstaller/Scoop/issues/2939')
 #
+# SCOOP_HOLD_DAYS: 1
+#       Allow to disable Scoop itself updates for 'SCOOP_HOLD_DAYS' days.
```
niheaven, line 43:
```suggestion
# SCOOP_HOLD: $true|$false|<n>
#       Allow to disable Scoop itself updates, '$true' for 1 day, '$false' for reenable, '<n>' for n days.
```

libexec/scoop-hold.ps1
```diff
@@ -32,6 +32,11 @@ if ($global -and !(is_admin)) {
 $apps | ForEach-Object {
     $app = $_
 
+    if ($app -eq 'scoop') {
+        set_config 'SCOOP_HOLD_DAYS' 1 | Out-Null
+        success "$app is now held and can not be updated for one day."
```
niheaven, line 37:
```suggestion
        set_config 'SCOOP_HOLD' 1 | Out-Null
        success "$app is now held and might not be updated for one day."
```

libexec/scoop-unhold.ps1
```diff
@@ -32,6 +32,11 @@ if ($global -and !(is_admin)) {
 $apps | ForEach-Object {
     $app = $_
 
+    if ($app -eq 'scoop') {
+        set_config 'SCOOP_HOLD_DAYS' $null | Out-Null
```
niheaven, line 36:
```suggestion
        set_config 'SCOOP_HOLD' $null | Out-Null
```

libexec/scoop-update.ps1
```diff
@@ -57,7 +57,15 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
...
-    if(!(Test-CommandAvailable git)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
+    if (!(Test-CommandAvailable git)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
+
+    $SCOOP_HOLD_DAYS = get_config 'SCOOP_HOLD_DAYS' 0
+    if (((New-TimeSpan (scoop config lastUpdate)).Days) -lt $SCOOP_HOLD_DAYS) {
+        warn "'SCOOP_HOLD_DAYS' has been setting to '$SCOOP_HOLD_DAYS' and skip updating Scoop..."
+        warn "If you want to update Scoop itself, use 'scoop unhold scoop' to enable it."
+        warn "If you want to change 'SCOOP_HOLD_DAYS', use 'scoop config SCOOP_HOLD_DAYS <days>' to set value."
+        return
+    }
```
niheaven, line 68:
```suggestion
    $SCOOP_HOLD = get_config SCOOP_HOLD 0
    if (((New-TimeSpan (get_config lastUpdate)).Days) -lt $SCOOP_HOLD) {
        warn "'SCOOP_HOLD' has been setting to '$SCOOP_HOLD' and skip updating Scoop for $([int]$SCOOP_HOLD) day(s)..."
        warn "If you want to update Scoop itself immediately, use 'scoop unhold scoop; scoop update'."
        return
    } else {
        set_config SCOOP_HOLD $null | Out-Null
    }
```

niheaven (commented):
Change the option name to `SCOOP_HOLD` and remove the option after holding days.

niheaven (approved):

niheaven:
Wow, you forget the CHANGELOG.

yi-Xu-0100:
@niheaven Some new change in https://github.com/ScoopInstaller/Scoop/pull/5089/commits/7f63df2cf7297017e723cc0f37611cf18f3a1fd4

I will update changelog soon

libexec/scoop-update.ps1
```diff
@@ -57,12 +57,21 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
 
 function update_scoop() {
     # check for git
-    if(!(Test-CommandAvailable git)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
+    if (!(Test-CommandAvailable git)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
 
-    Write-Host "Updating Scoop..."
+    $SCOOP_HOLD = get_config SCOOP_HOLD $false
     $last_update = $(last_scoop_update)
```
niheaven, line 63:
```suggestion
    $last_update = last_scoop_update
```

This should work, I don't know why use `$()`

yi-Xu-0100: @niheaven It is old code in files, may no one know why. 😂

niheaven (dismissed):

niheaven:
Wait to see if other comments.

libexec/scoop-update.ps1
yi-Xu-0100: @niheaven I find a problem about it. I move the `set_config`, but it will set `lastupdate` every time if update bucket well.

When I update next time and have configuration about `SCOOP_HOLD`, it will refresh the `lastupdate` if I did `scoop update` in $`SCOOP_HOLD` days. It may cause disable update for long time more than $`SCOOP_HOLD` days.

yi-Xu-0100:
@niheaven I find a problem about it. I move the `set_config`, but it will set `lastupdate` every time if update bucket well.

When I update next time and have configuration about `SCOOP_HOLD`, it will refresh the `lastupdate` if I did `scoop update` in $`SCOOP_HOLD` days. It may cause disable update for long time more than $`SCOOP_HOLD` days.

libexec/scoop-config.ps1
```diff
@@ -39,6 +39,10 @@
 #       Could be used for testing specific functionalities before released into all users.
 #       If you want to receive updates earlier to test new functionalities use develop (see: 'https://github.com/ScoopInstaller/Scoop/issues/2939')
 #
+# SCOOP_HOLD: $true|$false|<n>
+#       Allow to disable Scoop itself updates, '$true' for 1 day, '$false' for reenable, '<n>' for n days.
+#       This configuration havs the same effect with 'scoop (un)hold scoop'.
```
rashil2000, line 44:
```suggestion
#       Disables Scoop self-updates: '$true' for 1 day, '$false' for reenable, '<n>' for n days.
#       This configuration has the same effect as 'scoop (un)hold scoop'.
```

libexec/scoop-update.ps1
```diff
@@ -57,12 +57,21 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
...
+    if (!(Test-CommandAvailable git)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
 
-    Write-Host "Updating Scoop..."
-    $last_update = $(last_scoop_update)
+    $SCOOP_HOLD = get_config SCOOP_HOLD $false
+    $last_update = last_scoop_update
     if ($null -eq $last_update) {$last_update = [System.DateTime]::Now}
     $last_update = $last_update.ToString('s')
+    if (((New-TimeSpan ($last_update)).Days) -lt $SCOOP_HOLD) {
+        warn "'SCOOP_HOLD' has been setting to '$SCOOP_HOLD' and skip updating Scoop for $([int]$SCOOP_HOLD) day$(if($SCOOP_HOLD -gt 1){'s'})..."
```
rashil2000, line 67:
```suggestion
        warn "Skipping self-update as 'SCOOP_HOLD' has been set to $([int]$SCOOP_HOLD) day$(if($SCOOP_HOLD -gt 1){'s'})..."
```

rashil2000 (changes requested):
Grammar fixes

libexec/scoop-update.ps1
niheaven: Hmm, yes, so we need another config that saves the ending time. Lots of works...

niheaven:
> @niheaven I find a problem about it. I move the `set_config`, but it will set `lastupdate` every time if update bucket well.
> 
> When I update next time and have configuration about `SCOOP_HOLD`, it will refresh the `lastupdate` if I did `scoop update` in $`SCOOP_HOLD` days. It may cause disable update for long time more than $`SCOOP_HOLD` days.

Found a solution, but this may hold no more than 1 day. I'll make review soon.

niheaven:
image

libexec/scoop-config.ps1
```diff
@@ -39,6 +39,10 @@
 #       Could be used for testing specific functionalities before released into all users.
 #       If you want to receive updates earlier to test new functionalities use develop (see: 'https://github.com/ScoopInstaller/Scoop/issues/2939')
 #
+# SCOOP_HOLD: $true|$false|<n>
+#       Disables Scoop self-updates: '$true' for 1 day, '$false' for reenable, '<n>' for n days.
+#       This configuration has the same effect as 'scoop (un)hold scoop'.
+#
 # proxy: [username:password@]host:port
```
niheaven, line 46:
```suggestion
# proxy: [username:password@]host:port
```

libexec/scoop-hold.ps1
```diff
@@ -32,6 +32,11 @@ if ($global -and !(is_admin)) {
 $apps | ForEach-Object {
     $app = $_
 
+    if ($app -eq 'scoop') {
+        set_config 'SCOOP_HOLD' $true | Out-Null
+        success "$app is now held and might not be updated for one day."
```
niheaven, line 37:
```suggestion
        $update_until = [System.DateTime]::Now.AddDays(1)
        set_config 'update_until' ($update_until.ToString('o')) | Out-Null
        success "$app is now held and might not be updated until $($update_until.ToLocalTime())."
```

libexec/scoop-unhold.ps1
```diff
@@ -32,6 +32,11 @@ if ($global -and !(is_admin)) {
 $apps | ForEach-Object {
     $app = $_
 
+    if ($app -eq 'scoop') {
+        set_config 'SCOOP_HOLD' $null | Out-Null
```
niheaven, line 36:
```suggestion
        set_config 'update_until' $null | Out-Null
```

libexec/scoop-update.ps1
```diff
@@ -57,12 +57,23 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
...
+        $last_update = [System.DateTime]::Now.ToString('o')
+        set_config lastUpdate $last_update | Out-Null
+    }
+    if (((New-TimeSpan ($last_update)).Days) -lt $SCOOP_HOLD) {
+        warn "Skipping self-update as 'SCOOP_HOLD' has been set to $([int]$SCOOP_HOLD) day$(if($SCOOP_HOLD -gt 1){'s'})..."
+        warn "If you want to update Scoop itself immediately, use 'scoop unhold scoop; scoop update'."
+        return
+    } else {
+        set_config SCOOP_HOLD $null | Out-Null
+    }
```
niheaven, line 74:
```suggestion
    $last_update = last_scoop_update
    $now = [System.DateTime]::Now
    if ($null -eq $last_update) {
        $last_update = $now.ToString('o')
    }
    $update_until = get_config update_until
    if ($null -ne $update_until -and $update_until.GetType() -eq [System.String]) {
        try {
            $update_until = [System.DateTime]::Parse($update_until)
        } catch {
            $update_until = $null
        }
    }
    if ($null -ne $update_until -and $now -lt $update_until) {
        warn "Skipping self-update until $($update_until.ToLocalTime())..."
        warn "If you want to update Scoop itself immediately, use 'scoop unhold scoop; scoop update'."
        return
    } else {
        set_config update_until $null | Out-Null
    }
```

niheaven (commented):
image

libexec/scoop-update.ps1
```diff
@@ -317,6 +333,9 @@ if (-not ($apps -or $all)) {
 
     if ($updateScoop) {
         update_scoop
+        update_bucket
+        set_config lastUpdate ([System.DateTime]::Now.ToString('o')) | Out-Null
```
r15ch13, line 337: Remove this `set_config` call otherwise, it will always overwrite.

libexec/scoop-update.ps1
```diff
@@ -305,6 +318,9 @@ if (-not ($apps -or $all)) {
         exit 1
     }
     update_scoop
+    update_bucket
+    set_config lastUpdate ([System.DateTime]::Now.ToString('o')) | Out-Null
```
r15ch13, line 322: Remove this `set_config` call otherwise, it will always overwrite.

libexec/scoop-update.ps1
```diff
@@ -57,12 +57,23 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
...
+    if (!(Test-CommandAvailable git)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
+
+    $SCOOP_HOLD = get_config SCOOP_HOLD $false
+    $last_update = last_scoop_update
+    if ($null -eq $last_update) {
+        $last_update = [System.DateTime]::Now.ToString('o')
+        set_config lastUpdate $last_update | Out-Null
+    }
+    if (((New-TimeSpan ($last_update)).Days) -lt $SCOOP_HOLD) {
+        warn "Skipping self-update as 'SCOOP_HOLD' has been set to $([int]$SCOOP_HOLD) day$(if($SCOOP_HOLD -gt 1){'s'})..."
```
r15ch13, line 69:
```suggestion
        warn "Skipping self-update as 'SCOOP_HOLD' has been set to $([int]$SCOOP_HOLD) day$(if($SCOOP_HOLD -gt 1){'s'}) (until $($last_update.AddDays($SCOOP_HOLD)))"
```

libexec/scoop-update.ps1
r15ch13: There is no need for a new config value. Just letting the `lastUpdate` stay the same and adding `SCOOP_HOLD` to it returns the unhold datetime. :smile:

yi-Xu-0100:
For the latest change, I mainly refactored the code as @niheaven suggested. By handling many of the judgments in `try-catch` codes, the code became more streamlined. And `New-TimeSpan` makes it easier to compare times.

@niheaven @r15ch13 @rashil2000 Hope for your advice! 😊

r15ch13:
The name `update_until` doesn't reflect what it actually does. 🤔

Also missing the description in `scoop-config.ps1`

yi-Xu-0100:
@r15ch13 

I think @niheaven wants to use this variable as a built-in variable, similar to `lastUpdate` that also does not appear in the config description, and the general user should only control it through `scoop (un)hold scoop`.

niheaven:
> I think @niheaven wants to use this variable as a built-in variable, similar to `lastUpdate` that also does not appear in the config description, and the general user should only control it through `scoop (un)hold scoop`.

Yes, it is, this should be controlled by `scoop hold scoop` only, or we met the case that any later update will extend the holding period.

lib/core.ps1
```diff
@@ -996,29 +996,15 @@ function show_app($app, $bucket, $version) {
...
-    $now = [System.DateTime]::Now
-    if($null -eq $last_update) {
-        set_config lastupdate $now.ToString('o')
-        # enforce an update for the first time
+    try{
+        $expireHour = (New-TimeSpan $(get_config lastUpdate)).TotalHours
+        return ($expireHour -ge 3)
+    } catch {
+        # If not System.DateTime
+        set_config lastUpdate ([System.DateTime]::Now.ToString('o')) | Out-Null
```
niheaven, line 1005:
```suggestion
    $now = [System.DateTime]::Now
    try{
        $expireHour = (New-TimeSpan (get_config lastUpdate) $now).TotalHours
        return ($expireHour -ge 3)
    } catch {
        # If not System.DateTime
        set_config lastUpdate ($now.ToString('o')) | Out-Null
```

`New-TimeSpan` implied `[System.DateTime]::Now`, and make it explicit will be better.

libexec/scoop-update.ps1
```diff
@@ -57,12 +57,22 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
```
niheaven, line 57:
```suggestion
$show_update_log = get_config 'show_update_log' $true

```

`$show_update_log` is used in both `update_scoop()` and `update_bucket()`

libexec/scoop-update.ps1
```diff
@@ -57,12 +57,22 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
 
 function update_scoop() {
```
niheaven, line 58:
```suggestion
function update_scoop($show_update_log) {
```

libexec/scoop-update.ps1
```diff
@@ -57,12 +57,22 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
...
+        return
+    } else {
+        set_config update_until $null | Out-Null
+    }
 
     Write-Host "Updating Scoop..."
-    $last_update = $(last_scoop_update)
-    if ($null -eq $last_update) {$last_update = [System.DateTime]::Now}
-    $last_update = $last_update.ToString('s')
     $show_update_log = get_config 'show_update_log' $true
```
niheaven, line 76:
```suggestion
```

libexec/scoop-update.ps1
```diff
@@ -135,6 +145,11 @@ function update_scoop() {
     # }
 
     shim "$currentdir\bin\scoop.ps1" $false
+}
+
+function update_bucket() {
```
niheaven, line 150:
```suggestion
function update_bucket($show_update_log) {
```

libexec/scoop-update.ps1
```diff
@@ -305,6 +317,9 @@ if (-not ($apps -or $all)) {
         exit 1
     }
     update_scoop
+    update_bucket
```
niheaven, line 320:
```suggestion
    update_scoop $show_update_log
    update_bucket $show_update_log
```

libexec/scoop-update.ps1
```diff
@@ -317,6 +332,9 @@ if (-not ($apps -or $all)) {
 
     if ($updateScoop) {
         update_scoop
+        update_bucket
```
niheaven, line 335:
```suggestion
        update_scoop $show_update_log
        update_bucket $show_update_log
```

libexec/scoop-update.ps1
```diff
@@ -57,12 +57,22 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
 
 function update_scoop() {
     # check for git
-    if(!(Test-CommandAvailable git)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
+    if (!(Test-CommandAvailable git)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
+
+    try {
+        $update_until = ([System.DateTime]::Parse((get_config update_until)))
```
niheaven, line 63:
```suggestion
        $update_until = [System.DateTime]::Parse((get_config update_until))
```

libexec/scoop-update.ps1
```diff
@@ -57,12 +57,22 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
...
+    } catch {
+        $update_until = [System.DateTime]::now
```
niheaven, line 65:
```suggestion
        $update_until = $null
```

libexec/scoop-update.ps1
```diff
@@ -57,12 +57,22 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
...
+    }
+    if ((New-TimeSpan $update_until).TotalSeconds -lt 0) {
+        warn "Skipping self-update until $($update_until.ToLocalTime())..."
+        warn "If you want to update Scoop itself immediately, use 'scoop unhold scoop; scoop update'."
+        return
+    } else {
+        set_config update_until $null | Out-Null
+    }
```
niheaven, line 73:
```suggestion
    if ($null -ne $update_until) {
        if ((New-TimeSpan $update_until).TotalSeconds -lt 0) {
            warn "Skipping self-update until $($update_until.ToLocalTime())..."
            warn "If you want to update Scoop itself immediately, use 'scoop unhold scoop; scoop update'."
            return
        } else {
            set_config update_until $null | Out-Null
        }
    }
```

niheaven (changes requested):

libexec/scoop-update.ps1
```diff
@@ -54,16 +54,28 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
...
-function update_scoop() {
+function update_scoop($show_update_log) {
     # check for git
-    if(!(Test-CommandAvailable git)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
+    if (!(Test-CommandAvailable git)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
+
+    try {
+        $now = [System.DateTime]::Now.ToString('o')
+        $update_until = [System.DateTime]::Parse((get_config update_until $now))
+        if ((New-TimeSpan $update_until $now).TotalSeconds -lt 0) {
```
niheaven, line 66:
```suggestion
        $update_until = [System.DateTime]::Parse((get_config update_until))
        if ((New-TimeSpan $update_until).TotalSeconds -lt 0) {
```

Why use `$now` as default? `get_config update_until` is okay since there's `try-catch`

yi-Xu-0100: `[System.DateTime]::Parse` can not parser `$null`. If the user do not use `scoop hold scoop`, the `update_until` will not to be set and `get_config update_until` will return `$null`. That will be caught but should be silent.

niheaven: Hmm, got it, so I preferred using `if-else` in such case to make it clear.

niheaven: IF `update_until` is set, check it, IF NOT, go ahead.

yi-Xu-0100:
@niheaven @r15ch13 @rashil2000 I think this feature is ready to be merged, are there other comments?

niheaven:
I'm fine expect for the `try-catch` (`if-else` IMO).

lib/core.ps1
```diff
@@ -996,29 +996,16 @@ function show_app($app, $bucket, $version) {
...
-    return $last_update
-}
-
 function is_scoop_outdated() {
-    $last_update = $(last_scoop_update)
     $now = [System.DateTime]::Now
-    if($null -eq $last_update) {
-        set_config lastupdate $now.ToString('o')
-        # enforce an update for the first time
+    try{
```
rashil2000, line 1001:
```suggestion
    try {
```

rashil2000 (commented):
Looks fine to me

rashil2000 (approved):

niheaven (approved):

r15ch13:
The `update_until` value still does exactly the opposite of it's name.

niheaven:
So `update_after`?

r15ch13:
I would say `hold_update_until`. It also needs to be added to description in `scoop-config.ps1`

niheaven:
But this could not be set via `scoop config`, just likes `lastupdate`, since it is POSIX and is not human writable, in most cases.

PR #5090, opened 2022-08-07
perf(scoop-config): Update feedback for `scoop config <xxx> $null`

yi-Xu-0100:
#### Description

update feedback for `scoop config <xxx> $null`.

#### Checklist:

### DIFF
--- a/libexec/scoop-config.ps1
+++ b/libexec/scoop-config.ps1
@@ -147,7 +147,12 @@ if (!$name) {
     if($null -eq $value) {
         Write-Host "'$name' is not set"
     } else {
-        $value
+        # Since `scoop config <xxx> $null` is the same as `scoop config <xxx>`,
+        # so it only can remove <xxx> with `scoop config rm <xxx>`.
+        # https://github.com/ScoopInstaller/Scoop/pull/5090#issuecomment-1207389924
+        Write-Host "'$name' is '$value'."
+        Write-Host "-----"
+        Write-Host "Note: you can use 'scoop config rm $name' to remove '$name'."
     }
 }
 


niheaven:
`set_config` should remove a value if being set to `$null`, not `scoop config <xxx> $null`. You may misunderstood the test and `scoop config`.

yi-Xu-0100:
@niheaven I try to use `scoop config name $null`, but it will return value of name when the name has value. 😵‍💫

I think that the response is confused for user and need to give feedback, then I find the `set_config` can deal with the `$null`, so I think the scoop config should deal with it also.🤔

niheaven:
Since `scoop config <xxx> $null` is the same as `scoop config <xxx>`, the `$value` will be `$null` if ignored, so the command just output the config value.

In this PR, all command that may show option value will remove the option, such as:

image

yi-Xu-0100:
@niheaven I understand the reason for `scoop config name $null` not working. I update the feedback for it to notification the right way to remove.

libexec/scoop-config.ps1
```diff
@@ -147,7 +147,12 @@ if (!$name) {
     if($null -eq $value) {
         Write-Host "'$name' is not set"
     } else {
-        $value
+        # Since `scoop config <xxx> $null` is the same as `scoop config <xxx>`,
+        # so it only can remove <xxx> with `scoop config rm <xxx>`.
+        # https://github.com/ScoopInstaller/Scoop/pull/5090#issuecomment-1207389924
+        Write-Host "'$name' is '$value'."
```
rashil2000, line 153: Just a note: Using Write-Host here will prevent users from storing config values, e.g. in a script like this:
```pwsh
$updated_at = scoop config lastupdate
```

yi-Xu-0100:
@niheaven @rashil2000 I have no idea to notify, maybe only me need this. Thanks for all your review.

PR #5109, opened 2022-08-15
refactor(path): Use 'Convert-Path()' instead of 'Resolve-Path()'

niheaven:
#### Description

Use `Convert-Path()` instead of `Resolve-Path()` while the latter will *resolve* the **wildcard** characters in a path (we just want to *convert* a path from a PowerShell path to a PowerShell provider path or *expand* a path)

#### Motivation and Context

None

Internal tests passed, all functions should work well, and more tests should be made.

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 - **scoop-update:** Stash uncommitted changes before update ([#5091](https://github.com/ScoopInstaller/Scoop/issues/5091))
 - **install:** Show the running process ([#5102](https://github.com/ScoopInstaller/Scoop/issues/5102))
 
+### Code Refactoring
+
+- **path:** Use 'Convert-Path()' instead of 'Resolve-Path()' ([#5109](https://github.com/ScoopInstaller/Scoop/issues/5109))
+
 ## [v0.2.4](https://github.com/ScoopInstaller/Scoop/compare/v0.2.3...v0.2.4) - 2022-08-08
 
 ### Features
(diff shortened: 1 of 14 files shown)


niheaven:
Any comments here, guys?

rashil2000:
Are there any bugs we're facing with Resolve-Path? I don't see a reason to change what's already working.

niheaven:
As what I said, `Resolve-Path` is used to *RESOLVE* or *GUESS* the path, such as `Resolve-Path windows`, just see its examples [here](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/resolve-path?view=powershell-5.1#examples). What we used in Scoop is just `Convert-Path`, e.g. expand `.` or `~`.

And there is another benefit here, `Resolve-Path` outputs `[System.Management.Automation.PathInfo]` which may be handled incorrectly while `Convert-Path` outputs `[System.String]`.

rashil2000 (approved):

Hrxn:
> As what I said, `Resolve-Path` is used to _RESOLVE_ or _GUESS_ the path, such as `Resolve-Path windows`, just see its examples [here](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/resolve-path?view=powershell-5.1#examples). What we used in Scoop is just `Convert-Path`, e.g. expand `.` or `~`.
> 

Not really. That example is specific to `$PWD  = C:\`
`Resolve-Path` does not "guess" any paths..

> And there is another benefit here, `Resolve-Path` outputs `[System.Management.Automation.PathInfo]` which may be handled incorrectly while `Convert-Path` outputs `[System.String]`.
> 

Yeah, that's the one important difference between these two cmdlets.
One could argue that using `PathInfo` objects would be beneficial over using just strings, but this depends on the use case.
If you're dealing with paths represented as strings anyway, then `Convert-Path` is just fine, yes..

The other important difference is that `Resolve-Path` does support using `PSCredential` for credentialed access, but this is probably not relevant for scoop's case..

Issue #5094, opened 2022-08-10
[Bug]scoop shim info : ERROR: Option -global not recognized.


sebma:
#### Current Behavior
Hi, I'm trying to display the scoop shim information of the `sudo` command : 
```shell
# scoop shim info sudo
Local shim not found: sudo
But a global shim exists, run 'scoop shim info sudo -global' to show its info
```
This comment suggest us to type the command **scoop shim info sudo -global** : 
```shell
# scoop shim info sudo -global
ERROR: Option -global not recognized.
Usage: scoop shim <subcommand> [<shim_names>] [<command_path> [<args>...]] [-g(lobal)]
```
#### Expected Behavior
The shim information of `sudo` is displayed.

### System details

**Windows version:** 10 Pro

**OS architecture:** 64bit

**PowerShell version:**
```powershell
# echo $PSVersionTable.PSVersion
Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      19041  1682
```
**Additional software:** [(optional) e.g. ConEmu, Git]

#### Scoop Configuration

```json
{
    "lastUpdate":  "2022-08-10T17:20:34.7655612+02:00",
    "SCOOP_REPO":  "https://github.com/ScoopInstaller/Scoop",
    "SCOOP_BRANCH":  "master"
}
```

lewis-yeung:
Related code:

https://github.com/ScoopInstaller/Scoop/blob/4a31bd330244f7f89f16208cdddda3f9edac2d65/libexec/scoop-shim.ps1#L71-L76

rashil2000:
Discussion about avoiding `getopt` is here: https://github.com/ScoopInstaller/Scoop/pull/4736

lewis-yeung:
@rashil2000 @niheaven Alternatively, we can make `getopt` compatible with the GNU extensions to the POSIX conventions for command line options. See [The GNU C Library Reference Manual](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html):

> The argument -- terminates all options; any following arguments are treated as non-option arguments, even if they begin with a hyphen.

If so, we have to use `'--'` in PowerShell CLI since the unquoted `--` is also the terminator flag in PowerShell syntax, which makes it excluded from `$args`.

niheaven:
Hmm, it's a bug, yes. I'll fix it.

niheaven:
[...]

PR #5121, opened 2022-08-23
feat(getopt): Support option terminator (`--`)

lewis-yeung:
#### Description

Support the option terminator `--` in `getopt`, which is compatible with POSIX conventions for command line options (see [reference](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) for details).

#### Motivation and Context

This is also a prerequisite for fixing #5094.

Tested locally with updated [test/Scoop-GetOpts.Tests.ps1](https://github.com/ScoopInstaller/Scoop/blob/876d0ee2d1673c316ea14e84872eece1278fe49d/test/Scoop-GetOpts.Tests.ps1). All passed.

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Features
 
+- **getopt:** Support option terminator (`--`) ([#5121](https://github.com/ScoopInstaller/Scoop/issues/5121))
 - **scoop-(un)hold:** Support `scoop (un)hold scoop` ([#5089](https://github.com/ScoopInstaller/Scoop/issues/5089))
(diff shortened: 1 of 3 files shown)


lewis-yeung:
> but the question is, PowerShell won't pass `--` in its subprocess, so you canot get this `--` in subcommand. A `--` support in `getopt` has nothing to do with the bug.

@niheaven We can pass `'--'` which is treated as a string in PowerShell.

niheaven (approved):
LGTM, now could move on.

rashil2000 (approved):

PR #5122, opened 2022-08-23
feat(chore): Improve git.exe execution and add parallel bucket updates

### LINKED ISSUE #1615: bucket add fails behind proxy
Successfully installed scoop and after settings proxy (`scoop config proxy currentuser@default`) was able to `scoop install` however `scoop bucket add extras` fails with following error
```
Checking repo... 'https://github.com/lukesampson/scoop-extras.git' doesn't look like a valid git repository

Error given:
fatal: unable to access 'https://github.com/lukesampson/scoop-extras.git/': Couldn't resolve proxy 'default'
```

r15ch13:
#### Description

- By using the direct path to `$appdir\git\current\mingw64\bin\git.exe` this prevents spawning subprocesses and improves execution time
  - This removes calling two shims. `$scoopdir\shims\git.exe` calls `$appdir\git\current\bin\git.exe` which is also a shim.
- Fixes #1615 by replacing `currentuser@` with `:@`
- Spawns cmd.exe only if the `PROXY` config value is set and required by the git command
- Uses PS7 feature `Foreach-Object -Parallel` to run bucket updates in parallel to improve execution time
  - Not sure how to show the changelog right now, so I just added `[extras]` behind the commit hash
- Fallback to sequential updates for PS5
- Improve git log output

Using a fixed `git.exe` could be a problem if someone uses git not installed via scoop because this will not detect it with `Get-Command` anymore.

#### Motivation and Context
1. Improve updating with many buckets
2. Fixes #1615

- Locally tested

#### Checklist:

Before:
image
image

After:
image
image

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -132,7 +132,7 @@ function Invoke-Git {
     [CmdletBinding()]
     [OutputType([String])]
     param(
-        [Parameter(Mandatory = $true, Position = 0)]
+        [Parameter(Mandatory = $false, Position = 0)]
         [Alias('PSPath', 'Path')]
         [ValidateNotNullOrEmpty()]
         [String]
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -182,6 +182,23 @@ function Invoke-Git {
     return Invoke-Command $sb
 }
 
+function Invoke-GitLog {
+    [CmdletBinding()]
+    Param (
+        [Parameter(Mandatory, ValueFromPipeline)]
+        [String]$Path,
+        [Parameter(Mandatory, ValueFromPipeline)]
+        [String]$CommitHash,
+        [String]$Name = ''
+    )
+    Process {
+        if($Name) {
+            $Name = "%Cgreen$($Name.PadRight(12, ' ').SubString(0, 12))%Creset "
+        }
+        Invoke-Git -Path $Path -ArgumentList @('--no-pager', 'log', '--color', '--no-decorate', "--grep='^(chore)'", '--invert-grep', '--abbrev=12', "--format='tformat: * %C(yellow)%h%Creset %<|(72,trunc)%s $Name%C(cyan)%cr%Creset'", "$CommitHash..HEAD")
+    }
+}
+
 # helper functions
 function coalesce($a, $b) { if($a) { return $a } $b }
 
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -128,13 +128,35 @@ function setup_proxy() {
     }
 }
 
-function git_cmd {
+function Invoke-Git {
+    [CmdletBinding()]
+    [OutputType([String])]
+    param(
+        [String] $Path,
+        [PSObject[]] $ArgumentList
+    )
+
     $proxy = get_config PROXY
-    $cmd = "git $($args | ForEach-Object { "$_ " })"
-    if ($proxy -and $proxy -ne 'none') {
-        $cmd = "SET HTTPS_PROXY=$proxy&&SET HTTP_PROXY=$proxy&&$cmd"
+    $git = Get-HelperPath -Helper Git
+    $arguments = $ArgumentList -join ' '
+    if ($Path) {
+        $cmd = "`"$git`" -C `"$Path`" $arguments"
+    } else {
+        $cmd = "`"$git`" $arguments"
     }
-    cmd.exe /d /c $cmd
+
+    # invoke git if proxy is not set or not required
+    if ([String]::IsNullOrEmpty($proxy) -or $proxy -eq 'none' -or $arguments -NotMatch "\b(clone|checkout|pull|fetch|ls-remote)\b") {
+        return Invoke-Command ([scriptblock]::Create("& $cmd"))
+    }
+
+    # convert proxy setting for git
+    if ($proxy.StartsWith('currentuser@')) {
+        $proxy = $proxy.Replace("currentuser@",":@")
+    }
+
+    # use proxy with cmd.exe to prevent polluting PowerShell session
+    cmd.exe /d /c "SET HTTPS_PROXY=$proxy&&SET HTTP_PROXY=$proxy&&$cmd"
 }
 
 # helper functions
@@ -293,12 +315,16 @@ Function Test-CommandAvailable {
     Return [Boolean](Get-Command $Name -ErrorAction Ignore)
 }
 
+Function Test-GitAvailable {
+    Return [Boolean](Test-Path (Get-HelperPath -Helper Git) -ErrorAction Ignore)
+}
+
 function Get-HelperPath {
     [CmdletBinding()]
     [OutputType([String])]
     param(
         [Parameter(Mandatory = $true, Position = 0, ValueFromPipeline = $true)]
-        [ValidateSet('7zip', 'Lessmsi', 'Innounp', 'Dark', 'Aria2', 'Zstd')]
+        [ValidateSet('Git', '7zip', 'Lessmsi', 'Innounp', 'Dark', 'Aria2', 'Zstd')]
         [String]
         $Helper
     )
@@ -307,6 +333,7 @@ function Get-HelperPath {
     }
     process {
         switch ($Helper) {
+            'Git' { $HelperPath = "$(versiondir 'git' 'current')\mingw64\bin\git.exe" }
             '7zip' {
                 $HelperPath = Get-AppFilePath '7zip' '7z.exe'
                 if ([String]::IsNullOrEmpty($HelperPath)) {
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -132,31 +132,54 @@ function Invoke-Git {
     [CmdletBinding()]
     [OutputType([String])]
     param(
-        [String] $Path,
-        [String[]] $ArgumentList
+        [Parameter(Mandatory = $true, Position = 0)]
+        [Alias('PSPath', 'Path')]
+        [ValidateNotNullOrEmpty()]
+        [String]
+        $WorkingDirectory,
+        [Parameter(Mandatory = $true, Position = 1)]
+        [Alias('Args')]
+        [String[]]
+        $ArgumentList
     )
 
     $proxy = get_config PROXY
     $git = Get-HelperPath -Helper Git
     $arguments = $ArgumentList -join ' '
-    if ($Path) {
-        $cmd = "`"$git`" -C `"$Path`" $arguments"
-    } else {
-        $cmd = "`"$git`" $arguments"
+    $cmd = "`"$git`" $arguments"
+
+    if ($WorkingDirectory) {
+        $cmd = "`"$git`" -C `"$WorkingDirectory`" $arguments"
+    }
+    $sb = [scriptblock]::Create("& $cmd")
+
+    if([String]::IsNullOrEmpty($proxy) -or $proxy -eq 'none')  {
+        return Invoke-Command $sb
     }
 
-    # invoke git if proxy is not set or not required
-    if ([String]::IsNullOrEmpty($proxy) -or $proxy -eq 'none' -or $arguments -NotMatch "\b(clone|checkout|pull|fetch|ls-remote)\b") {
-        return Invoke-Command ([scriptblock]::Create("& $cmd"))
+    if($arguments -Match '\b(clone|checkout|pull|fetch|ls-remote)\b') {
+        $old_https = $env:HTTPS_PROXY
+        $old_http = $env:HTTP_PROXY
+        try {
+            # convert proxy setting for git
+            if ($proxy.StartsWith('currentuser@')) {
+                $proxy = $proxy.Replace('currentuser@', ':@')
+            }
+            $env:HTTPS_PROXY = $proxy
+            $env:HTTP_PROXY = $proxy
+            return Invoke-Command $sb
+        }
+        catch {
+            error $_
+            return
+        }
+        finally {
+            $env:HTTPS_PROXY = $old_https
+            $env:HTTP_PROXY = $old_http
+        }
     }
 
-    # convert proxy setting for git
-    if ($proxy.StartsWith('currentuser@')) {
-        $proxy = $proxy.Replace("currentuser@",":@")
-    }
-
-    # use proxy with cmd.exe to prevent polluting PowerShell session
-    cmd.exe /d /c "SET HTTPS_PROXY=$proxy&&SET HTTP_PROXY=$proxy&&$cmd"
+    return Invoke-Command $sb
 }
 
 # helper functions
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -133,7 +133,7 @@ function Invoke-Git {
     [OutputType([String])]
     param(
         [String] $Path,
-        [PSObject[]] $ArgumentList
+        [String[]] $ArgumentList
     )
 
     $proxy = get_config PROXY
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -192,8 +192,11 @@ function Invoke-GitLog {
         [String]$Name = ''
     )
     Process {
-        if($Name) {
-            $Name = "%Cgreen$($Name.PadRight(12, ' ').SubString(0, 12))%Creset "
+        if ($Name) {
+            if ($Name.Length -gt 12) {
+                $Name = "$($Name.Substring(0, 10)).."
+            }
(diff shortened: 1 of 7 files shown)


lib/core.ps1
```diff
@@ -307,6 +333,7 @@ function Get-HelperPath {
     }
     process {
         switch ($Helper) {
+            'Git' { $HelperPath = "$(versiondir 'git' 'current')\mingw64\bin\git.exe" }
```
rashil2000, line 336: Well this is a problem because I don't use Git installed by Scoop 😅

Ideally there should be a fallback here, which can `Test-Path` this path and fall back to `Get-Command`.

r15ch13: Dang it, I knew it. :grin:

Yeah, will implement a fallback :smile:

rasa:
This will almost certainly fix https://github.com/ScoopInstaller/Scoop/issues/3877 "Git spawns thousands of times".

libexec/scoop-update.ps1
```diff
@@ -56,14 +56,18 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
 }
 $show_update_log = get_config SHOW_UPDATE_LOG $true
 
-function update_scoop($show_update_log) {
+function Sync-Scoop {
+    [CmdletBinding()]
+    Param (
+        [Boolean]$OutputLog = $true
```
niheaven, line 62:
```suggestion
        [Switch]$Log
```

libexec/scoop-update.ps1
```diff
@@ -113,76 +117,109 @@ function update_scoop($show_update_log) {
...
+            Invoke-Git -Path $currentdir -ArgumentList "reset --hard origin/$configBranch -q"
         } else {
-            git_cmd -C "`"$currentdir`"" pull -q
+            Invoke-Git -Path $currentdir -ArgumentList "pull -q"
         }
 
         $res = $lastexitcode
-        if ($show_update_log) {
-            git -C "$currentdir" --no-pager log --no-decorate --grep='^(chore)' --invert-grep --format='tformat: * %C(yellow)%h%Creset %<|(72,trunc)%s %C(cyan)%cr%Creset' "$previousCommit..HEAD"
+        if ($OutputLog) {
```
niheaven, line 138:
```suggestion
        if ($Log) {
```

libexec/scoop-update.ps1
```diff
@@ -113,76 +117,109 @@ function update_scoop($show_update_log) {
...
-    if (!(Test-CommandAvailable git)) { abort "Scoop uses Git to update main bucket and others. Run 'scoop install git' and try again." }
-
-    foreach ($bucket in Get-LocalBucket) {
-        Write-Host "Updating '$bucket' bucket..."
+function Sync-Bucket {
+    [CmdletBinding()]
+    Param (
+        [Parameter(Mandatory, ValueFromPipeline)]
+        [PSObject[]]$Name,
+        [Boolean]$OutputLog = $true
```
niheaven, line 155:
```suggestion
        [Switch]$Log
```

libexec/scoop-update.ps1
```diff
@@ -113,76 +117,109 @@ function update_scoop($show_update_log) {
...
-    # check for git
-    if (!(Test-CommandAvailable git)) { abort "Scoop uses Git to update main bucket and others. Run 'scoop install git' and try again." }
-
-    foreach ($bucket in Get-LocalBucket) {
-        Write-Host "Updating '$bucket' bucket..."
+function Sync-Bucket {
+    [CmdletBinding()]
+    Param (
+        [Parameter(Mandatory, ValueFromPipeline)]
+        [PSObject[]]$Name,
```
niheaven, line 154:
```suggestion
        [String[]]$Name,
```

libexec/scoop-update.ps1
```diff
@@ -113,76 +117,109 @@ function update_scoop($show_update_log) {
...
 
-        $previousCommit = git -C "$bucketLoc" rev-parse HEAD
-        git_cmd -C "`"$bucketLoc`"" pull -q
-        if ($show_update_log) {
-            git -C "$bucketLoc" --no-pager log --no-decorate --grep='^(chore)' --invert-grep --format='tformat: * %C(yellow)%h%Creset %<|(72,trunc)%s %C(cyan)%cr%Creset' "$previousCommit..HEAD"
+        Write-Host "Updating '$Name' bucket..."
+
+        $previousCommit = Invoke-Git -Path $bucketLoc -ArgumentList "rev-parse HEAD"
+        Invoke-Git -Path $bucketLoc "pull -q"
+        if ($OutputLog) {
```
niheaven, line 170:
```suggestion
        if ($Log) {
```

libexec/scoop-update.ps1
```diff
@@ -113,76 +117,109 @@ function update_scoop($show_update_log) {
...
+            Invoke-Git -Path $bucketLoc -ArgumentList "--no-pager log --no-decorate --grep='^(chore)' --invert-grep --format='tformat: * %C(yellow)%h%Creset %<|(72,trunc)%s %C(cyan)%cr%Creset [$Name]' '$previousCommit..HEAD'"
+        }
+    }
+}
+
+function Sync-Buckets {
+    Param (
+        [Boolean]$OutputLog = $true
```
niheaven, line 178:
```suggestion
        [Switch]$Log
```

libexec/scoop-update.ps1
```diff
@@ -113,76 +117,109 @@ function update_scoop($show_update_log) {
...
+        # Parallel parameter is available since PowerShell 7
+        $buckets | Where-Object { $_.valid } | ForEach-Object -ThrottleLimit 5 -Parallel {
+            . "$using:PSScriptRoot\..\lib\core.ps1"
+
+            $bucketLoc = $_.path
+            $name = $_.name
+
+            $previousCommit = Invoke-Git -Path $bucketLoc -ArgumentList "rev-parse HEAD"
+            Invoke-Git -Path $bucketLoc -ArgumentList "pull -q"
+            if ($using:OutputLog) {
```
niheaven, line 216:
```suggestion
            if ($using:Log) {
```

libexec/scoop-update.ps1
```diff
@@ -113,76 +117,109 @@ function update_scoop($show_update_log) {
...
+                Invoke-Git -Path $bucketLoc -ArgumentList "--no-pager log --no-decorate --grep='^(chore)' --invert-grep --format='tformat: * %C(yellow)%h [$name]%Creset %<|(72,trunc)%s %C(cyan)%cr%Creset' '$previousCommit..HEAD'"
+            }
+        }
+    } else {
+        $buckets | Where-Object { $_.valid } | ForEach-Object { Sync-Bucket $_.name -OutputLog $OutputLog }
```
niheaven, line 221:
```suggestion
        $buckets | Where-Object { $_.valid } | ForEach-Object { Sync-Bucket $_.name -Log:$Log }
```

libexec/scoop-update.ps1
```diff
@@ -319,8 +356,8 @@ if (-not ($apps -or $all)) {
         error 'scoop update: --no-cache is invalid when <app> is not specified.'
         exit 1
     }
-    update_scoop $show_update_log
-    update_bucket $show_update_log
+    Sync-Scoop -OutputLog $show_update_log
+    Sync-Buckets -OutputLog $show_update_log
```
niheaven, line 360:
```suggestion
    Sync-Scoop -Log:$show_update_log
    Sync-Buckets -Log:$show_update_log
```

libexec/scoop-update.ps1
```diff
@@ -334,8 +371,8 @@ if (-not ($apps -or $all)) {
     $apps_param = $apps
 
     if ($updateScoop) {
-        update_scoop $show_update_log
-        update_bucket $show_update_log
+        Sync-Scoop -OutputLog $show_update_log
+        Sync-Buckets -OutputLog $show_update_log
```
niheaven, line 375:
```suggestion
        Sync-Scoop -Log:$show_update_log
        Sync-Buckets -Log:$show_update_log
```

niheaven (commented):
Using standard PowerShell argument `[Switch]Log`, see https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/activity-parameters?view=powershell-5.1

lib/core.ps1
```diff
@@ -128,13 +128,35 @@ function setup_proxy() {
     }
 }
 
-function git_cmd {
+function Invoke-Git {
+    [CmdletBinding()]
+    [OutputType([String])]
+    param(
+        [String] $Path,
+        [PSObject[]] $ArgumentList
```
niheaven, line 136:
```suggestion
        [String[]] $ArgumentList
```

And could we use `Invoke-ExternalCommand()` here? (The function cannot add EnvVars now, I'll submit PR to support it)

r15ch13: I tried using `Invoke-ExternalCommand()` but the redirecting if `stdout` breaks the output of git. E.g.:

image

r15ch13:
I tried using `Invoke-ExternalCommand()` but the redirecting if `stdout` breaks the output of git. E.g.:

image

Used `Invoke-Git()` code:
```pwsh
function Invoke-Git {
    [CmdletBinding()]
    [OutputType([String])]
    param(
        [Parameter(Mandatory = $true, Position = 0)]
        [Alias('PSPath', 'Path')]
        [ValidateNotNullOrEmpty()]
        [String]
        $WorkingDirectory,
        [Parameter(Mandatory = $true, Position = 1)]
        [Alias('Args')]
        [String[]]
        $ArgumentList
    )

    $proxy = get_config PROXY
    $git = Get-HelperPath -Helper Git

    $EnvironmentVariables = @{}
    if($proxy -and $proxy -ne 'none')  {
        if(($ArgumentList -join ' ') -Match "\b(clone|checkout|pull|fetch|ls-remote)\b") {
            # convert proxy setting for git
            if ($proxy.StartsWith('currentuser@')) {
                $proxy = $proxy.Replace("currentuser@",":@")
            }
            $EnvironmentVariables.Add('HTTPS_PROXY', $proxy)
            $EnvironmentVariables.Add('HTTP_PROXY', $proxy)
        }
    }

    Invoke-ExternalCommand -FilePath $git -WorkingDirectory $WorkingDirectory -ArgumentList $ArgumentList -EnvironmentVariables $EnvironmentVariables
}
```

niheaven:
`True` is the output of `Invoke-ExternalCommand` and should be omitted by `Out-Null`.

And for the colorful output, a `--color` switch must be explicitly added to git's argument (can be done in `Invoke-Git`). But for `--oneline`, the ref inside bracket are missing (i.e., `(HEAD -> master, origin/master, origin/HEAD)`)

rashil2000:
Yes, the `--color` flag should be added

r15ch13:
And `--decorate` shows the missing `(HEAD -> master, origin/master, origin/HEAD)`.

rashil2000:
I think we can skip on using `iec` for Invoke-Git for now. When `iec` becomes more robust, we can switch to using it.

r15ch13:
Unified the output for both bucket update methods to:
image

Uses try-catch-finally to revert the environment variables.

niheaven:
Is the `@()` necessary for `ArgumentList`? If you use comma (,), it is an [Array] for the argument already.

r15ch13:
Just a rebase onto the current dev branch.

r15ch13:
Abbreviated the hash to 12 hex digits to prevent inconsistent output length.

Ready for review :smile:

niheaven:
It works for me, and two questions here:

1. `Sync-Bucket` is only used by `Sync-Buckets` once, so why not copy its code in `Sync-Buckets` and rename latter to `Sync-Bucket`?
2. Buckets name and updated time are separated by space, should updated time be aligned?

```pwsh
    if ($PSVersionTable.PSVersion.Major -ge 7) {
        # Parallel parameter is available since PowerShell 7
        $buckets | Where-Object { $_.valid } | ForEach-Object -ThrottleLimit 5 -Parallel {
            . "$using:PSScriptRoot\..\lib\core.ps1"

            $bucketLoc = $_.path
            $name = $_.name

            $previousCommit = Invoke-Git -Path $bucketLoc -ArgumentList @('rev-parse', 'HEAD')
            Invoke-Git -Path $bucketLoc -ArgumentList @('pull', '-q')
            if ($using:Log) {
                Invoke-Git -Path $bucketLoc -ArgumentList @('--no-pager', 'log', '--color', '--no-decorate', "--grep='^(chore)'", '--invert-grep', '--abbrev=12', "--format='tformat: * %C(yellow)%h%Creset %<|(72,trunc)%s %Cgreen$name%Creset %C(cyan)%cr%Creset'", "$previousCommit..HEAD")
            }
        }
    } else {
        $buckets | Where-Object { $_.valid } | ForEach-Object {
            $bucketLoc = $_.path
            $name = $_.name

            $previousCommit = Invoke-Git -Path $bucketLoc -ArgumentList @('rev-parse', 'HEAD')
            Invoke-Git -Path $bucketLoc -ArgumentList @('pull', '-q')
            if ($Log) {
                Invoke-Git -Path $bucketLoc -ArgumentList @('--no-pager', 'log', '--color', '--no-decorate', "--grep='^(chore)'", '--invert-grep', '--abbrev=12', "--format='tformat: * %C(yellow)%h%Creset %<|(72,trunc)%s %Cgreen$name%Creset %C(cyan)%cr%Creset'", "$previousCommit..HEAD")
            }
        }
    }
```

image

r15ch13:
The idea for `Sync-Bucket` was that it just works with one bucket. For later use:tm: :smile: (whatever that may be)
Didn't want to truncate the bucket name for the alignment. But it could determine the longest name and align it.

r15ch13:
new output:
image

niheaven:
image

Rename it to `Sync-Bucket`?

r15ch13:
Yes, but it doesn't update one bucket. So the name is misleading, I know PSScriptAnalyzer doesn't like it :smile:
Any other name suggestions?

rashil2000:
> new output: image

This output is quite wide, IMO. Default terminal emulator widths are like 30-40 columns. May I suggest outputting a PSObject list for this? (similar to `scoop list`)

We will lose the individual colors, but PowerShell will automatically take care of padding, alignment and terminal adjustment. (Not to mention the output will become parseable by PowerShell. Can't think of a use case for now, but who knows 😅)

r15ch13:
image
Fits in both default terminals of Windows 11

This is just output for the user. If someone wants parse this it would be better to use Git directly on the buckets itself :smile:

niheaven (approved):

niheaven:
@r15ch13 Occasional the pull process will be outputted, why?

image

r15ch13:
hm, I don't know 🤷🏽
Never happend while testing this

niheaven:
Tend to be appeared after a long time pull. (and maybe a poor network?)

Have seen it several times when I don't `scoop update` for several days.

Change `LAST_UPDATE` and reset bucket to weeks or months ago and run `scoop update`? (I'll change it and test)

niheaven:
Wow, it is.

image

I've reset my buckets repo to 3 weeks ago (one or two is enough) and run `scoop update`, the red output appears again.

r15ch13:
How did you reset the repo? I tried a hard reset to an older commit but it doesn't hit the network to change back to the latest.

rashil2000:
Also the `scoop bucket` subcommands seem to be broken after this PR:
image

r15ch13:
Can't reproduce both issues. 😕 

Output of `git --version`?
image

rashil2000:
`git version 2.39.0.windows.1`

niheaven:
image

For specific git version?

niheaven:
@r15ch13 If `git reset` gives `Updating files: 100% (513/513), done.` Then `scoop update` will output the red `Updating files` too.

image

HUMORCE:
screenshot_20221218153543
screenshot_20221218153548

cannot reproduce, both Powershell 7.2 and 5.1.

`git version 2.39.0.windows.1`

rashil2000:
My Scoop home is located at a non default directory, with a space in the path. Maybe that's the reason?

niheaven:
See the `Updating files` prompt, @HUMORCE you haven't met them, so the output is fine.

For @rashil2000's bug, I have a non-default location of scoop w/o space, and cannot reproduce. Maybe it is the space?

rashil2000:
Looks like all commands related to Git have an issue when Scoop home is a spaced path.

PR #5125, opened 2022-08-26
refactor(scoop-shim): Use `getopt` to parse arguments

lewis-yeung:
### Description

Use `getopt` to parse arguments, which avoids some problems. BTW, add some logic and tweak some error messages.

### Motivation and Context

These problems will be resolved:

1. For example, run `scoop shim add myapp ''`, the second parameter (`<command_path>`) gets an empty value, so Scoop will throw an exception, which is not so friendly. :

	```
	Cannot validate argument on parameter 'Name'. The argument is null or empty. Provide an argument that is not null or empty, and then try the command again.
	```

2. For example, run `scoop shim list *`, the regex pattern `*` is invalid, so Scoop will throw an exception, which is not so friendly. :

	```
	Invalid pattern '*' at offset 1. Quantifier {x,y} following nothing.
	```

3. The `-g`/`--global` option cannot be parsed in `scoop shim ...`. See #5094.

### How Has This Been Tested?

Now with this:

1. `scoop shim add myapp ''` shows an error:

	```
	ERROR <command_path> must be specified for subcommand 'add'
	```

2. `scoop shim list *` shows an error:

	```
	ERROR: Invalid pattern: *
	```

3. To create a local shim `myapp` with `arg` and `--global` as arguments:

	```powershell
	scoop shim add myapp 'D:\path\myapp.exe' '--' arg --global
	```

	Note that we have to use a QUOTED command option terminator (`'--'`) in PowerShell CLI.

4. To create a global shim `myapp` with `arg` as the argument:

	```powershell
	scoop shim add -g myapp 'D:\path\myapp.exe' arg
	```

No major changes to other subcommands.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 ### Code Refactoring
 
 - **path:** Use 'Convert-Path()' instead of 'Resolve-Path()' ([#5109](https://github.com/ScoopInstaller/Scoop/issues/5109))
+- **scoop-shim:** Use `getopt` to parse arguments ([#5125](https://github.com/ScoopInstaller/Scoop/issues/5125))
 
 ### Builds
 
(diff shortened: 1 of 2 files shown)


niheaven:
I'll review it in a few days, thank you for the PR.

lewis-yeung:
@niheaven Any suggestions? :)

niheaven:
Just see my commit, now LGTM.

lewis-yeung:
Shall we move on?

niheaven (approved):

PR #5176, opened 2022-10-10
Import-File-Tests.ps1: Add more file exceptions

AntonOks:


test/Import-File-Tests.ps1
```diff
@@ -9,8 +9,8 @@ Describe 'Style constraints for non-binary project files' {
         # gather all files except '*.exe', '*.zip', or any .git repository files
         $repo_files |
             Where-Object { $_.fullname -inotmatch $($project_file_exclusions -join '|') } |
-            Where-Object { $_.fullname -inotmatch '(.exe|.zip|.dll)$' } |
-            Where-Object { $_.fullname -inotmatch '(unformatted)' }
+            Where-Object { $_.fullname -inotmatch '(.exe|.zip|.dll|.reg|.ps1|.md)$' } |
+            Where-Object { $_.fullname -inotmatch '(unformated)' }
```
niheaven, line 13:
```suggestion
            Where-Object { $_.fullname -inotmatch '(unformatted)' }
```

rashil2000:
Why is this required?

niheaven:
It's not required except for the typo.

AntonOks:
> It's not required except for the typo.

@niheaven - Why do you think / say, it's not needed?
You for sure know what this scripts and the variable is good for, right? From what I understand, expact and want to have... this change prevents some tests, which are ment to be ran on the bucket json files... and not and files one could also host on "his" scoop repo, like additional powershell scripts, registry files and the like... and exactly thats the "errors" I do NOT need / want to see in my scoop repo...

niheaven:
In `reg` file? `md` and `ps1` files should pass the tests IMO, they are for plain text.

What's error with this now?

AntonOks:
error messages seen, as difined in the test file, where i.e. "end of line" in the last line and so on...


 

I doubt the SCOOP related / defined tests (!!!!) make any sense for (own) registry files, right?

BTW., beside this test here, there are also "pester test", which do check the PS1 files, as far as I remember.

And the "SCOOP defined json format with key words and so on, are alos tested.. somewhere else?!

 

Gesendet: Dienstag, 12. September 2023 um 03:54 Uhr
Von: "Hsiao-nan Cheung" ***@***.***>
An: "ScoopInstaller/Scoop" ***@***.***>
Cc: "Anton Oks" ***@***.***>, "Author" ***@***.***>
Betreff: Re: [ScoopInstaller/Scoop] Import-File-Tests.ps1: Add more file exceptions (PR #5176)


 

In reg file? md and ps1 files should pass the tests IMO, they are for plain text.

What's error with this now?

PR #5235, opened 2022-11-01
refactor(unix): Remove unix.ps1

### LINKED ISSUE #737: nodejs - specified path is too long
I tried to install nodejs with:

    scoop install nodejs

and got this error in my log file:

    === Logging started: 2016-03-11  23:36:53 ===
    Error 1320. The specified path is too long: C:\Users\StephenWalsh\AppData\Local\scoop\apps\nodejs\5.7.0\_scoop_extract\nodejs\node_modules\npm\node_modules\node-gyp\node_modules\npmlog\node_modules\gauge\node_modules\lodash.padright\node_modules\lodash._createpadding\node_modules\lodash.repeat
    === Logging stopped: 2016-03-11  23:37:06 ===

Any thoughts on how to resolve?

niheaven:
#### Description

Uniform paths on Windows and Linux/macOS, and remove unused `unix.ps1`.

Scoop's default folder is under `UserProfile` (`C:\Users\<user>` on Windows, `/home/<user>` on Linux/macOS), and global folder is under `CommonApplicationData` (`C:\ProgramData` on Windows, `/usr/share` on Linux/macOS).

Summaries:

||User Root (`UserProfile`)|User Root (`LocalApplicationData`)|Global Root|
|---|---|---|---|
|Windows|`C:\Users\<user>\scoop`|`C:\Users\<user>\AppData\Local\scoop`|`C:\ProgramData\scoop`|
|Linux/macOS|`/home/<user>/scoop`|`/home/<user>/.local/share/scoop`|`/usr/share/scoop`|

**RFC**

As said in https://learn.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=netframework-4.5, `UserProfile` shouldn't be used to store apps and/or data.

But scoop has changed its root folder from `LocalApplicationData` to `UserProfile` in https://github.com/ScoopInstaller/Scoop/commit/fc54b2fd58b6f8da0930cfe5dd7b7a000d24036c to fix #737, there may not be sufficient reason to change back.

#### Motivation and Context

Simplify code base and development on Linux/macOS platform.

There aren't tests for these config.

#### Checklist:


chawyehsu:
> C:\Users\<user>\AppData\Local

Not a fan of this, we changed to ~/scoop and resolved the long path issue in 2016. No reason to revert it back to break things again.

niheaven:
Then change the special folder to `UserProfile` and anything else is the same.

PS, does the long path issue still existed? And is it the same one as #4327?

niheaven:
I've updated summaries table, have all locations. So keep it as is? @ScoopInstaller/maintainers

rashil2000:
> C:\Users\<user>\AppData\Local

I'm in favor of this location as well (I don't like polluting my user home folder, and besides, LocalAppData is the recommended location for installing apps), but this is a **HUGE** breaking change, so I'm not sure if we should do this.

We should capitalize the `S` in `scoop`. Like `C:\Users\<user>\AppData\Local\Scoop` it looks more professional this way.

Also, we need to change backslashes to forward slashes, otherwise Scoop will remain incompatible on Unix.

rashil2000:
Also @niheaven, did you update the branch protection rules for buckets? The excavator action is failing in Main bucket.

niheaven:
> Also @niheaven, did you update the branch protection rules for buckets? The excavator action is failing in Main bucket.

Yes，I forgot excavator's permission😅 Now protection rules are disabled.

niheaven:
> Also, we need to change backslashes to forward slashes, otherwise Scoop will remain incompatible on Unix.

PowerShell uses backslashes even on Unix, and can coerce them easily. The only exception is .NET, which only recognizes slashes on Unix. But a good news here is .NET could use slashes on Windows too.

Please refer to Pester 5 test files (Import-Bucket-Tests and Manifest.Tests), they are both passed on Ubuntu and macOS.

chawyehsu:
> I'm in favor of this location as well (I don't like polluting my user home folder, and besides, LocalAppData is the recommended location for installing apps)

> it looks more professional this way.

If this is what you believe in a _professional way_, then you should not use Scoop to manage apps. Scoop does not always follow Windows conventions in design, it provides a *nix user interface while its internal implementation evolves. The codebase still uses the architecture of `/bin`, `/lib`, and `/libexec`.

If this is what you believe in a _professional way_, I assume that you want to challenge the core philosophy of Scoop. Scoop puts apps' data to a `persist` folder rather than the `~\AppData\Romaing`. Lots of the manifests are crafted carefully by contributors to persist data to the `persist` folder. Does this look unprofessional to you? Do you want to break this in a "professional way", in other words, the _Windows conventional way_ you may say? Then what about the cache, should it go to `$env:TEMP` since that will be a _professional way_? If so, you're actually challenging the philosophy.

`C:\Users\<user>\AppData\Local\scoop` has been practiced and it's terrible because of the long path hell on Windows. And the problem is still there despite the `LongPathsEnabled` option. Ignoring the problem and changing to use it thinking it's a _professional way_ is in fact not a professional way to me.

niheaven:
Hmm, I think @rashil2000's *professional way* is **S**coop, the first capital letter?

I've changed back to `UserProfile` and now except for Unix Global Path, all is the same as before.

rashil2000:
Jeez @chawyehsu, I was talking about the capital 'S'. I'm actually against the location change (see my comment above).

chawyehsu:
I'm sorry for the misunderstanding. But I still stand for the point I told: there shouldn't be a judgment of what's _more professional_ in this case. Why would `S` be more _professional_ than another that has been used since day one... `S` may be more **conventional** or **classical**, but `s` is definitely not so-called less professional.

Further, I would use the phrase _Like `C:\Users\<user>\Scoop`..._ to describe the case for less ambiguity.

rashil2000 (approved):

chawyehsu:
I would leave `scoop` as-is rather than `Scoop` since you want it to be used on *nix based on the purpose of this PR, and there is no point in changing it to a so-called more _professional_ capitalized one. Because it is not tied to the Windows platform.

All the other changes are fine.

niheaven:
That's right, on Linux/macOS lower capital such as `scoop` is more common since it's capital sensitive.

So change back to `scoop` and that's all.

PR #5321, opened 2023-01-06
Move nirsoft bucket to ScoopInstaller org

rasa:
Per https://github.com/ScoopInstaller/Scoop/discussions/5293


niheaven:
Hmm, base on `develop`?

PR #4818, opened 2022-03-16
Improve local search performance

bartvanandel:
#### Description
Improve local search performance by pruning candidates using `git ls-files` and `git grep`. This drastically reduces the number of files that are read into memory to find the desired results.

The code accounts for buckets that are not maintained using `git`. Not sure if that really applies, but for buckets that are not maintained using `git`, the code will fall back to the previous approach.

#### Motivation and Context

Closes #4239

#### How Has This Been Tested?
Compared the output of the previous search method with the new method. The code uses standard PowerShell and git functions that have existed for a long time.

#### Checklist:

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).
- [ ] ~I have updated the documentation accordingly.~ N/A. Well, I sparsely documented the approach in code.
- [ ] ~I have updated the tests accordingly.~ N/A, there were no pre-existing tests.

PR #5324, opened 2023-01-06
Another attempt to improve the search performance

### LINKED ISSUE #4239: Improve Search Performance

r15ch13:
#### Description
This is another attempt to improve the local search performance.
- Not using the `manifest()` function
- Use only one loop over `$apps`
- Push results to `$list` directly
- Use `[System.IO.File]::ReadAllText()` instead of `Get-Content`

#### Motivation and Context

Closes #4239

Relates to #4818

```pwsh
Measure-Command { scoop search telegram }
Measure-Command { .\bin\scoop.ps1 search telegram }
```

image

#### Checklist:

### DIFF
--- a/libexec/scoop-search.ps1
+++ b/libexec/scoop-search.ps1
@@ -11,6 +11,11 @@ param($query)
 
 $list = [System.Collections.Generic.List[PSCustomObject]]::new()
 
+if(!$query -or $query -in @($null, '-h', '--help', '/?')) {
+    my_usage
+    exit 0
+}
+
 try {
     $query = New-Object Regex $query, 'IgnoreCase'
 } catch {
--- a/libexec/scoop-search.ps1
+++ b/libexec/scoop-search.ps1
@@ -53,7 +53,7 @@ function search_bucket($bucket, $query) {
             $bin = bin_match $json $query
             if ($bin) {
                 $list.Add([PSCustomObject]@{
-                    Name = $_
+                    Name = $_.BaseName
                     Version = $json.Version
                     Source = $bucket
                     Binaries = $bin -join ' | '
--- a/libexec/scoop-search.ps1
+++ b/libexec/scoop-search.ps1
@@ -65,7 +65,6 @@ function search_bucket($bucket, $query) {
     $apps = Get-ChildItem (Find-BucketDirectory $bucket) -Filter '*.json' -Recurse
 
     $apps | ForEach-Object {
-        # $manifest = [System.IO.File]::ReadAllText($_.FullName) | ConvertFrom-Json -ErrorAction Continue
         $json = [System.Text.Json.JsonDocument]::Parse([System.IO.File]::ReadAllText($_.FullName))
         $name = $_.BaseName
 
@@ -73,18 +72,44 @@ function search_bucket($bucket, $query) {
             $list.Add([PSCustomObject]@{
                 Name = $name
                 Version = $json.RootElement.GetProperty("version")
-                # Version = $manifest.Version
                 Source = $bucket
                 Binaries = ""
             })
         } else {
-            # $bin = bin_match $manifest $query
             $bin = bin_match_json $json $query
             if ($bin) {
                 $list.Add([PSCustomObject]@{
                     Name = $name
                     Version = $json.RootElement.GetProperty("version")
-                    # Version = $manifest.Version
+                    Source = $bucket
+                    Binaries = $bin -join ' | '
+                })
+            }
+        }
+    }
+}
+
+# fallback function for PowerShell 5
+function search_bucket_legacy($bucket, $query) {
+    $apps = Get-ChildItem (Find-BucketDirectory $bucket) -Filter '*.json' -Recurse
+
+    $apps | ForEach-Object {
+        $manifest = [System.IO.File]::ReadAllText($_.FullName) | ConvertFrom-Json -ErrorAction Continue
+        $name = $_.BaseName
+
+        if ($name -match $query) {
+            $list.Add([PSCustomObject]@{
+                Name = $name
+                Version = $manifest.Version
+                Source = $bucket
+                Binaries = ""
+            })
+        } else {
+            $bin = bin_match $manifest $query
+            if ($bin) {
+                $list.Add([PSCustomObject]@{
+                    Name = $name
+                    Version = $manifest.Version
                     Source = $bucket
                     Binaries = $bin -join ' | '
                 })
@@ -151,7 +176,11 @@ function search_remotes($query) {
 }
 
 Get-LocalBucket | ForEach-Object {
-    search_bucket $_ $query
+    if ($PSVersionTable.PSVersion.Major -ge 7) {
+        search_bucket $_ $query
+    } else {
+        search_bucket_legacy $_ $query
+    }
 }
 
 if ($list.Count -gt 0) {
--- a/libexec/scoop-search.ps1
+++ b/libexec/scoop-search.ps1
@@ -9,7 +9,7 @@ param($query)
 . "$PSScriptRoot\..\lib\manifest.ps1" # 'manifest'
 . "$PSScriptRoot\..\lib\versions.ps1" # 'Get-LatestVersion'
 
-$list = @()
+$list = [System.Collections.Generic.List[PSCustomObject]]::new()
 
 try {
     $query = New-Object Regex $query, 'IgnoreCase'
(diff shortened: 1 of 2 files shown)


niheaven:
WIP or Ready for review?

rashil2000:
Some more areas for optimization:
- Since we are listing lots of files in function `apps_in_bucket`, using `[System.IO.Directory]::EnumerateFiles()` instead of `Get-ChildItem` should be much faster ([ref](https://stackoverflow.com/questions/14970692/powershell-io-directory-find-file-types-in-all-subdirectories))
- Since we are deserializing a lot of JSON, using `[System.Text.Json.Utf8JsonReader]` should be faster than `ConvertFrom-Json`
- While searching in remotes, since we are making multiple requests, a .NET method should be faster than `Invoke-WebRequest`
- I'm not sure I understand this, but why do we use `Select-Object -ExpandProperty` instead of just directly accessing the field on lines 69, 90 and 100?

libexec/scoop-search.ps1
```diff
@@ -37,19 +37,31 @@ function bin_match($manifest, $query) {
...
 
 function search_bucket($bucket, $query) {
-    $apps = apps_in_bucket (Find-BucketDirectory $bucket) | ForEach-Object { @{ name = $_ } }
-
-    if ($query) {
-        $apps = $apps | Where-Object {
-            if ($_.name -match $query) { return $true }
-            $bin = bin_match (manifest $_.name $bucket) $query
+    $bucket_dir = Find-BucketDirectory $bucket
+    $apps = apps_in_bucket $bucket_dir
```
niheaven, line 41:
```suggestion
    $apps = Get-ChildItem (Find-BucketDirectory $bucket) -Filter '*.json' -Recurse
```

libexec/scoop-search.ps1
```diff
@@ -37,19 +37,31 @@ function bin_match($manifest, $query) {
...
+
+    $apps | ForEach-Object {
+        $json = [System.IO.File]::ReadAllText("$bucket_dir\$_.json") | ConvertFrom-Json -ErrorAction Continue
```
niheaven, line 44:
```suggestion
        $json = [System.IO.File]::ReadAllText($_.FullName) | ConvertFrom-Json -ErrorAction Continue
```

libexec/scoop-search.ps1
```diff
@@ -37,19 +37,31 @@ function bin_match($manifest, $query) {
...
+
+        if($app.name -match $query) {
+            $list.Add([PSCustomObject]@{
+                Name = $_
```
niheaven, line 48:
```suggestion
                Name = $_.BaseName
```

libexec/scoop-search.ps1
```diff
@@ -37,19 +37,31 @@ function bin_match($manifest, $query) {
...
+                Source = $bucket
+                Binaries = ""
+            })
+        } else {
+            $bin = bin_match $json $query
             if ($bin) {
-                $_.bin = $bin
-                return $true
+                $list.Add([PSCustomObject]@{
+                    Name = $_
```
niheaven, line 57:
```suggestion
                    Name = $_.BaseName
```

niheaven (changes requested):
Just for nested manifests

niheaven:
> Since we are listing lots of files in function apps_in_bucket, using [System.IO.Directory]::EnumerateFiles() instead of Get-ChildItem should be much faster ([ref](https://stackoverflow.com/questions/14970692/powershell-io-directory-find-file-types-in-all-subdirectories))

image

Sorry but I found `Get-ChildItem` is faster than `[System.IO.Directory]::EnumerateFiles()` here.

r15ch13:
> * Since we are listing lots of files in function `apps_in_bucket`, using `[System.IO.Directory]::EnumerateFiles()` instead of `Get-ChildItem` should be much faster ([ref](https://stackoverflow.com/questions/14970692/powershell-io-directory-find-file-types-in-all-subdirectories))
It's only run once for every bucket. Could be a problem if you have 1000 buckets. :smile:
> * Since we are deserializing a lot of JSON, using `[System.Text.Json.Utf8JsonReader]` should be faster than `ConvertFrom-Json`
Is it available in PowerShell?

@niheaven 
We need to take the filesystem cache and the drive type into account here. (In my case SSD)

libexec/scoop-search.ps1
```diff
@@ -37,19 +37,30 @@ function bin_match($manifest, $query) {
...
-            if ($_.name -match $query) { return $true }
-            $bin = bin_match (manifest $_.name $bucket) $query
+    $apps = Get-ChildItem (Find-BucketDirectory $bucket) -Filter '*.json' -Recurse
+
+    $apps | ForEach-Object {
+        $json = [System.IO.File]::ReadAllText($_.FullName) | ConvertFrom-Json -ErrorAction Continue
+
+        if($app.name -match $query) {
+            $list.Add([PSCustomObject]@{
+                Name = $_.BaseName
```
niheaven, line 47:
```suggestion
        $json = [System.IO.File]::ReadAllText($_.FullName) | ConvertFrom-Json -ErrorAction Continue
        $name = $_.BaseName

        if ($name -match $query) {
            $list.Add([PSCustomObject]@{
                Name = $name
```

libexec/scoop-search.ps1
```diff
@@ -37,19 +37,30 @@ function bin_match($manifest, $query) {
...
+                Source = $bucket
+                Binaries = ""
+            })
+        } else {
+            $bin = bin_match $json $query
             if ($bin) {
-                $_.bin = $bin
-                return $true
+                $list.Add([PSCustomObject]@{
+                    Name = $_.BaseName
```
niheaven, line 56:
```suggestion
                    Name = $name
```

niheaven (changes requested):

libexec/scoop-search.ps1
niheaven: We don't have a variable called `$app`, so all search fallback to bin search.

image

r15ch13: yep, missed that one :smile:

niheaven:
That's LGTM, and it needs a updated changelog :).

rashil2000:
> Is it available in PowerShell?

Yes

image

rashil2000:
> Sorry but I found Get-ChildItem is faster than [System.IO.Directory]::EnumerateFiles() here.

On multiple repeated runs (i.e. warm cache), EnumerateFiles is almost always twice as fast as Get-ChildItem in my testing.

niheaven:
image

https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsondocument.parse?view=net-7.0

Not available for .NET 4.5

r15ch13:
Oh, I was searching for `[System.Text.Json.Utf8JsonReader]` explicitly... :grin:

niheaven:
> On multiple repeated runs (i.e. warm cache), EnumerateFiles is almost always twice as fast as Get-ChildItem in my testing.

If we just run `Get-ChildItem` once for one bucket, that's not worthwhile to use `EnumerateFiles`. And the output of `EnumerateFiles` is `string[]` which doesn't have `FullName` or `BaseName`.

niheaven:
Out of topic here, the minimal requirements of Scoop are WindowsPowerShell and .Net Framework 4.5 which really lacks some new features of modern PowerShell and .Net.

Could this be updated? (To newest PowerShell)

rashil2000:
> * I'm not sure I understand this, but why do we use `Select-Object -ExpandProperty` instead of just directly accessing the field on lines 69, 90 and 100?

Alright, what about this one?

r15ch13:
It goes down to around 700ms from 1.3sec by using `JsonDocument.Parse()`. My C# implementation runs at around 300ms.

rashil2000:
Since `[System.Text.Json]` is unavailable in PS 5.1, there's another way: we could reuse the NewtonSoft.Json DLL. As mentioned in this [SO post](https://stackoverflow.com/a/58169326), it is also much faster than `ConvertFrom-Json`.

niheaven:
Too complicated for the JSON parsing function here... And `[System.Text.Json]` is not available for .Net Frameworks (4.5, 5 and all version before .Net Core)

The loop and generic list (this is not needed even, IMO. I've tried to replace all array '@()' to generic list but the time savings is 😭) are enough for `scoop search`, and if `ConvertFrom-Json` matters, let's create a new PR.

r15ch13:
Added a fallback if `System.Text.Json` is not available

niheaven:
I still don't think we need to speed up `ConvertFrom-Json`. From tens of seconds to seconds is important, but from seconds to milliseconds is not.

If this is needed indeed, tweak (or migrate to) the `parse_json()` function is better imo.

r15ch13:
Performance right after starting the PC still sucks.

image

niheaven:
The two different searching time are all using this PR's solution? Wow...

r15ch13:
Yes, just ran the same code twice after restarting the PC.

libexec/scoop-search.ps1
```diff
@@ -11,6 +11,11 @@ param($query)
 
 $list = [System.Collections.Generic.List[PSCustomObject]]::new()
 
+if(!$query -or $query -in @($null, '-h', '--help', '/?')) {
```
rashil2000, line 14: Uh.. this is a breaking change 😅

`scoop search` without arguments has been used to list all local manifests, facilitating commands like `scoop search | ? source -eq extras | measure`.

r15ch13: Oh, yeah for pipelines. okay reverted 😄

justrainer:
Hello. Whats the status of this? From reading the messages here, it seems to me all painpoints have been agreed upon, no? As someone coming back to Windows from Arch, I find it super painful to have to wait multiple seconds for every search I do, and I'd much appreciate any improvement to the search speed, no matter how big or small. Just wanted to add my two cents - thanks for all the hard work!

JVimes:
Really excited about this one 🙂 Is it still on the radar?

r15ch13:
Messed up the branch. Created a new PR https://github.com/ScoopInstaller/Scoop/pull/5644

PR #5325, opened 2023-01-06
5313 bug cant add a shim to an exe with spaces in its filepath

### LINKED ISSUE #5313: [Bug] Can't add a shim to an exe with spaces in it's filepath
So I'm trying to add a shim to open firefox from the command line, seems like the easiest way as I don't want to clutter up PATH, but because Windows land is very liberal with filepaths with spaces, this does not seem to be possible.

Here's the command I'm trying:
```
scoop add shim firefox "C:\Program Files\Mozilla Firefox\firefox.exe"
```
The error I receive is:
```
ERROR: Command path does not exist: C:\Program
```
I have tried this with every configuration of the path I can think of. Forward slashes, escaping backslashes, single and double quotes, no quotes with escaped spaces, quotes with escaped spaces, including escaped quotes within the quotes, using MSYS instead of powershell with all of the above and it's `/c/` paths, every attempt results in the same error. Additionally, for my own sanity, I tried reversing the arguments, in case I was failing to understand how the command worked, and that simply gave an error that `Files\Mozilla` did not exist. Seems the command is not liking a filepath with spaces.

Valinor:
#### Correct #5313 

Currently, getops parse with split space. With this correction, getops is more resilient and can parse in a more robust way with quote. Previously, 'parse this "sort of text"' => parse/this/sort/of/text
Now, 'parse this "sort of text"' => parse/this/sort of text
This correction can now handle parsing path with space (ex: c:\Program Files).

Closes #5313 

Use bug description, test on some other cases

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - **decompress:** Exclude '*.nsis' that may cause error ([#5294](https://github.com/ScoopInstaller/Scoop/issues/5294))
 - **autoupdate:** Fix file hash extraction ([#5295](https://github.com/ScoopInstaller/Scoop/issues/5295))
+- **shims parse:** Improve getops to split with quote and space ([#5313](Can't add a shim to an exe with spaces in it's filepath))
 
 ### Code Refactoring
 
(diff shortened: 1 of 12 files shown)


Valinor:
Bad branch

PR #5328, opened 2023-01-07
feat(bucket): Switch nirsoft bucket to ScoopInstaller/Nirsoft

rasa:
Per https://github.com/ScoopInstaller/Scoop/discussions/5293

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Features
 
 - **scoop-update:** Add support for parallel syncing buckets in PowerShell 7 and improve output ([#5122](https://github.com/ScoopInstaller/Scoop/issues/5122))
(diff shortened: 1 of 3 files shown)


rasa:
Should we instead create a "nirsoft2" bucket to allow users to "upgrade" to the new bucket when they're ready?

rashil2000:
> @ScoopInstaller/maintainers : Should we instead create a "nirsoft2" bucket to allow users to "upgrade" to the new bucket when they're ready?

I don't think that should be necessary.

However, I would like to stress again on the fact that we need more bucket maintainers, since more and more buckets are getting moved into the ScoopInstaller org. Should we make a community call for Nirsoft bucket, preferably before this PR is actually merged?

niheaven (approved):

PR #5346, opened 2023-01-17
feat(core): Add `-Silent` switch for `Invoke-ExternalCommand`

lewis-yeung:
### Description

In function `Invoke-ExternalCommand`, the `$Process.StartInfo.CreateNoWindow` was forcibly set to `$true` by mistake in #5066, which may cause problems in some cases. In this PR, a `-Silent` switch was introduced. The switch controls whether an app should run silently, that means, without any output for a console app or without a window for a GUI app.

### Checklist:

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -568,6 +568,9 @@ function Invoke-ExternalCommand {
         [Parameter(ParameterSetName = "UseShellExecute")]
         [Switch]
         $RunAs,
+        [Parameter(ParameterSetName = "UseShellExecute")]
+        [Switch]
+        $Quiet,
         [Alias("Msg")]
         [String]
         $Activity,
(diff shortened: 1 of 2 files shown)


lib/core.ps1
```diff
@@ -600,29 +603,33 @@ function Invoke-ExternalCommand {
...
-            $s = $_ -replace '(\\+)"', '$1$1"'
-            # escape N consecutive backslash(es), which are at the end of the string, to 2N consecutive ones
-            $s = $s -replace '(\\+)$', '$1$1'
-            # escape double quotes
-            $s = $s -replace '"', '\"'
-            # quote the argument
-            "`"$s`""
+    }
+    if ($Silent) {
+        $Process.StartInfo.UseShellExecute = $true
```
niheaven, line 608: Why `$true` here?

lib/core.ps1
```diff
@@ -600,29 +603,33 @@ function Invoke-ExternalCommand {
...
+        $Process.StartInfo.WindowStyle = [System.Diagnostics.ProcessWindowStyle]::Hidden
+    }
+    if ($ArgumentList.Length -gt 0) {
```
niheaven, line 611: Does `if ($ArgumentList)` work?

lib/core.ps1
```diff
@@ -571,6 +571,9 @@ function Invoke-ExternalCommand {
         [Parameter(ParameterSetName = "UseShellExecute")]
         [Switch]
         $RunAs,
+        [Parameter(ParameterSetName = "UseShellExecute")]
```
niheaven, line 574: This is for `UseShellExecute`, and what's normal case?

lib/core.ps1
niheaven, line 611: Does `if ($ArgumentList)` work?

lib/core.ps1
niheaven: I mean where we use `CreateNoWindow` before.

lewis-yeung: We don't need `CreateNoWindow` any more since it does not affect a GUI app. As for a console app, no **new console window** is actually created, regardless of the `CreateNoWindow` value.

lewis-yeung: The only limitation is that we cannot use `-Silent` together with `-LogPath` since the latter requires `UseShellExecute = $false`.

lib/core.ps1
lewis-yeung: Not always. When `$ArgumentList` contains only one empty string, we expect the result of `if ($ArgumentList)` to be `$true`, while it is actually `$false`. Just like both `$false -eq @('')` and `$false -eq @()` will output `True`.

lib/core.ps1
niheaven: CleanShot 2023-01-20 at 22 50 08@2x

It is using `/lwe` for `msiexec` here.

lewis-yeung: @niheaven I know it but this cannot go with the new `-Silent` switch. Any current app manifests will not be broken even if this PR is merged. Any new app manifests can just use a `/l*` flag (e.g., `/lwe`) in `msiexec`'s arguments if `-Silent` is also needed.

lewis-yeung: Shall we move on?

niheaven: One or two days after I could review it?

niheaven: 🤣 Will review now.

niheaven:
Hmm, renaming to `$Quiet`? Ref: https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/activity-parameters?view=powershell-7.3 

> Quiet
Data type: SwitchParameter
> Implement this parameter so that the cmdlet suppresses user feedback during its actions when the parameter is specified.

lewis-yeung:
@niheaven Fair point. Will do.

lewis-yeung:
> @niheaven Fair point. Will do.

Done.

niheaven (approved):

Issue #4498, opened 2021-11-06
Making it possible to store Scoop config in the root directory for increased portability.


pratikpc:
Current behaviour is to save at ~/.config/scoop/config.json.

Could we add an optional behaviour that if a user creates a config.json file in the root directory, then Scoop would prioritise that over the Scoop present in ~/.config/scoop/config.json.

If absent, ~/.config/scoop/config.json is used by default.

This would probably make Scoop a bit more portable and for most users would make it possible to copy paste Scoop from one directory to another or one system to another or one user to another without facing major issues.

hgkamath:
How about a ```SCOOP_CONFIG_DIR``` environment variable pointing to wherever ```.../scoop``` folder containing ```config.json```
and if the environment variable does not exist then fall back to the hard-coded dir.

hgkamath:
Scoop windows start-menu shortcuts may also require recreation/resyncingfor increased portability #4562

PR #5369, opened 2023-01-31
feat(config): Support portable config file

chawyehsu:
#### Description

Support portable config.json for increased portability.

#### Motivation and Context

Relates to #4498

Copy `~/.config/scoop/config.json` to or create an empty config.json in scoop root directory, it will then load the portable config file.

#### Checklist:

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -1304,7 +1304,20 @@ $configFile = "$configHome\scoop\config.json"
 $coreRoot = Split-Path $PSScriptRoot
 $pathExpected = ($coreRoot -replace '\\','/') -like '*apps/scoop/current*'
 if ($pathExpected) {
-    # Portable config is located in root directory: .\current\scoop\apps\<root>
+    # Portable config is located in root directory:
+    #    .\current\scoop\apps\<root>\config.json  <- a reversed path
+    # Imagine `<root>/apps/scoop/current/` in a reversed format,
+    # and the directory tree:
+    #
+    # ```
+    # <root>:
+    # ├─apps
+    # ├─buckets
+    # ├─cache
+    # ├─persist
+    # ├─shims
+    # ├─config.json
+    # ```
     $configPortablePath = fullpath "$coreRoot\..\..\..\config.json"
     if (Test-Path $configPortablePath) {
         $configFile = $configPortablePath
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -1298,13 +1298,17 @@ function Out-UTF8File {
 Optimize-SecurityProtocol
 
 # Load Scoop config
-$configFilename = 'config.json'
-# Portable config is located in root directory: .\lib\current\scoop\apps\<root>
-$configPortablePath = "$PSScriptRoot\..\..\..\..\$configFilename"
 $configHome = $env:XDG_CONFIG_HOME, "$env:USERPROFILE\.config" | Select-Object -First 1
-$configFile = "$configHome\scoop\$configFilename"
-if (Test-Path $configPortablePath) {
-    $configFile = $configPortablePath
+$configFile = "$configHome\scoop\config.json"
+# Check if it's the expected install path for scoop: <root>/apps/scoop/current
+$coreRoot = Split-Path $PSScriptRoot
+$pathExpected = ($coreRoot -replace '\\','/') -like '*apps/scoop/current*'
+if ($pathExpected) {
+    # Portable config is located in root directory: .\current\scoop\apps\<root>
+    $configPortablePath = fullpath "$coreRoot\..\..\..\config.json"
+    if (Test-Path $configPortablePath) {
+        $configFile = $configPortablePath
+    }
 }
 $scoopConfig = load_cfg $configFile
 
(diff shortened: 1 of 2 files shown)


lib/core.ps1
```diff
@@ -1298,8 +1298,14 @@ function Out-UTF8File {
 Optimize-SecurityProtocol
 
 # Load Scoop config
+$configFilename = 'config.json'
```
rashil2000, line 1301: Why separate variable for filename?

chawyehsu: It's just a _draft_ for now. there are problems with this implementation, it'll try to access the wrong location when Scoop is cloned to an unpredictable folder for use/test for instance.

rashil2000: Oh okay

lib/core.ps1
```diff
@@ -1300,6 +1300,16 @@ Optimize-SecurityProtocol
 # Load Scoop config
 $configHome = $env:XDG_CONFIG_HOME, "$env:USERPROFILE\.config" | Select-Object -First 1
 $configFile = "$configHome\scoop\config.json"
+# Check if it's the expected install path for scoop: <root>/apps/scoop/current
+$coreRoot = Split-Path $PSScriptRoot
```
chawyehsu, line 1304: This in fact should be identical to [`$scoopdir`](https://github.com/ScoopInstaller/Scoop/pull/5369/files#diff-cfd4446b6fd477007943865889b4cdbf84e1c089d960b82cd3b57c6bf019fce3L1336) below, unless the latter is pointed to somewhere else wrong. Maybe we could deprecate `$env:SCOOP` and `confg:rootPath` but I'm afraid that there might be something downstream depends on it, who knows.

chawyehsu: Better to not touch `$env:SCOOP` I guess. Anyways marked it ready for review.

niheaven: I use `$env:SCOOP` to mark customized path for scoop, and maybe a better way is migrate scoop under `.config` to scoop's own directory, so make scoop portable.

chawyehsu: Other than the `config.json` file, there is nothing in `~/.config/scoop/`, hence migrating the whole folder does not make sense imho.

If you mean automatical migration, I would say that's not my flavor, since not everyone needs such portability (considering the adoption of the nonportable bucket). The current implementation simply allows existing users to migrate themselves by manually creating/copying the `config.json` when they want. And it's a one-time operation. There is no mandatory migration.

A flag for initialization of portable config could be introduced to the standalone installer. Therefore, new users have a choice to enable it or not.

niheaven: I mean the `config.json` itself, and I could not find a reason that why we put config file in that dir.

Think about single-user PC, all scoop-related files should be in scoop's own dir, and that's portable.

Then multi-users PC, if scoop was installed in personal dir (`USERPROFILE`, as default), the same as single-user's. Or if scoop was installed in shared dir, why use a user specific config file for shared scoop?

Please correct me if I'm mistaken.

chawyehsu: It was introduced because of the cross-platform concern. It's been stored to `XDG_CONFIG_HOME` therefore it follows [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Though Scoop functions completely on Windows only, Scoop has been used for test purposes under *nix platforms, XDG is still a better implementation in this case I believe.

niheaven: That's okay. Now I use scoop under linux by setting env vars and forget this convenient config file.

lib/core.ps1
```diff
@@ -1300,6 +1300,16 @@ Optimize-SecurityProtocol
...
+$pathExpected = ($coreRoot -replace '\\','/') -like '*apps/scoop/current*'
+if ($pathExpected) {
+    # Portable config is located in root directory: .\current\scoop\apps\<root>
```
niheaven, line 1307:
```suggestion
    # Portable config is located in root directory: .\current\scoop\<root>
```

`$coreRoot` is the `current` dir of scoop, and three `..` go to `$scoopdir`. A typo?

chawyehsu: No, it's not.

- `$coreRoot\` - `.\current\`
- `..\..\` - `scoop\apps\`
- `..\config.json` - `<root>\config.json`.

niheaven: I mean the config file this loaded is under `scoop`, not `apps`.

niheaven: Aha, yes, this is what I mean.

niheaven (approved):

niheaven:
Ready for merge?

chawyehsu (commented):
I'm ready. See if it can have more reviews.

r15ch13 (approved):

lib/core.ps1
r15ch13: I guess we could migrate it or at least start adding the config to `<root>\config.json` via the installer for new installations.

rashil2000 (approved):

lib/core.ps1
niheaven: You are syncing scoop via different machine? I know someone is using dot file syncing for scoop's config, and machine-specific settings could be added in config file.

r15ch13: No I don't do that, but it could be a feature. It was just an idea. :smile:

PR #5391, opened 2023-02-18
fix(core): Fix `is_in_dir` under Unix

niheaven:
#### Description

Fix RegEx in `is_in_dir()` to support Unix `/`

#### Motivation and Context

- Relates to #5235 

#### Checklist:

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -632,7 +632,7 @@ function cookie_header($cookies) {
 function is_in_dir($dir, $check) {
     $check = "$(fullpath $check)"
     $dir = "$(fullpath $dir)"
-    $check -match "^$([regex]::escape("$dir"))(\\|`$)"
+    $check -match "^$([regex]::Escape("$dir"))([/\\]|$)"
 }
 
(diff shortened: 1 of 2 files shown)


lib/install.ps1
```diff
@@ -632,7 +632,7 @@ function cookie_header($cookies) {
 function is_in_dir($dir, $check) {
     $check = "$(fullpath $check)"
     $dir = "$(fullpath $dir)"
-    $check -match "^$([regex]::escape("$dir"))(\\|`$)"
+    $check -match "^$([regex]::Escape("$dir"))([/\\]|$)"
```
rashil2000, line 635: `` ` `` is missing, no?

niheaven: I found the regex works fine without `` ` `` and removed it.

rashil2000: I think we should leave it in for more clarity

niheaven: OK

rashil2000 (approved):

Issue #1813, opened 2017-11-03
Isolate scoop path to separate environment variable


masaeedu:
I have lots of other utilities that are constantly fiddling with my user PATH, and sometimes these utilities and scoop get into fights and wreck things. It'd be nice if scoop could limit all its changes to a single `$env:SCOOP_PATH` variable, which I could interpolate into my own `PATH` at whatever position (and hence precedence) I want.

excitoon:
Hi.
Please follow my PR at #2917 .

lefb766:
This would also solve an issue I have.

In the current implementation, it is tricky to get shims override commands in system dirs.
For example, zeit/now-cli wouldn't be able to override rktools' now.exe out of box.

This is due to Windows nature in which system PATH will be prepended before user PATH.
As a workaround, I ended up prepending another `$HOME\scoop\shims` to $env:Path in the $profile.

nonatorw:
I created an idea that refers to it, that I think it could be a good inclusion to the discuss: #5059,

candrapersada:
is there an option to make `\.config\scoop` portable?

nonatorw:
Guys, why this issue isn't evolving? Is there any block?

rashil2000:
No block as of now, contributions welcome :)

nonatorw:
> No block as of now, contributions welcome :)

I'll see what I can do, but Dot.Net is not my main skill. Thanks

rashil2000:
The codebase does not use .NET or C# directly, so no need to worry about that, it's just one language - PowerShell (which is internally based on .NET).

hagaigold:
> No block as of now, contributions welcome :)

@rashil2000, what will be a reasonable migration to a `SCOOP_PATH` env, if we apply this change?   
maybe with `scoop update`?


>all my variables included in the PATH are changed for their absolute paths.

This is different issue: 
`getEnvironmentVariable` can't handle env variables without expanding them.  
Actually, `install.ps1` do it [correctly](https://github.com/ScoopInstaller/Install/blob/656e17b67e8468edc5c5bd51bdad55642a03d9a4/install.ps1#L362), while [scoop ](https://github.com/ScoopInstaller/Scoop/blob/f93028001fbe5c78cc41f59e3814d2ac8e595724/lib/core.ps1#L591)not.

Should we have a different issue for this?

hagaigold:
just so this: #5395 
there was a PR to solve the %% expanding issue.

[...]

PR #5395, opened 2023-02-20
fix(env): Avoid automatic expansion of `%%` in env

### LINKED ISSUE #5394: [Bug] Environment variables are automatically expanded
## Bug Report

The environment variable of "Path" is automatically expanded due to the modification of the environment variable by Scoop.

#### Current Behavior

1. Add an item to the user's environment variable: `Foo=Bar`;
2. Add `%Foo%` to the user's "Path" environment variable;
3. Then `Bar` can be obtained in the user's "Path" by `%Foo%`;
4. Try some installation that can change the user's "Path", such as `scoop install Python`;
5. Then `%Foo%` will be directly expanded to `Bar` in "Path".

#### Expected Behavior

`%Foo%` should not be expanded after Scoop's modification of "Path"

#### Additional context

Usually environment variable string expansion will not affect anything, especially `%USERPROFILE%\AppData\Local\Microsoft\WindowsApps` is often modified by Scoop.

But there are too many items in my "Path" environment variable, and I have a problem with "PATH too long installer unable to modify Path".

So I want to use the `%%` method to nest another custom environment variable to reduce the length of "Path".

#### Possible Solution

Try the method without expanding the environment variable.(e.g. PowerShell/PowerShell#17518)

### System details

* **Windows version:** Windows 11 25300
* **OS architecture:** amd64
* **PowerShell version:** 7.3.2

#### Scoop Configuration

```json
{
  "last_update": "2023-02-20T12:28:30.9418781+08:00",
  "aria2-warning-enabled": false,
  "scoop_branch": "master"
}

```

WHYBBE:
#### Description

For example, the following command is used to obtain system environment variables.

```ps1
(Get-Item -Path "HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment").GetValue(
    "Path",
    "",
    [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames)
```

The string obtained by it will not replace the `%%`.

#### Motivation and Context

Use the Registry to get the unexpanded environment variable string inspired by PowerShell/PowerShell#17518. 
Fixes #5394
Related to #1813

#### Checklist:

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -594,12 +594,17 @@ function env($name, $global, $val='__get') {
         $RegisterKeyPath = 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager'
     }
     $RegisterKey = Get-Item -Path $RegisterKeyPath
+    $EnvRegisterKey = $RegisterKey.OpenSubKey('Environment')
     if ($val -eq '__get') {
-        $RegisterKey.OpenSubKey('Environment').GetValue($name, $null,
+        $EnvRegisterKey.GetValue($name, $null,
             [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames)
     } else {
-        $RegisterKey.OpenSubKey('Environment', $true).SetValue($name, $val,
-            [Microsoft.Win32.RegistryValueKind]::ExpandString)
+        $EnvRegisterKey = $RegisterKey.OpenSubKey('Environment', $true)
+        $RegistryValueKind = [Microsoft.Win32.RegistryValueKind]::ExpandString
+        if ($EnvRegisterKey.GetValue($name)) {
+            $RegistryValueKind = $EnvRegisterKey.GetValueKind($name)
+        }
+        $EnvRegisterKey.SetValue($name, $val, $RegistryValueKind)
     }
 }
 
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -589,15 +589,17 @@ function Invoke-ExternalCommand {
 }
 
 function env($name, $global, $val='__get') {
-    $EnvironmentRegisterKey = 'HKCU:\Environment'
+    $RegisterKeyPath = 'HKCU:'
     if ($global) {
-        $EnvironmentRegisterKey = 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment'
+        $RegisterKeyPath = 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager'
     }
+    $RegisterKey = Get-Item -Path $RegisterKeyPath
     if ($val -eq '__get') {
-        (Get-Item -Path $EnvironmentRegisterKey).
-            GetValue($name, '', [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames)
+        $RegisterKey.OpenSubKey('Environment').GetValue($name, $null,
+            [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames)
     } else {
-        Set-ItemProperty -Path $EnvironmentRegisterKey -Name $name -Value $val -Type 'ExpandString'
+        $RegisterKey.OpenSubKey('Environment', $true).SetValue($name, $val,
+            [Microsoft.Win32.RegistryValueKind]::ExpandString)
     }
 }
 
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -589,9 +589,18 @@ function Invoke-ExternalCommand {
 }
 
 function env($name,$global,$val='__get') {
-    $target = 'User'; if($global) {$target = 'Machine'}
-    if($val -eq '__get') { [environment]::getEnvironmentVariable($name,$target) }
-    else { [environment]::setEnvironmentVariable($name,$val,$target) }
+    $target = 'User'
+    $EnvironmentRegisterKey = 'HKCU:\Environment'
+    if ($global) {
+        $target = 'Machine'
+        $EnvironmentRegisterKey = 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment'
+    }
+    if ($val -eq '__get') {
+        (Get-Item -Path $EnvironmentRegisterKey).
+            GetValue($name, '', [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames)
+    } else {
+        [environment]::setEnvironmentVariable($name,$val,$target)
+    }
 }
 
 function isFileLocked([string]$path) {
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -668,17 +668,20 @@ function env($name, $global, $val = '__get') {
     } else {
         Get-Item -Path 'HKCU:'
     }
+    $EnvRegisterKey = $RegisterKey.OpenSubKey('Environment', $val -ne '__get')
 
     if ($val -eq '__get') {
-        $EnvRegisterKey = $RegisterKey.OpenSubKey('Environment')
         $RegistryValueOption = [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames
         $EnvRegisterKey.GetValue($name, $null, $RegistryValueOption)
+    } elseif ($val -eq $null) {
+        $EnvRegisterKey.DeleteValue($name)
     } else {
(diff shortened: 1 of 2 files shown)


lib/core.ps1
```diff
@@ -589,9 +589,18 @@ function Invoke-ExternalCommand {
...
+    $EnvironmentRegisterKey = 'HKCU:\Environment'
+    if ($global) {
+        $target = 'Machine'
+        $EnvironmentRegisterKey = 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment'
+    }
+    if ($val -eq '__get') {
+        (Get-Item -Path $EnvironmentRegisterKey).
+            GetValue($name, '', [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames)
+    } else {
+        [environment]::setEnvironmentVariable($name,$val,$target)
```
rashil2000, line 602: We can use `Set-Item -Path $EnvironmentRegisterKey` here as well for uniformity.

WHYBBE: At first I thought there was no difference in setting the environment variable with the previous method or with `Set-Item` etc, so it might be better not to change the code.
However when I modified the code, I found that if I used `[environment]::setEnvironmentVariable($name, $val, $target)`, the corresponding environment variable type would become `REG_SZ`. This can be seen in detail in `regedit`.

This not only defeats my original intention(avoid `%` substitutions in env), but also leads to some unusable situations.
Because the `%%` variable in the `REG_SZ` type environment variable will not be replaced with the corresponding content when using. And it needs to use the `REG_EXPAND_SZ` type to make it work.
So I used `-Type 'ExpandString'` to make it correct, refer to [Microsoft's documentation](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/set-itemproperty?view=powershell-7.2#-type).

I installed Python on another machine with my modified code and it worked fine.

Digression: When I want to use `Get-ItemProperty` to keep correspondence with `Set-ItemProperty`, I find that I can't get unexpanded environment variables through it.([an answer from stackoverflow](https://stackoverflow.com/questions/51776242/how-to-retrieve-original-unresolved-unexpanded-registry-value)). There may be other better ways, but this will work and be consistent to a certain extent.

r15ch13:
Probably fixes [#5059](https://github.com/ScoopInstaller/Scoop/discussions/5059) and with this implemented we should look into using dedicated `%SCOOP_*%` variables :smile:

lib/core.ps1
```diff
@@ -588,10 +588,19 @@ function Invoke-ExternalCommand {
...
 
-function env($name,$global,$val='__get') {
-    $target = 'User'; if($global) {$target = 'Machine'}
-    if($val -eq '__get') { [environment]::getEnvironmentVariable($name,$target) }
-    else { [environment]::setEnvironmentVariable($name,$val,$target) }
+function env($name, $global, $val='__get') {
+    $target = 'User'
+    $EnvironmentRegisterKey = 'HKCU:\Environment'
+    if ($global) {
+        $target = 'Machine'
```
r15ch13, line 595: Both `$target` variables are no longer used

WHYBBE:
> Probably fixes [#5059](https://github.com/ScoopInstaller/Scoop/discussions/5059) and with this implemented we should look into using dedicated `%SCOOP_*%` variables 😄
> 
> It should also be fixed in https://github.com/ScoopInstaller/Install as well

Thank you for pointing out the related issue for me. I haven't found it after searching for a long time, so I specially raised a new issue.

I don't know what other unrelated functions my modification will affect. Perhaps more detailed testing is needed before it can be applied to ScoopInstaller/Install or used to implement `%SCOOP_*%`.

lib/core.ps1
rashil2000: > Digression: When I want to use `Get-ItemProperty` to keep correspondence with `Set-ItemProperty`, I find that I can't get unexpanded environment variables through it.([an answer from stackoverflow](https://stackoverflow.com/questions/51776242/how-to-retrieve-original-unresolved-unexpanded-registry-value)). There may be other better ways, but this will work and be consistent to a certain extent.

Hmm, what about using [`[Microsoft.Win32.RegistryKey].GetValue()`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.win32.registrykey.getvalue?view=netframework-4.7.2) and [`[Microsoft.Win32.RegistryKey].SetValue()`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.win32.registrykey.setvalue?view=netframework-4.7.2), as the answer mentioned?

lib/core.ps1
WHYBBE: I forgot to modify it, thanks for pointing it out.

lib/core.ps1
WHYBBE: Haha I accidentally missed the link in the answer.

Calling C# code in Powershell is too difficult for me, but I still have some discoveries in my study.

`[Microsoft.Win32.Registry]::CurrentUser.OpenSubKey('Environment', $true).SetValue('Foo','Bar')` is used to set user environment variables. The system's environment variable setting method will be longer.

When using `(Get-Item -Path $EnvironmentRegisterKey).SetValue('Foo','Bar')`, it will prompt "Cannot write to the registry key.". So the object obtained in this way is read-only, and there is no way to use the `SetValue()` method.

So I combined both ways. Sample code: `(Get-Item -Path 'HKCU:').OpenSubKey('Environment', $true).SetValue('Foo', 'Bar')`

So I modified my code and it works fine now!

rashil2000: Great! I've pushed a commit to simplify variables a bit, let me know if it's okay.

WHYBBE: Of course！

rashil2000:
LGTM!

Please update the changelog.

rashil2000:
So we need this change in the Installer as well - https://github.com/ScoopInstaller/Install

Similar to `Get-Env`, we can create a function `Set-Env` and use it on https://github.com/ScoopInstaller/Install/blob/c99c2f1d53cf699d1a02af1d5ad0ec897c9b6120/install.ps1#L390

rashil2000 (approved):

WHYBBE:
> So we need this change in the Installer as well - https://github.com/ScoopInstaller/Install
> 
> Similar to `Get-Env`, we can create a function `Set-Env` and use it on https://github.com/ScoopInstaller/Install/blob/c99c2f1d53cf699d1a02af1d5ad0ec897c9b6120/install.ps1#L390

Modified in ScoopInstaller/Install#44

lib/core.ps1
```diff
@@ -662,10 +662,26 @@ function Invoke-ExternalCommand {
...
+        Get-Item -Path 'HKCU:'
+    }
+
+    if ($val -eq '__get') {
+        $EnvRegisterKey = $RegisterKey.OpenSubKey('Environment')
+        $RegistryValueOption = [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames
+        $EnvRegisterKey.GetValue($name, $null, $RegistryValueOption)
+    } else {
+        $EnvRegisterKey = $RegisterKey.OpenSubKey('Environment', $true)
+        $RegistryValueKind = if ($EnvRegisterKey.GetValue($name)) {
```
r15ch13, line 678: The String type should be chosen by the existence of `%` in the value, not based on the old value.

WHYBBE: I don't think so, maybe the user just wants to have `%%` (not expanded) in the environment variable, it should keep the original type in the registry.

It's hard to say what standard we should specify to set the type, maybe keeping it as it is is a feasible way. However, there is indeed a problem here. If the user sets the Path to `REG_SZ` in some cases, there will be problems directly adding `%SCOOP_*%` to the Path.

r15ch13: If you add a `%%` variable via System Properties then it will change it automatically to `REG_EXPAND_SZ`. We should mimic the default windows behavior in this case.

WHYBBE: This is a nice perspective. However, I think it is better to add a parameter to distinguish whether it should be changed. The detection % is too difficult to say that it will definitely meet expectations. It is up to the function caller to decide the new type, otherwise it will remain the same, but this will require modifying the caller in various places, which will involve many places.

I just want to make it simpler, that is, to change the original default behavior of `env` in the smallest scope.

WHYBBE: I modified my code, you see if it fits.

r15ch13 (dismissed):

rashil2000 (approved):

niheaven (approved):

niheaven (approved):

PR #5398, opened 2023-02-20
feat(bucket): Make official buckets higher priority

### LINKED ISSUE #5131: [Feature] Add bucket sorting (order)
Is it possible to add the use of manual bucket sorting for getting information and installing packages?

**If the same application is available in several packages, scoop will use this sorting to determine the "priority" bucket** (now alphabetical sorting is used instead). 

This is useful when using several buckets (I use about 10), and need to install some package whose name I know, but do not want to figure out which bucket it is available in. But I want to install it by priority from certain buckets.
image

Use bucket order for comands:
- info, install, search results, depends, download, cat.

### 1. Current version
```
$> scoop search scoop-completion
Name             Version Source    Binaries
----             ------- ------    --------
scoop-completion 0.2.3   dev-tools
scoop-completion 0.2.3   dodorz
scoop-completion 0.2.3   extras
scoop-completion 0.2.3   naderi

$> scoop install scoop-completion
Installing 'scoop-completion' from dev-tools bucket
```

### 2. Set order
```
Buckets:

1. main
2. extras
3. dodorz
4. naderi
5. dev-tools
```

### 3. Result
```
$> scoop search scoop-completion
Name             Version Source    Binaries
----             ------- ------    --------
scoop-completion 0.2.3   extras
scoop-completion 0.2.3   dodorz
scoop-completion 0.2.3   naderi
scoop-completion 0.2.3   dev-tools

$> scoop install scoop-completion
Installing 'scoop-completion' from extras bucket
```

## Solution suggestions

Sorting is used for all buckets, new buckets are added to the end.
We can specify a specific order in the list for each bucket

### Variant 2: Use the 'order' parameter

We can prioritize only the buckets we want - the others will be used alphabetically the way they are prioritized by default.
Each bucket will have an "order" parameter.

niheaven:
#### Description

Sort buckets and make official buckets first, with order in https://github.com/ScoopInstaller/Scoop/blob/master/buckets.json

Change the function `Get-LocalBucket()` in `buckets.ps1` that should affect all commands related to unknown bucket in app name. (i.e., `app` not `bucket/app`)

#### Motivation and Context

- Closes #5131 
- Relates to https://github.com/ScoopInstaller/Main/pull/4409
- Relates to https://github.com/ScoopInstaller/Extras/pull/10396
- Relates to https://github.com/ScoopInstaller/Versions/pull/973
- Others?

Before:
image

After:
image

#### Checklist:

### DIFF
--- a/lib/buckets.ps1
+++ b/lib/buckets.ps1
@@ -58,10 +58,18 @@ function Get-LocalBucket {
     .SYNOPSIS
         List all local buckets.
     #>
-    $bucketNames = (Get-ChildItem -Path $bucketsdir -Directory).Name
+    $bucketNames = [System.Collections.Generic.List[String]](Get-ChildItem -Path $bucketsdir -Directory).Name
     if ($null -eq $bucketNames) {
         return @() # Return a zero-length list instead of $null.
     } else {
+        $knownBuckets = known_buckets
+        for ($i = $knownBuckets.Count - 1; $i -ge 0 ; $i--) {
+            $name = $knownBuckets[$i]
+            if ($bucketNames.Contains($name)) {
+                [void]$bucketNames.Remove($name)
+                $bucketNames.Insert(0, $name)
+            }
+        }
         return $bucketNames
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Features
 
 - **scoop-update:** Add support for parallel syncing buckets in PowerShell 7 and improve output ([#5122](https://github.com/ScoopInstaller/Scoop/issues/5122))
+- **buckets:** Make official buckets higher priority ([#5398](https://github.com/ScoopInstaller/Scoop/issues/5398))
 
 ### Bug Fixes
 


lib/buckets.ps1
```diff
@@ -58,10 +58,18 @@ function Get-LocalBucket {
...
     #>
-    $bucketNames = (Get-ChildItem -Path $bucketsdir -Directory).Name
+    $bucketNames = [System.Collections.Generic.List[String]](Get-ChildItem -Path $bucketsdir -Directory).Name
     if ($null -eq $bucketNames) {
         return @() # Return a zero-length list instead of $null.
     } else {
+        $knownBuckets = known_buckets
+        for ($i = $knownBuckets.Count - 1; $i -ge 0 ; $i--) {
+            $name = $knownBuckets[$i]
+            if ($bucketNames.Contains($name)) {
```
rashil2000, line 68: This method will sort names that are same as official bucket names, but still have a different/custom URL. We should check for bucket URL as well while sorting.

niheaven: I've thought of it, but checking URL is too time-consuming and one solution is only topping `main`, `extras` and `version` buckets (or adding `non-portable`) to avoid conflict.

This solves most problems, and someone could use the official buckets' name as their own buckets at their own risks.

lewis-yeung: @niheaven Agreed. And we don't need to check URLs because someone may use a proxy/mirror URL for known buckets, e.g., `https://ghproxy.com/https://github.com/ScoopInstaller/Main` for the `main` bucket, which is what I use.

rashil2000: Hmm okay

HUMORCE:
Sorry, I was in a car crash. I will test and feedback next week.

HUMORCE (approved):

HUMORCE:
But I don't think this PR can be closed #5131.

niheaven:
Hmm, partially solved.

PR #5433, opened 2023-03-16
fix(shim): Use bash executable directly

rashil2000:
#### Description

I just realized that when installing Git through the installer, it puts the `bin` directory in PATH, which contains the bash executable. 

Likewise, when installing Git through Scoop, we create a shim for bash executable: https://github.com/ScoopInstaller/Main/blob/master/bucket/git.json#L40

#### Motivation and Context

Closes https://github.com/ScoopInstaller/Main/issues/4564

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - **env:** Avoid automatic expansion of `%%` in env ([#5395](https://github.com/ScoopInstaller/Scoop/issues/5395))
 - **install:** Fix download from private GitHub repositories ([#5361](https://github.com/ScoopInstaller/Scoop/issues/5361))
 - **scoop-info:** Fix errors in file size collection when `--verbose` ([#5352](https://github.com/ScoopInstaller/Scoop/pull/5352))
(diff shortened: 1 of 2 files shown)


niheaven:
What if no git installed?

rashil2000:
> What if no git installed?

That's precisely the advantage of this PR. Even if git is not installed, the installation of programs like Neofetch will not fail. It will succeed. Users will just see an error while calling neofetch, saying 'bash not found', which will indicate that they need to install something that provides bash (i.e. Git, MSYS2 etc.)

Currently the installation of neofetch fails if git isn't found (see linked issue).

niheaven (approved):

chawyehsu:
I'm sure after this change, shell shim calls WSL's bash.exe stub. I don't if this would be expected but it might cause unexpected behaviors.

rashil2000:
You're right, a range of Windows 10 versions does have WSL's bash executable on PATH, but it was removed by Microsoft a long time ago. If the issue arises, we can tell people what's going on...

PR #5424, opened 2023-03-10
chore(release): Bump to version 0.4.0

niheaven:
### Features

- **scoop-update:** Add support for parallel syncing buckets in PowerShell 7 and improve output (#5122)
- **bucket:** Switch nirsoft bucket to ScoopInstaller/Nirsoft (#5328)
- **bucket:** Make official buckets higher priority (#5398)
- **config:** Support portable config file (#5369)
- **core:** Add `-Quiet` switch for `Invoke-ExternalCommand` (#5346)
- **core:** Allow global install of PowerShell modules (#5611)
- **path:** Isolate Scoop apps' PATH (#5840)

### Bug Fixes

- **scoop-alias:** Prevent overwrite existing file when adding alias (#5577)
- **scoop-checkup:** Skip defender check in Windows Sandbox (#5519)
- **scoop-checkup:** Change the message level of helpers from ERROR to WARN (#5614)
- **scoop-checkup:** Don't throw 7zip error when external 7zip is used (#5703)
- **scoop-(un)hold:** Correct output the messages when manifest not found, (already|not) held (#5519)
- **scoop-info:** Fix errors in file size collection when `--verbose` (#5352)
- **scoop-reset:** Don't abort when multiple apps are passed and an app is running (#5687)
- **scoop-update:** Change error message to a better instruction (#5677)
- **scoop-virustotal:** Fix `scoop-virustotal` when `--all` has been passed without app (#5593)
- **scoop-virustotal:** Fix the issue that escape character not available in PowerShell 5.1 (#5870)
- **autoupdate:** Fix file hash extraction (#5295)
- **autoupdate:** Fix bug that 'WebClient' doesn't auto-extract 'gzip' (#5901)
- **buckets:** Avoid error messages for unexpected dir (#5549)
- **config:** Warn users about misconfigured GitHub token (#5777)
- **core:** Fix scripts' calling parameters (#5365)
- **core:** Fix `is_in_dir` under Unix (#5391)
- **core:** Rewrite config file when needed (#5439)
- **core:** Prevents leaking HTTP(S)_PROXY env vars to current sessions after Invoke-Git in parallel execution (#5436)
- **core:** Handle scoop aliases and broken(edited,copied) shim (#5551)
- **core:** Avoid error messages when deleting non-existent environment variable (#5547)
- **core:** Use relative path as fallback of `$scoopdir` (#5544)
- **core:** Fix detection of Git (#5545)
- **core:** Do not call `scoop` externally from inside the code (#5695)
[...]

PR #5452, opened 2023-04-01
fix(env): Apply env immediately by `SendMessageTimeout`

WHYBBE:
Fix the problem caused by the merge of #5395, mentioned in [5424#issuecomment-1479734238](https://github.com/ScoopInstaller/Scoop/pull/5424#issuecomment-1479734238), similar to ScoopInstaller/Install#49.

Use `SendMessageTimeout` in user32 to update the env in explorer.exe, and then the new env will be used in new process of powershell/cmd.

I have tested it in some cases and it works fine. I can't guarantee that all situations will be known to me. For example, a known issue is that if you use Windows Terminal, updating env in one tab does not update env in another tab of the same window, as mentioned in microsoft/terminal#14999.


niheaven (approved):

niheaven:
> For example, a known issue is that if you use Windows Terminal, updating env in one tab does not update env in another tab of the same window, as mentioned in https://github.com/microsoft/terminal/pull/14999.

It seems to be fixed.

PR #5484, opened 2023-04-21
refactor Update Scoop-00File.Tests.ps1

AntonOks:
Exclude some more file types / extensions


niheaven:
Why exclude these text files?

AntonOks:
> Why exclude these text files?

Why are EXE, ZIP and DLL excluded?

niheaven:
The following tests are based on text files, so exclude binary files. So, are there some special reasons to exclude `.reg|.ps1|.md`? (and `.ps1` is here which is the one that must be tested.)

PR #5545, opened 2023-06-19
fix(core): Fix detection of Git

HUMORCE:
#### Description

1. Fix function `Test-GitAvailable`: avoid outputs additional error in powershell 5.x

CAUSE: git's path are `Test-Path` twice.
https://github.com/ScoopInstaller/Scoop/blob/1d140585a4895c533a39b08e17ab615cdbfe323f/lib/core.ps1#L352
https://github.com/ScoopInstaller/Scoop/blob/1d140585a4895c533a39b08e17ab615cdbfe323f/lib/core.ps1#L371-L375

2. Fix function `Get-HelperPath`: fallback git(32bit).

CAUSE: git(32bit) is not expected before.
https://github.com/ScoopInstaller/Scoop/blob/1d140585a4895c533a39b08e17ab615cdbfe323f/lib/core.ps1#L370

#### Motivation and Context

Relates to #5122

1. `Test-GitAvailable`

Before:
```
PS C:\Users\WDAGUtilityAccount> $PSVersionTable.PSVersion.Major
5
PS C:\Users\WDAGUtilityAccount> scoop update
Test-Path : Cannot bind argument to parameter 'Path' because it is null.
At C:\Users\WDAGUtilityAccount\scoop\apps\scoop\current\lib\core.ps1:352 char:32
+     Return [Boolean](Test-Path (Get-HelperPath -Helper Git) -ErrorAct ...
+                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidData: (:) [Test-Path], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationErrorNullNotAllowed,Microsoft.PowerShell.Commands.TestPathCom
   mand

Scoop uses Git to update itself. Run 'scoop install git' and try again.
PS C:\Users\WDAGUtilityAccount> pwsh
PowerShell 7.3.4
PS C:\Users\WDAGUtilityAccount> scoop update
Scoop uses Git to update itself. Run 'scoop install git' and try again.
```

After:
```
PS C:\Users\WDAGUtilityAccount> $PSVersionTable.PSVersion.Major
5
PS C:\Users\WDAGUtilityAccount> scoop update
Scoop uses Git to update itself. Run 'scoop install git' and try again.
PS C:\Users\WDAGUtilityAccount> pwsh
PowerShell 7.3.4
PS C:\Users\WDAGUtilityAccount> scoop update
Scoop uses Git to update itself. Run 'scoop install git' and try again.
```

2. `Get-HelperPath`

Before:
```
PS C:\Users\WDAGUtilityAccount> . C:\Users\WDAGUtilityAccount\scoop\apps\scoop\current\lib\core.ps1
PS C:\Users\WDAGUtilityAccount> Get-HelperPath -Helper Git
PS C:\Users\WDAGUtilityAccount> scoop install git -a 32bit -u
...
PS C:\Users\WDAGUtilityAccount> Get-HelperPath -Helper Git
C:\Users\WDAGUtilityAccount\scoop\shims\git.exe
PS C:\Users\WDAGUtilityAccount> scoop uninstall git; scoop install git -a 64bit -u
...
PS C:\Users\WDAGUtilityAccount> Get-HelperPath -Helper Git
C:\Users\WDAGUtilityAccount\scoop\apps\git\current\mingw64\bin\git.exe
```

After:
```
PS C:\Users\WDAGUtilityAccount> . C:\Users\WDAGUtilityAccount\scoop\apps\scoop\current\lib\core.ps1
PS C:\Users\WDAGUtilityAccount> Get-HelperPath -Helper Git
PS C:\Users\WDAGUtilityAccount> scoop install git -a 32bit -u
...
PS C:\Users\WDAGUtilityAccount> Get-HelperPath -Helper Git
C:\Users\WDAGUtilityAccount\scoop\apps\git\current\mingw32\bin\git.exe
PS C:\Users\WDAGUtilityAccount> scoop uninstall git; scoop install git -a 64bit -u
...
PS C:\Users\WDAGUtilityAccount> Get-HelperPath -Helper Git
C:\Users\WDAGUtilityAccount\scoop\apps\git\current\mingw64\bin\git.exe
```

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - **scoop-info:** Fix errors in file size collection when `--verbose` ([#5352](https://github.com/ScoopInstaller/Scoop/pull/5352))
 - **shim:** Use bash executable directly ([#5433](https://github.com/ScoopInstaller/Scoop/issues/5433))
 - **scoop-checkup:** Skip defender check in Windows Sandbox ([#5519]https://github.com/ScoopInstaller/Scoop/issues/5519)
+- **core:** Fix detection of Git ([#5545]https://github.com/ScoopInstaller/Scoop/issues/5545)
 
 ### Performance Improvements
 
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -367,7 +367,7 @@ function Get-HelperPath {
     process {
         switch ($Helper) {
             'Git' {
-                $internalgit = "$(versiondir 'git' 'current')\mingw64\bin\git.exe"
+                $internalgit = (Get-AppFilePath 'git' 'mingw64\bin\git.exe'), (Get-AppFilePath 'git' 'mingw32\bin\git.exe') | Where-Object { $_ -ne $null }
                 if (Test-Path $internalgit) {
                     $HelperPath = $internalgit
                 } else {
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -368,7 +368,7 @@ function Get-HelperPath {
         switch ($Helper) {
             'Git' {
                 $internalgit = (Get-AppFilePath 'git' 'mingw64\bin\git.exe'), (Get-AppFilePath 'git' 'mingw32\bin\git.exe') | Where-Object { $_ -ne $null }
-                if ($internalgit -and (Test-Path $internalgit)) {
+                if ($internalgit) {
                     $HelperPath = $internalgit
                 } else {
                     $HelperPath = (Get-Command git -ErrorAction Ignore).Source
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -368,7 +368,7 @@ function Get-HelperPath {
         switch ($Helper) {
             'Git' {
                 $internalgit = (Get-AppFilePath 'git' 'mingw64\bin\git.exe'), (Get-AppFilePath 'git' 'mingw32\bin\git.exe') | Where-Object { $_ -ne $null }
-                if (Test-Path $internalgit) {
+                if ($internalgit -and (Test-Path $internalgit)) {
                     $HelperPath = $internalgit
                 } else {
                     $HelperPath = (Get-Command git -ErrorAction Ignore).Source
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -349,7 +349,7 @@ Function Test-CommandAvailable {
 }
 
 Function Test-GitAvailable {
-    Return [Boolean](Test-Path (Get-HelperPath -Helper Git) -ErrorAction Ignore)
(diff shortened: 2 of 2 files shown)


rashil2000:
I want to introduce another change here - in order to allow searching for paths of helpers that are installed independently by users. 

For example, if aria2c.exe or 7z.exe already exists on the user's machine, we should use that. See https://github.com/ScoopInstaller/Scoop/discussions/5592 for example.

@HUMORCE is it okay if I add this change in the `Get-AppFilePath` function in this PR itself?

HUMORCE:
`7zip` already has a config entry that allows use external installation.

For the download and extraction utils, we should impose some restrictions. this is also why I made https://github.com/ScoopInstaller/Scoop/pull/5548

rashil2000:
What would be the rationale behind not allowing people to use their own helpers? 

We already allow that for git.exe and 7z.exe. People might similarly have aria2c.exe or dark.exe already installed.

HUMORCE:
Due to the design, it has some inherent flaws from the beginning.

For stability, all helpers should be installed/invoked like `7zip19.00-helper`, IMO. otherwise, we would have to change the relevant function in the event of a breaking change to helpers.

rashil2000:
> Due to the design, it has some inherent flaws from the beginning.

I'm not sure I understand, which flaws are we talking about.

> we would have to change the relevant function in the event of a breaking change to helpers.

If we require a specific feature of the helper, i.e. a specific version (and above), we just tell users to update their helpers. We won't have to make any changes to Core.

HUMORCE:
Then why not use that version in beginning? By using specific versions of helpers, the user experience is consistent for all users.

examples:

We cannot use openssl 3.x in a program that depend on openssl 1.x

7zip 19.00 is enough to extact all manifest requires 7zip. once any breaking change in next version of 7zip(assuming it does not support 7z extraction anymore), we must refined the 7z extract function or switch to a helper like 7zip19.00-helper, because manifests of helpers are rolling updates.

rashil2000:
We aren't really using specific versions of helpers though, are we? When a helper is not found, we automatically install the latest version of it.

HUMORCE:
This is why I emphasize we need to impose restrictions for helpers. It would be better if we do dependency management, because helpers are dependecies of specific functions of Scoop Core, in fact.

The reason users need to use external installation may simply be that they don't want app install by Scoop to override their external installation, then make all helpers like 7zip19.00-helper is enough. in additional, disk space saving are not make sense.

Availability of functions is more important than extensibility.

---

We should merge this PR asap, if you still think allowing external installations are necessary, please make a separate PR.

PR #5559, opened 2023-06-24
fix(#1606): Remove console window for GUI apps, without changing the manifests.

### LINKED ISSUE #1606: Support shims to both console and GUI apps

spider2048:
#### Description

Added 2 functions to get the subsystem of the shimmed PE binaries and change the subsystem of the shim executable while installing.

#### Motivation and Context

Closes #1606 as it is annoying to deal with the Console Window.

I've checked my code changes by installing packages: nvim, vim, wireshark, firefox.
Every GUI app from the above packages (nvim-qt, gvim, gvimdiff, wireshark, firefox) doesn't pop-up a console window and works fine.
The changed code only affects the `subsystem` value in the PE `Optional Headers` of the shim exe.

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Bug Fixes
 
-- **gui-shim:** remove console window for GUI applications ([#1606](https://github.com/ScoopInstaller/Scoop/issues/1606))
+- **shim:** Remove console window for GUI applications ([#5559](https://github.com/ScoopInstaller/Scoop/issues/5559))
 - **decompress:** Exclude '*.nsis' that may cause error ([#5294](https://github.com/ScoopInstaller/Scoop/issues/5294))
 - **autoupdate:** Fix file hash extraction ([#5295](https://github.com/ScoopInstaller/Scoop/issues/5295))
 - **getopt:** Stop split arguments in `getopt()` and ensure array by explicit arguments type ([#5326](https://github.com/ScoopInstaller/Scoop/issues/5326))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Bug Fixes
 
+- **gui-shim:** remove console window for GUI applications ([#1606](https://github.com/ScoopInstaller/Scoop/issues/1606))
 - **decompress:** Exclude '*.nsis' that may cause error ([#5294](https://github.com/ScoopInstaller/Scoop/issues/5294))
 - **autoupdate:** Fix file hash extraction ([#5295](https://github.com/ScoopInstaller/Scoop/issues/5295))
 - **getopt:** Stop split arguments in `getopt()` and ensure array by explicit arguments type ([#5326](https://github.com/ScoopInstaller/Scoop/issues/5326))
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -1,9 +1,8 @@
 # Returns the subsystem of the EXE
 function Get-Subsystem($filePath) {
     try {
-        $fileStream = [System.IO.FileStream]::new($filePath, [System.IO.FileMode]::Open, [System.IO.FileAccess]::ReadWrite)
+        $fileStream = [System.IO.FileStream]::new($filePath, [System.IO.FileMode]::Open, [System.IO.FileAccess]::Read)
         $binaryReader = [System.IO.BinaryReader]::new($fileStream)
-        $binaryWriter = [System.IO.BinaryWriter]::new($fileStream)
     } catch {
         return -1  # leave the subsystem part silently
     }
@@ -896,7 +895,7 @@ function shim($path, $global, $name, $arg) {
         $target_subsystem = Get-Subsystem $resolved_path
 
         if (($target_subsystem -ne 3) -and ($target_subsystem -ge 0)) { # Subsystem -eq 3 means `Console`, -ge 0 to ignore
-            Write-Output "Changing Subsystem 3[current] to $target_subsystem[new] for $shim.exe"
+            Write-Output "Making $shim.exe a GUI binary."
             Change-Subsystem "$shim.exe" $target_subsystem
         }
     } elseif ($path -match '\.(bat|cmd)$') {
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -777,7 +777,6 @@ function shim($path, $global, $name, $arg) {
             Write-Output "args = $arg" | Out-UTF8File "$shim.shim" -Append
         }
 
-        echo "Resolved path: $resolved_path | $path"
         $target_subsystem = Get-Subsystem $resolved_path
 
         if ($target_subsystem -ne 3) { # Subsystem 3 means `Console`
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -1,3 +1,59 @@
+# Returns the subsystem of the EXE
+function Get-Subsystem($filePath) {
+    try {
+        $fileStream = [System.IO.FileStream]::new($filePath, [System.IO.FileMode]::Open, [System.IO.FileAccess]::ReadWrite)
+        $binaryReader = [System.IO.BinaryReader]::new($fileStream)
+        $binaryWriter = [System.IO.BinaryWriter]::new($fileStream)
+    } catch {
+        return -1
+    }
+
+    try {
+        $fileStream.Seek(0x3C, [System.IO.SeekOrigin]::Begin) | Out-Null
+        $peOffset = $binaryReader.ReadInt32()
+
+        $fileStream.Seek($peOffset, [System.IO.SeekOrigin]::Begin) | Out-Null
+        $fileHeaderOffset = $fileStream.Position
+
+        $fileStream.Seek(18, [System.IO.SeekOrigin]::Current) | Out-Null
+        $fileStream.Seek($fileHeaderOffset + 0x5C, [System.IO.SeekOrigin]::Begin) | Out-Null
+
+        return $binaryReader.ReadInt16()
+    } finally {
+        $binaryReader.Close()
+        $fileStream.Close()
+    }
+}
+
+function Change-Subsystem($filePath, $targetSubsystem) {
+    try {
+        $fileStream = [System.IO.FileStream]::new($filePath, [System.IO.FileMode]::Open, [System.IO.FileAccess]::ReadWrite)
+        $binaryReader = [System.IO.BinaryReader]::new($fileStream)
+        $binaryWriter = [System.IO.BinaryWriter]::new($fileStream)
+    } catch {
+        Write-Output "Error opening File:'$filePath'"
+        return
+    }
+
+    try {
+        $fileStream.Seek(0x3C, [System.IO.SeekOrigin]::Begin) | Out-Null
+        $peOffset = $binaryReader.ReadInt32()
+
+        $fileStream.Seek($peOffset, [System.IO.SeekOrigin]::Begin) | Out-Null
+        $fileHeaderOffset = $fileStream.Position
+
+        $fileStream.Seek(18, [System.IO.SeekOrigin]::Current) | Out-Null
+        $fileStream.Seek($fileHeaderOffset + 0x5C, [System.IO.SeekOrigin]::Begin) | Out-Null
+
+        $binaryWriter.Write([System.Int16] $targetSubsystem)
+
+        Write-Output "Subsystem 3[current] -> $targetSubsystem[new]"
+    } finally {
+        $binaryReader.Close()
+        $fileStream.Close()
+    }
+}
+
 function Optimize-SecurityProtocol {
     # .NET Framework 4.7+ has a default security protocol called 'SystemDefault',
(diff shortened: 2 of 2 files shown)


spider2048:
Hey, all! Umm.. any reasons for rejecting/keeping this pull request stale?

rashil2000:
Hey @spider2048! 

I apologize for letting this PR get stale - I recently switched jobs so I haven't been active on GitHub for a long while. 
This has been a very long standing issue in Scoop, so thank you for taking the time to work on it. 

I just have a couple of hunches - post which I'll merge this ASAP.

lib/core.ps1
```diff
@@ -1,3 +1,57 @@
+# Returns the subsystem of the EXE
+function Get-Subsystem($filePath) {
+    try {
+        $fileStream = [System.IO.FileStream]::new($filePath, [System.IO.FileMode]::Open, [System.IO.FileAccess]::ReadWrite)
```
rashil2000, line 4:
```suggestion
        $fileStream = [System.IO.FileStream]::new($filePath, [System.IO.FileMode]::Open, [System.IO.FileAccess]::Read)
```

This function is just reading the bytes, so we could change it to Read-only, right?

lib/core.ps1
```diff
@@ -838,6 +892,13 @@ function shim($path, $global, $name, $arg) {
         if ($arg) {
             Write-Output "args = $arg" | Out-UTF8File "$shim.shim" -Append
         }
+
+        $target_subsystem = Get-Subsystem $resolved_path
+
+        if (($target_subsystem -ne 3) -and ($target_subsystem -ge 0)) { # Subsystem -eq 3 means `Console`, -ge 0 to ignore
+            Write-Output "Changing Subsystem 3[current] to $target_subsystem[new] for $shim.exe"
```
rashil2000, line 899: Could we change this message to something more user-friendly? I'm sure most of the users won't know what a PE header or a subsystem is - they might get scared off thinking this is an error.

rashil2000 (changes requested):
Also, could you provide a brief list of validation steps to test this PR? 

Thanks again!

rashil2000:
Great! I have tested it with Notepad++ and Paint.NET. 

Final step: please add an entry in the Changelog.md

spider2048:
Hello @rashil2000, I've added an entry in the changelog.md as you requested. 
Is the PR good to go now?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Bug Fixes
 
+- **gui-shim:** remove console window for GUI applications ([#1606](https://github.com/ScoopInstaller/Scoop/issues/1606))
```
rashil2000, line 13:
```suggestion
- **shim:** Remove console window for GUI applications ([#5559](https://github.com/ScoopInstaller/Scoop/issues/5559))
```

We include the PR number here

spider2048: Oh, sorry for the mistake

spider2048 (commented):
reviewed

rashil2000 (approved):

HUMORCE:
```
~\Desktop
❯ ni 1app.exe

    Directory: C:\Users\humorce\Desktop

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---           10/9/2023  4:11 AM              0 1app.exe

~\Desktop
❯ scoop shim add 1app .\1app.exe
Adding local shim 1app...
MethodInvocationException: C:\Users\humorce\scoop\apps\scoop\current\lib\core.ps1:12
Line |
  12 |          $peOffset = $binaryReader.ReadInt32()
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception calling "ReadInt32" with "0" argument(s): "Unable to read beyond the end of the stream."
```

This is not a bug technically, but consider this scene.

spider2048:
Hello, @HUMORCE!
Thanks for pointing out the bug.
Please consider looking at #5684, I've fixed the issue along with a few minor changes. 
I think it is good to go.

r15ch13:
So, I just installed `syncthingtray` which includes cli executable on a fresh installation of Win11 and Windows Defender doesn't seem to like this implementation.

image

Maybe compiling a cli version of shim.exe and using the corresponding shim after reading the subsystem is a better approach to this issue.

spider2048:
Hello @r15ch13!
I am able to reproduce the issue on my PC as well. As you said, windows defender didn't like the application which was installed and not the shim. The following image describes it.

image

Also, if we look at the error message:
```
InvalidOperation: D:\Sources\Scoop\lib\core.ps1:19
Line |
  19 |          $binaryReader.Close()
     |          ~~~~~~~~~~~~~~~~~~~~~
     | You cannot call a method on a null-valued expression.
InvalidOperation: D:\Sources\Scoop\lib\core.ps1:20
Line |
  20 |          $fileStream.Close()
     |          ~~~~~~~~~~~~~~~~~~~
     | You cannot call a method on a null-valued expression.
```

The error occurs in the `Get-PESubsystem` function (L1-23) because windows defender stops it. (My bad for overlooking it in the `finally` block)

So, compiling the shim wouldn't really fix this issue as the user would try to run the false positive anyway.
I think the maintainers of the application should take care of this issue.

brian6932:
Another program that kinda has some weird behavior with this PR is rufus, which has a cli mode https://github.com/pbatard/rufus/issues/111#issuecomment-746313565 but it can't be accessed due to:
```
Creating shim for 'rufus'.
Making ~\scoop\shims\rufus.exe a GUI binary.
```
You can ofc get around this with `& (scoop which rufus)`

spider2048:
Hello, @brian6932!

Rufus' CLI developement has a [very low priority](https://github.com/pbatard/rufus/wiki/FAQ#do-you-plan-to-add-command-line-support). As a result, we have the following:

1. The Rufus executable is compiled with the subsystem windows.
2. To implement CLI features, the application tries to check if the calling process has a valid console handle and tries to [attach](https://github.com/pbatard/rufus/blob/58c56eb398dcc2d4168d4d4dee221197c109f32c/src/rufus.c#L3410C4-L3410C4) to it.
3. If Rufus attached to the parent processes' console, it would extract a `rufus.com` in the current folder, which Rufus calls the `Hogger`.
4. Here comes the best part. Rufus tries to simulate keystrokes `rufus.com<CR>` to the terminal (to execute it).
5. The `rufus.com` executable waits for the global mutex creted by the main `Rufus.exe` process.
6. After finishing, Rufus tries to delete the `rufus.com` file, which didn't really happen in my case.

So, when an application (like Rufus) compiled with the windows subsystem wants a console window, they use `AllocConsole` to get a new console window. But the developers already consider the disadvantage of the floating console window. But, Rufus, on the other hand tries to attach to the existing console window using `AttachConsole`. If the application was executed normally in a shell (using its actual path), Rufus can obtain a console handle directly. 

If we use the CLI version of the shim (without the PR), Rufus would think that it is operating from a shell which wouldn't be the case everytime. Also, it creates some weird lag because the main Rufus application blocks the console and the keystrokes events get processed after the application quits (creating weird lag). And because I use powershell:

```text
spider@Sp1d3R 0 0.074s
% rufus

spider@Sp1d3R 0 6.403s (command took 6 seconds even I closed the Rufus' main window within 2 seconds)
% rufus.com
rufus.com: The term 'rufus.com' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

Suggestion [3,General]: The command rufus.com was not found, but does exist in the current location. PowerShell does not load commands from the current location by default. If you trust this command, instead type: ".\rufus.com". See "get-help about_Command_Precedence" for more details.
```
... which is not a good experience for a user who needs to use Rufus' GUI.

If we wanted to use the CLI features:

```text
spider@Sp1d3R 0 0.007s
% rufus -h

Usage: rufus [-x] [-g] [-h] [-f FILESYSTEM] [-i PATH] [-l LOCALE] [-w TIMEOUT]
  -x, --extra-devs
     List extra devices, such as USB HDDs
  -g, --gui
     Start in GUI mode (disable the 'rufus.com' commandline hogger)
  -i PATH, --iso=PATH
     Select the ISO image pointed by PATH to be used on startup
  -l LOCALE, --locale=LOCALE
     Select the locale to be used on startup
  -f FILESYSTEM, --filesystem=FILESYSTEM
     Preselect the file system to be preferred when formatting
  -w TIMEOUT, --wait=TIMEOUT
     Wait TIMEOUT tens of seconds for the global application mutex to be released.
     Used when launching a newer version of Rufus from a running application.
  -h, --help
     This usage guide.

spider@Sp1d3R 0 3.667s
% rufus.com
rufus.com: The term 'rufus.com' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

Suggestion [3,General]: The command rufus.com was not found, but does exist in the current location. PowerShell does not load commands from the current location by default. If you trust this command, instead type: ".\rufus.com". See "get-help about_Command_Precedence" for more details.
```
... which is an okay-ish experience (~3.5 seconds for showing the help).

On the other hand, using the GUI version of the shim (with the PR), Rufus won't be able to attach to the console. This would not create any weird lags if the user had to use only the GUI features. We can't use the CLI features using this shim. So, if we wanted the CLI features, we would prefer to use the workaround you proposed. But the GUI features work perfectly.

```
spider@Sp1d3R 0 0.001s
% rufus

spider@Sp1d3R 0 0.052s
%
```
... which is a good experience (~0.05s to show the UAC prompt)

So, I would like to conclude that:
1. The maintainers of Rufus should rethink the way they implement the CLI features: create a console-only application/allocate a new console instead of simulating keystrokes.
2. Using this PR, Rufus' GUI features work perfectly - No weird lags, but using the CLI shim would create some lag to use both the GUI and CLI features.

brian6932:
Yea I agree that Rufus' CLI implementation is scuffed, I just wanted to point out what I witnessed, as I don't doubt there could be other GUI+CLI bins that experience similar issues. Potentially a manual setting for GUI/CLI bin shims could be required for certain programs.

spider2048:
Yes, I agree with you. I think we can probably adjust the manifests and add a command line flag, if we experience issues. (for packages like Rufus). Choosing between `AttachConsole` and `AllocConsole` is totally up to the developers.

melMass:
Is there a way to enforce this feature? I'm trying to create a manifest for the rio terminal, but the shim creates a visible terminal running the shim that stays open but at least can be closed, but I tried to make a shortcut and this creates a "conhost.exe" in the background that I have to force quit.

spider2048:
Hello @melMass!

I've tried to install `rio` using your manifest. This is what I observed after installing it:
1. Using this PR, running `rio` from the command line didn't create any "visible terminal". The only window which was created was the console window of Rio.
2. The shim doesn't start an additional `conhost.exe` process. It just starts the target process. The target process is responsible for starting a `conhost.exe` if it is using a console. 

image

The above picture is the process tree after running `rio` from the command line. The second process is the main `rio` process, which starts `conhost`.

On the other hand, if you _didn't_ use this PR, you would see a blank console window (if you run the shim using the shortcut in the startmenu) along with the Rio's main window.

image

I would also like to point out that, using the shortcut to the actual `rio.exe` (not the shim) would panic and create huge amounts of lag: (even when the PR is not used)

image

It's because the [shortcut](https://github.com/melMass/scoop-mtb/blob/4b402b867119db6f6edf5fa945a6e589b8abd059/bucket/rio.json#L39C40-L39C40) uses the variable `%HOMEPATH%` which expands to `\Users\spider` (in my case) without the drive name. It should be changed to `%HOMEDRIVE%%HOMEPATH%` which includes the drive letter as well. Now, making this fix in the shortcut didn't give me an extra `conhost.exe` process which I had to force quit.

Now, there's something I am worried about: [here](https://github.com/raphamorim/rio/blob/a4cb9ae43f8a017d44da3a0347b9db706ec35b55/rio/src/main.rs#L100), which would not work if we are using this PR, but I will make a fix for it soon.

melMass:
> It's because the [shortcut](https://github.com/melMass/scoop-mtb/blob/4b402b867119db6f6edf5fa945a6e589b8abd059/bucket/rio.json#L39C40-L39C40) uses the variable `%HOMEPATH%` which expands to `\Users\spider` (in my case) without the drive name. It should be changed to `%HOMEDRIVE%%HOMEPATH%` which includes the drive letter as well. Now, making this fix in the shortcut didn't give me an extra `conhost.exe` process which I had to force quit.
> 
> Now, there's something I am worried about: [here](https://github.com/raphamorim/rio/blob/a4cb9ae43f8a017d44da3a0347b9db706ec35b55/rio/src/main.rs#L100), which would not work if we are using this PR, but I will make a fix for it soon.

Thanks a lot for this answer, I will try out more based on it

PR #5572, opened 2023-07-09
Add support for resetting of apps to be imported

### LINKED ISSUE #5525: [Feature] Scoop import shall reset installed apps
### Problem
I use scoopfiles a lot in my daily environment to define a working and tested build environment in my repositories. But when my repositories use different versions of the same package (e.g., mingw-llvm or python) builds might still fail because an import does not reset the packages (which are all already installed). E.g., when a repo needs python 3.11 when python 3.10 and python 3.11 are already installed and python 3.10 was installed after 3.11 I need to reset 3.11 manually after the import.

#### Solution
When scoop import is intended to produce a reproducible environment it should reset the packages, too.

What do the scoop experts think about this? I would be happy to implement this by the way.

xxthunder:
#### Description
Added a command line option for scoop import:
* -r, --reset: reset each app of imported JSON after installation

#### Motivation and Context
I use the import command quite often to define apps and their versions as dependencies in my projects. The reason for this is of course reproducibility of builds. But when different versions from different buckets of an app are already installed, it might still happen that the wrong apps and/or versions are used.
Closes #5525 

As there are no unit tests available for the changes file (actually there are no functions in scoop-import.ps1) I tested the changes exploratorily with my local scoopfile.json's.
My changes do affect the scoop import feature only.

#### Checklist:

### DIFF
--- a/libexec/scoop-import.ps1
+++ b/libexec/scoop-import.ps1
@@ -1,16 +1,18 @@
-# Usage: scoop import <path/url to scoopfile.json>
+# Usage: scoop import <path/url to scoopfile.json> [options]
 # Summary: Imports apps, buckets and configs from a Scoopfile in JSON format
 # Help: To replicate a Scoop installation from a file stored on Desktop, run
 #      scoop import Desktop\scoopfile.json
+#
+# Options:
+#   -r, --reset                     Reset the app after installation
 
-param(
-    [Parameter(Mandatory)]
-    [String]
-    $scoopfile
-)
-
+. "$PSScriptRoot\..\lib\getopt.ps1"
 . "$PSScriptRoot\..\lib\manifest.ps1"
 
+$opt, $scoopfile, $err = getopt $args 'r' 'reset'
+if ($err) { "scoop import: $err"; exit 1 }
+
+$reset = $opt.r -or $opt.reset
 $import = $null
 $bucket_names = @()
 $def_arch = Get-DefaultArchitecture
@@ -58,6 +60,10 @@ foreach ($item in $import.apps) {
     }
 
     & "$PSScriptRoot\scoop-install.ps1" $app @instArgs
+
+    if ($reset) {
+        & "$PSScriptRoot\scoop-reset.ps1" $app
+    }
 
     if ('Held package' -in $info) {
         & "$PSScriptRoot\scoop-hold.ps1" $item.Name @holdArgs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Features
 
+- **scoop-import:** Add support for resetting of apps to be imported ([#5525](https://github.com/ScoopInstaller/Scoop/issues/5525))
 - **scoop-update:** Add support for parallel syncing buckets in PowerShell 7 and improve output ([#5122](https://github.com/ScoopInstaller/Scoop/issues/5122))
 - **bucket:** Switch nirsoft bucket to ScoopInstaller/Nirsoft ([#5328](https://github.com/ScoopInstaller/Scoop/issues/5328))
 - **config:** Support portable config file ([#5369](https://github.com/ScoopInstaller/Scoop/issues/5369))


libexec/scoop-import.ps1
```diff
@@ -59,6 +63,14 @@ foreach ($item in $import.apps) {
 
     & "$PSScriptRoot\scoop-install.ps1" $app @instArgs
 
+    if ($update) {
+        & "$PSScriptRoot\scoop-update.ps1" $app$global
```
HUMORCE, line 67: Can this passing args correctly?

libexec/scoop-import.ps1
```diff
@@ -59,6 +63,14 @@ foreach ($item in $import.apps) {
...
+    }
+
+    if ($reset) {
+        & "$PSScriptRoot\scoop-reset.ps1" $app
+    }
```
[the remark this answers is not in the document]
xxthunder: Maybe it should not, but it is. As I described in the issue https://github.com/ScoopInstaller/Scoop/issues/5525 if install different versions of the same app, no install or update command gives you what you want.

Example: I have the apps python310 and python311 out of the versions bucket installed. python311 was the last one and so the shim python.exe points to that of python311. If I now import a scoopfile.json refering to python310, nothing is changed during import. I end up with the wrong python.exe in the PATH. Therefore the reset argument.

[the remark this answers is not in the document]
HUMORCE: This happens after both `python310` and `python311` have been installed already, then the import operation is performed right?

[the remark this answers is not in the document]
xxthunder: Exactly.

libexec/scoop-import.ps1
xxthunder: No, $global does not exist. You are right. Fix is on the way.

libexec/scoop-import.ps1
```diff
@@ -1,16 +1,20 @@
...
-param(
-    [Parameter(Mandatory)]
-    [String]
-    $scoopfile
-)
-
+. "$PSScriptRoot\..\lib\getopt.ps1"
 . "$PSScriptRoot\..\lib\manifest.ps1"
 
+$opt, $scoopfile, $err = getopt $args 'ru:' 'reset', 'update'
```
rashil2000, line 13:
```suggestion
$opt, $scoopfile, $err = getopt $args 'ru' 'reset', 'update'
```

rashil2000:
@xxthunder I don't think the update flag is required at all. There are basically three types of apps in the imported scoopfile:

```pwsh
    $app = if ($item.Source -in $bucket_names) {
        "$($item.Source)/$($item.Name)"
    } elseif ($item.Source -eq '<auto-generated>') {
        "$($item.Name)@$($item.Version)"
    } else {
        $item.Source
    }
```

1. The first case corresponds to apps of the form `<bucket>/<app>`. When this case is present, always the latest version of it will be installed.
2. The second case corresponds to a fixed/pinned version. Updates should not be carried out in this case.
3. The third case is similar to the second - it points either to a remote URL or a local filepath, where the manifest has a fixed version.

Basically, only those apps, which belong to a particular bucket, are eligible for updates. And this is covered in the (1) scenario (because buckets are always up-to-date, as Scoop follows the rolling release model of updates).

xxthunder:
@rashil2000 you are right, removed the update flag again. Added two small fixes in import and install. Please have a look when you have time.

libexec/scoop-import.ps1
```diff
@@ -50,7 +52,11 @@ foreach ($item in $import.apps) {
     }
 
     $app = if ($item.Source -in $bucket_names) {
-        "$($item.Source)/$($item.Name)"
+        if ($item.Version) {
```
rashil2000, line 55: This check is redundant. `Version` is always present in the $item object.

libexec/scoop-import.ps1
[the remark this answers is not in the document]
rashil2000: @xxthunder technically, this is also not required. To see what I mean:  

https://github.com/ScoopInstaller/Scoop/blob/6cdcc75ad84add9deedad81f1c9c16dc43f868c0/lib/install.ps1#L58-L66 

The functions that `scoop-reset` will run (see below) have already been run by `scoop-install` (see above). 

https://github.com/ScoopInstaller/Scoop/blob/6cdcc75ad84add9deedad81f1c9c16dc43f868c0/libexec/scoop-reset.ps1#L80-L87

libexec/scoop-import.ps1
xxthunder: You are right, the version is inside the $item object. But I had the problem, that the version was ignored, when I imported a scoopfile with a version given in one of the apps. Maybe my expectation is wrong, but I want to install specific versions with a scoopfile in a project/repo. Therefore I wanted to give that version to scoop-install.ps1 some lines below this.

rashil2000: If you specify a version in the `scoop-install` command, that version will get locked - and you will not be able to update the application. 

Specifying versions auto-generates the manifest, which is similar to case (2) in my above comment.

xxthunder: Mmmmh, so I guess, my use case to install specific versions of a package with "scoop import" is not a valid one. Without this change it will not work.

rashil2000: Correct. If you want to install specific versions, the machine you're exporting from should have those specific/pinned versions (i.e. the `<auto-generated>` source) in the first place.

rashil2000:
Check my comment on HUMORCE's thread.

libexec/scoop-import.ps1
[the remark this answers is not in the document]
xxthunder: @rashil2000 @HUMORCE I think this is not true. I want the reset no matter what is already installed. So if I import with, e.g., cmake version 3.24 in my scoopfile.json and cmake 3.24 and 3.26 both already installed and p3.26 was installed the last time. Then the shim cmake points to cmake 3.26. Now I can import my scoopfile as often as I want, nothing will be installed because 3.24 is already installed and nothing will be resetted. So the shim cmake points to cmake 3.26. Exactly this my problem with scoop import. My option --reset solves this.

PR #5577, opened 2023-07-11
fix(scoop-alias): Prevent overwrite existing file when adding alias

HUMORCE:
#### Motivation and Context

This feature exists in the tests, but is not implemented.

before:
```
~\scoop
❯ ni .\shims\scoop-xxx.ps1 | Out-Null
~\scoop
❯ scoop alias add xxx 'new alias'
~\scoop
❯ cat .\shims\scoop-xxx.ps1
# Summary:
new alias
~\scoop
❯
```

after:
```
~\scoop
❯ ni .\shims\scoop-xxx.ps1 | Out-Null
~\scoop
❯ scoop alias add xxx 'new alias'
File 'scoop-xxx.ps1' already exists in shims directory.
~\scoop
❯ cat .\shims\scoop-xxx.ps1
~\scoop
❯
```

`bin\test.ps1` in powershell 5 and 7, all passed

#### Checklist:


test/Scoop-Alias.Tests.ps1
```diff
@@ -23,19 +23,18 @@ Describe 'Manipulate Alias' -Tag 'Scoop' {
...
-        add_alias 'rm' 'test'
-        & $alias_file | Should -Not -Be 'test'
-    }
+    #     add_alias 'rm' '"test"'
+    #     & $alias_file | Should -Be $null
+    # }
 
     It 'Removes an existing alias' {
         $alias_file = "$shimdir\scoop-rm.ps1"
-        add_alias 'rm' '"hello, world!"'
```
HUMORCE, line 37: emmm, the alias and its file already been created in pre-phase.

test/Scoop-Alias.Tests.ps1
```diff
@@ -23,19 +23,18 @@ Describe 'Manipulate Alias' -Tag 'Scoop' {
...
-        $alias_file | Should -Exist
+    # NOTE: `abort()` in `add_alias()` cause tests error.
+    # It 'Does not change existing file if its filename same as alias name' {
+    #     $alias_file = "$shimdir\scoop-rm.ps1"
+    #     New-Item $alias_file -Type File -Force
+    #     $alias_file | Should -Exist
 
-        add_alias 'rm' 'test'
-        & $alias_file | Should -Not -Be 'test'
-    }
```
HUMORCE, line 33: it works when replace `abort()` with `error(); return` in `add_alias()`.

libexec/scoop-alias.ps1
```diff
@@ -44,13 +44,16 @@ function add_alias($name, $command) {
...
-        abort "Alias $name already exists."
+        abort "Alias '$name' already exists."
     }
 
     $alias_file = "scoop-$name"
 
     # generate script
     $shimdir = shimdir $false
+    if (Test-Path "$shimdir\$alias_file.ps1") {
+        abort "File '$alias_file.ps1' already exists in shims directory."
```
rashil2000, line 55: Is there a case where this file would exist but `$aliases.$name` wouldn't? (unless of course we manually create it).

Because we're already checking for `$aliases.$name` above.

HUMORCE: > Is there a case where this file would exist but $aliases.$name wouldn't?

That's why this PR has been created. 

```
> touch ~\scoop\shims\scoop-random.ps1
> scoop help
Command    Summary
-------    -------
alias      Manage scoop aliases
...
random
```

or maybe the alias entries does not need to be stored in the config

rashil2000: No but here you are creating this file manually, with a specific naming scheme. 

I don't think users will do that on their own (there's no reason to manually create that file). 

They will always use the `scoop alias <...>` command to do that.

HUMORCE: When migration.

Copied Scoop dir from another PC without config file.

rashil2000: Oh right

rashil2000 (dismissed):

test/Scoop-Alias.Tests.ps1
niheaven: Using `Should -Invoke`

niheaven:
image

niheaven (dismissed):

PR #5684, opened 2023-10-09
fix(shim): fix a minor issue with `Get-Subsystem` (and refactor)

spider2048:
#### Description
Fixes minor bugs

#### Motivation and Context
The `Get-Subsystem` function doesn't catch errors when it fails to read a binary.
The `Get-Subsystem` and `Change-Subsystem` functions are refactored to `Get-PESubsystem` and `Change-PESubsystem`.
The code will only try to change subsystem to 2 (which indicates [GUI](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#windows-subsystem))

Relates to #5559 

Credits to @HUMORCE [(original)](https://github.com/ScoopInstaller/Scoop/pull/5559#issuecomment-1752322307).

Steps to reproduce the bug:
```powershell
% touch empty.exe
% scoop shim add empty .\empty.exe
Adding local shim empty...
MethodInvocationException: D:\Sources\Scoop\lib\core.ps1:12
Line |
  12 |          $peOffset = $binaryReader.ReadInt32()
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception calling "ReadInt32" with "0" argument(s): "Unable to read beyond the end of the stream."
```

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - **scoop-virustotal:** Fix `scoop-virustotal` when `--all` has been passed without app ([#5593](https://github.com/ScoopInstaller/Scoop/pull/5593))
 - **scoop-checkup:** Change the message level of helpers from ERROR to WARN ([#5549](https://github.com/ScoopInstaller/Scoop/issues/5614))
 - **scoop-(un)hold:** Correct output the messages when manifest not found, (already|not) held ([#5519](https://github.com/ScoopInstaller/Scoop/issues/5519))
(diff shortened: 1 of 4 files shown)


rashil2000:
There should be a `catch` block in the `Change-PESubsystem` function too, right, just in case?

r15ch13:
And use a approved PowerShell verb by changing `Change-PESubsystem` to `Set-PESubsystem`

spider2048:
I've refactored `Change-PESubsystem` to `Set-PESubsystem` and added an additional `catch` block as requested.
I think it is good to go.

niheaven (approved):

PR #5685, opened 2023-10-09
fix `scoop uninstall scoop` (#5683)

### LINKED ISSUE #5683: [Bug] `scoop uninstall scoop` is broken
#### Current Behavior
`scoop uninstall scoop` is broken since commit 285ffd0
Each specific app uninstall is failing.

#### Expected Behavior
Uninstall of scoop & all apps.

#### Possible Solution
In `uninstall.ps1`, `unlink_current` is failing because of an invalid directory:
```
function do_uninstall($app, $global) {
    $version = Select-CurrentVersion -AppName $app -Global:$global
    $dir = versiondir $app $version $global
    ...
    ...
    # If a junction was used during install, that will have been used
    # as the reference directory. Othewise it will just be the version
    # directory.
    $refdir = unlink_current (appdir $app $global)   <<<<=============
```

While the fix is relatively easy, I wonder if the uninstall app procedure in, `do_uninstall` in `uninstall.ps1`, should be use at all.  
There is a different procedure in [`scoop-uninstall.ps1`](https://github.com/ScoopInstaller/Scoop/blob/3a3f41c556c044150d86f259c1e3583687b41929/libexec/scoop-uninstall.ps1#L52), that is use when uninstall is called for single app- 'scoop uninstall me'.  
On top of that, there is uninstall procedure in `scoop-update.ps1`.

I see at least 2 disparities between `uninstall.ps1` and `scoop-uninstall.ps1`:
- `Invoke-HookScript -HookType 'pre_uninstall' -Manifest $manifest -Arch $architecture`
- `rm_startmenu_shortcuts $manifest $global $architecture`

hagaigold:
#### Description
`scoop uninstall scoop` is broken

#### Motivation and Context
Closes #5683 

This is a "lazy" fixing of this issue by syncing the uninstall procedure commands from [scoop-uninstall.ps1](https://github.com/ScoopInstaller/Scoop/blob/3a3f41c556c044150d86f259c1e3583687b41929/libexec/scoop-uninstall.ps1#L52) to `uninstall.ps1`.  
Please See the issue for more context, but a more robust way, is to refactor to code for a single function of uninstalling an app.

Install 2 apps (one global) and then `scoop uninstall scoop` (see attached log)

#### Checklist:

[test-5683.txt](https://github.com/ScoopInstaller/Scoop/files/12849878/test-5683.txt)

### DIFF
--- a/bin/uninstall.ps1
+++ b/bin/uninstall.ps1
@@ -41,13 +41,19 @@ function do_uninstall($app, $global) {
     $architecture = $install.architecture
 
     Write-Output "Uninstalling '$app'"
+
+    Invoke-HookScript -HookType 'pre_uninstall' -Manifest $manifest -Arch $architecture
+
     run_uninstaller $manifest $architecture $dir
     rm_shims $app $manifest $global $architecture
+    rm_startmenu_shortcuts $manifest $global $architecture
 
     # If a junction was used during install, that will have been used
     # as the reference directory. Othewise it will just be the version
     # directory.
-    $refdir = unlink_current (appdir $app $global)
+    $refdir = unlink_current $dir
+
+    uninstall_psmodule $manifest $refdir $global
 
     env_rm_path $manifest $refdir $global $architecture
     env_rm $manifest $global $architecture
@@ -59,6 +65,8 @@ function do_uninstall($app, $global) {
         $errors = $true
         warn "Couldn't remove $(friendly_path $appdir): $_.Exception"
     }
+
+    Invoke-HookScript -HookType 'post_uninstall' -Manifest $manifest -Arch $architecture
 }
 
 function rm_dir($dir) {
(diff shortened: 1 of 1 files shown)


hagaigold:
another test - `scoop uninstall scoop -g`

[test-5683-uninstall-global.txt](https://github.com/ScoopInstaller/Scoop/files/12850008/test-5683-uninstall-global.txt)

rashil2000:
Do you think it would be better to invoke `scoop-uninstall` here, instead of invoking all the same functions? This will remove redundancy. 

We can just invoke `& "$PSScriptRoot\scoop-uninstall.ps1" $apps` and pass in the array of all the apps, global and local separately.

hagaigold:
> instead of invoking all the same functions?  

I agree this is more robust way of doing things.    
Because of my limited knowledge of this code base, the intention of this PR was **just** to fix the issue without too much effort.  

> Do you think it would be better to invoke `scoop-uninstall`  

if you think this approach making sense (from scoop architecture perspective), I can give it another look.

rashil2000:
> if you think this approach making sense (from scoop architecture perspective), I can give it another look.

Yes, please go ahead!

PR #5721, opened 2023-11-09
fix(shim): allow GUI applications to attach to the shell's console when launched using the GUI shim

spider2048:
#### Description
Allows patched GUI shims to access the shell's console window.

#### Motivation and Context

Fixes issues described in #5559 (for the `scoopcs` shim type)

#### How Has This Been Tested?

The new shim code only tries to `AttachConsole` to the parent process, so that it gets the console handles (when it doesn't find a console window). Now, the target process which the shim starts can also request for the same console handles and it can write to the shell's terminal.

Validation steps:
Let's use `alacritty`.
We know that, `alacritty.exe` is a GUI application, but we have:

```powershell
$ &(scoop which alacritty) -V
alacritty 0.12.3 (5efb069)
````

If we don't use this PR, we would not have this output, because the target process (alacritty) fails to get handles to the shell's console.

```powershell
$ alacritty -h

```

But, if we use this PR:
```powershell
$ scoop config shim scoopcs
$ scoop update -f alacritty
$ alacritty -V
alacritty 0.12.3 (5efb069)
```

This PR would restore the behaviour of GUI+CLI applications and avoids the need to implement new flags or adjust the manifests.

#### Checklist:

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).
- [x] I have ensured that I am targeting the `develop` branch.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.
- [ ] I have added an entry in the CHANGELOG.

PR #5728, opened 2023-11-14
Avoid shim blocking when running a GUI app 

milnak:
#### Description
Currently scoop shim will block when running a GUI app.  

To repro this, do the following from a  powershell console:

```PowerShell
scoop install extras/notepad2-zufuliu
notepad2
```

The console will be blocked until notepad exits.

I have added an environment variable that can be set to override this behavior.

#### Motivation and Context
Relates to #1606 

Build shim, verified against notepad2 and console apps.

#### Checklist:
- [X] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/supporting/shimexe/shim.cs
+++ b/supporting/shimexe/shim.cs
@@ -9,10 +9,12 @@ using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Runtime.InteropServices;
 
-namespace Scoop {
+namespace Scoop
+{
 
-    class Program {
-        [DllImport("kernel32.dll", SetLastError=true, CharSet=CharSet.Unicode)]
+    class Program
+    {
+        [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
         static extern bool CreateProcess(string lpApplicationName,
             string lpCommandLine, IntPtr lpProcessAttributes,
             IntPtr lpThreadAttributes, bool bInheritHandles,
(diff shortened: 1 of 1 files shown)


milnak:
looks like visual studio did some reformatting of the file. If you want me to undo those reformats, let me know and ill  push a new iter.

spider2048:
Using the develop branch (which had #5559 and #5684 merged recently) would solve the issue for every shim type.
So, this is implemented in the develop branch.

But, when a target application uses `AttachConsole` to obtain console handles, both the current develop branch and your's won't be able to. For instance,
```
$ scoop config shim scoopcs
$ scoop install alacritty
$ alacritty -h
<blank>
```

There is #5721 waiting for approval which would solve this issue.
Because of [this](https://github.com/ScoopInstaller/Scoop/blob/8c4988e1a36e592529ce3f355b7950bdd8e2c471/supporting/shimexe/shim.cs#L139C1-L150C10), the process can't obtain the handles to console just by using `AttachConsole`, because the shim exits.

milnak:
@spider2048 it looks like your changes handle this case and more, so this PR can likely be closed.

amorphobia:
Since the fix is not released yet (will be in 0.4.0), I provide a workaround for current version (0.3.1).

Save following as `np2.ps1` in your `$env:Path`:
```PowerShell
if ($args.Length -lt 1) {
    Start-Process -WindowStyle hidden notepad2.exe
} else {
    Start-Process -WindowStyle hidden notepad2.exe ($args -join ' ')
}
```

Just use `np2 [file_path]` from command line.

PR #5799, opened 2024-02-06
perf(scoop-update): Check for running process before wasting time on download

davidlwatsonjr:
This PR simply moves the check for an existing running process to BEFORE the download of the new version. This saves time especially on larger application downloads and slower connections.

#### Description

As stated, I've simply moved the `test_running_process` call to before the download of the application.

#### Motivation and Context

This change was motivated by an instance where I attempted to `scoop update *` and some of the applications were still running. To speed this up, it seemed reasonable to skip the update process for running applications. If and until it's possible to reliably update a running application, this seems a reasonable stop-gap to improve update performance of applications that can be updated.

This relates tangentially to https://github.com/ScoopInstaller/Scoop/issues/1372 as the update process will at least avoid the pointless download time if the application is running.

I ran this locally using `scoop update keepass` while keepass was running. After my change, everything ran as expected and the running process information was shown, bypassing the download step.

#### Checklist:


chawyehsu (approved):

libexec/scoop-update.ps1
```diff
@@ -241,6 +241,12 @@ function update($app, $global, $quiet = $false, $independent, $suggested, $use_c
 
     Write-Host "Updating '$app' ($old_version -> $version)"
 
+    #region Workaround for #2952
+    if (test_running_process $app $global) {
+        return
```
niheaven, line 246:
```suggestion
        Write-Host 'Running process detected, skip updating.
        return
```

niheaven (dismissed):

niheaven (approved):

SkeathyTomas:
I use proxy to update scoop apps. This feather annoys when I update my proxy app for when I quite the app first and I cannot download anything. Previous version downloading first strategy won't waste time on download as it doesn't delete packages. Can it be a config or scroll back?

chawyehsu:
There is not much difference in the case of updating your app - You have to close the app eventually to get it updated successfully. You may use `scoop download`. #5908

PR #5813, opened 2024-02-27
feat: Changed ENV variable name to allow for custom paths

### LINKED ISSUE #1813: Isolate scoop path to separate environment variable

CEbbinghaus:
#### Description

This PR adds the ability to define the `SCOOP_ENV` (Subject to change) environment variable which changes which environment variable scoop writes paths to. In my case I have all the scoop paths under `SCOOP_PATH` which can get as clutters as it would like. while my `PATH` just contains the reference to the scoop paths by containing `%SCOOP_PATH%`.

#### Motivation and Context

Fixes #1813

I was unable to edit my PATH environment variable through the windows dialogue as scoop had added more than a dozen different paths. The recommended answer was to create a different environment variable to contain all the paths & just link to it via %ENV_NAME%. I realized that Scoop did not have the ability to change the environment variable that the paths got added to so I just added it myself

It has been running as my local scoop version for the last few weeks & has worked. I will probably add a Unit test if this is something the Scoop team considered

#### Checklist:

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -740,10 +740,11 @@ public static extern IntPtr SendMessageTimeout(
 
 function env($name, $global, $val = '__get') {
     if(-not $name) {
-        $name = [environment]::getEnvironmentVariable("SCOOP_ENV", "User")
-        if(-not $name) {
-            $name = 'PATH'
-        }
+        $name = $scoop_path_env
+    }
+
+    if(-not $name) {
+        throw "Unable to evaluate path environment variable. Please set or remove PATH_ENV in your config.json."
     }
 
     $RegisterKey = if ($global) {
@@ -755,7 +756,7 @@ function env($name, $global, $val = '__get') {
 
     if ($val -eq '__get') {
         $RegistryValueOption = [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames
-        $EnvRegisterKey.GetValue($name, $null, $RegistryValueOption)
+        $EnvRegisterKey.GetValue($name, $null, $RegistryValueOption) + ""
     } elseif ($val -eq $null) {
         try { $EnvRegisterKey.DeleteValue($name) } catch { }
         Publish-Env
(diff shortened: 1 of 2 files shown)


CEbbinghaus:
Rebased onto `develop`. It looks like there is an issue which I was not previously aware of.
Personally I would prefer contain the environment variable name within the config but I wasn't sure how to achieve that (this was a bit of a hack job to just get what I needed working). As such I welcome any feedback on how I can improve the code (I especially hate fetching the environment variable every time but was hoping to replace it with the config).

CEbbinghaus:
Resolved the Merge conflicts

lib/core.ps1
```diff
@@ -748,7 +756,7 @@ function env($name, $global, $val = '__get') {
 
     if ($val -eq '__get') {
         $RegistryValueOption = [Microsoft.Win32.RegistryValueOptions]::DoNotExpandEnvironmentNames
-        $EnvRegisterKey.GetValue($name, $null, $RegistryValueOption)
+        $EnvRegisterKey.GetValue($name, $null, $RegistryValueOption) + ""
```
niheaven, line 759: What's this `""` for?

lib/core.ps1
```diff
@@ -1415,6 +1423,8 @@ if ($pathExpected) {
 }
 $scoopConfig = load_cfg $configFile
 
+$scoop_path_env = get_config PATH_ENV 'PATH'
```
niheaven, line 1426: Recommend using a config option to check if use isolated path.

niheaven (commented):
I'll submit a reviewed commit soon.

lib/core.ps1
CEbbinghaus: If the environment variable is empty we get a null exception later on. so we add the empty string to ensure it always returns a string even if the environment variable is empty.

lib/core.ps1
CEbbinghaus: I assumed we would want the default behavior to revert to the current behavior of just adding the paths to `PATH`. So we should have two different config options? A bool to enable custom paths and one to contain the path itself? I feel it makes more sense to just have a single config value that contains the Environment variable scoop should write to. That way we don't need extra checks and the user is always aware which environment variable is being written to.

rashil2000:
Closing this in favour of #5840

PR #5836, opened 2024-03-20
refactor(core): Rewrite and separate path-related functions to `system.ps1`

niheaven:
#### Description

Ed. Separate part of isolate PATH to #5840, this PR is for refactoring now.

~Use config option `use_isolated_path` to separate Scoop apps PATH to `SCOOP_PATH`, and just add `%SCOOP_PATH%` to `$env:Path` when install app with `env_add_path` first time.~

#### Motivation and Context

- Prepare for #5840

What the commits do:

- refactor(core): Remove unused `ensure_robocopy_in_path()`
  - The function is used when installing scoop itself, and the install script is moved to https://github.com/ScoopInstaller/Install/blob/master/install.ps1
- refactor(system): Move EnvVar-related functions to `system.ps1`
  - Move from `core.ps1`: `Publish-Env()`, `env()`, `search_in_path()`, `ensure_in_path()`, `strip_path()`, `add_first_in_path()`, `remove_from_path()`
  - Add necessary `system.ps1` importing
- refactor(EnvVar): Separate `env()` to `Get-EnvVar()` and `Set-EnvVar()`
  - Copied and modified from https://github.com/ScoopInstaller/Install/blob/master/install.ps1
- refactor(system): Remove `search_in_path()`
  - Replace with `Get-Command().Source`
- refactor(system): Refactor PATH-related functions
  - `strip_path()` -> `Find-Path()`
  - `ensure_in_path()`/`add_first_in_path()` -> `Add-Path()`
    - `ensure_in_path()` is only used in shim creation
    - Use `-Force` to forcedly add a path in the first position as `add_first_in_path()`
    - `Add-Path()` supports `%`
  - `remove_from_path()` -> `Remove-Path()`

#### How Has This Been Tested?

Passed all Scoop tests.

#### Checklist:

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).
- [x] I have ensured that I am targeting the `develop` branch.
- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.
- [x] I have added an entry in the CHANGELOG.

PR #5840, opened 2024-03-22
feat(path): Isolate Scoop apps' PATH

### LINKED ISSUE #1813: Isolate scoop path to separate environment variable

niheaven:
#### Description

Separate from #5836, will rebase on `develop` when #5836 is merged.

Use the `use_isolated_path` config option to separate Scoop apps PATH to `SCOOP_PATH` (or any other name), then add `%SCOOP_PATH%` to `$env:Path` when installing apps with `env_add_path` for the first time.

#### Motivation and Context

- Closes #1813

- Relates to #5813

This PR does the following:
- Use `-TargetEnvVar` in `Add-Path()` and `Remove-Path()` to add/remove path in different EnvVars
- Check `use_isolated_path` in `env_add_path()` and `env_rm_path()` to add/remove apps path in `%SCOOP_PATH%` or `%PATH%`
- Add `%SCOOP_PATH%` when first add path to it

Path switching is done with `Complete-ConfigChange()`.

Passed scoop test.
image

`%SCOOP_PATH%` is added correctly:
image
image

#### Checklist:


lib/system.ps1
```diff
@@ -88,6 +88,7 @@ function Test-PathLikeEnvVar {
 function Add-Path {
     param(
         [string]$Path,
+        [string]$Scope = 'PATH',
```
chawyehsu, line 91: [It's confusing using it for the variable name here](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_scopes). What do you think of `$TargetEnvVar`?

chawyehsu:
Is it a replacement of #5813?

lib/system.ps1
niheaven: Great, I'll change it.

niheaven:
> Is it a replacement of #5813?

Yes, after reviewing the above PR, I found a simpler and clearer way to implement the feature. This was especially true after refactoring the EnvVar-related functions.

Initially, I attempted to modify pull request #5813, but ultimately decided it would be easier to create a new one instead.

lib/install.ps1
```diff
@@ -929,13 +935,21 @@ function env_rm_path($manifest, $dir, $global, $arch) {
     $env_add_path = arch_specific 'env_add_path' $manifest $arch
     $dir = $dir.TrimEnd('\')
     if ($env_add_path) {
+        if (get_config USE_ISOLATED_PATH) {
+            $target_path = 'SCOOP_PATH'
+        } else {
+            $target_path = 'PATH'
+        }
```
chawyehsu, line 942: I'm think of a seamless transition of this feature. With current implementation, there are two situations I concern:

- If one turns on `use_isolated_path` and then updates one installed app, in this case the path added to `$env:PATH` by the installed version can't be removed and remains.
- If one turned on `use_isolated_path` and installed one app, then they turn off `use_isolated_path` and try to uninstall the app, in this case the path added to `$env:SCOOP_PATH` can't be removed and remains.

Both situations can be mitigated by checking `$env:PATH` and `$env:SCOOP_PATH` side by side whatever the `use_isolated_path` would be. WDYT?

niheaven: > For Path switching, I'll make a PR by using https://github.com/ScoopInstaller/Scoop/pull/5851's Complete-ConfigChange().

This is one possible solution, when the user changes the config, all paths will be moved. And of course, removing from all target env sounds reasonable.

chawyehsu: #5851 is not going to be included in v0.4.0 while this PR will be, so there should be a preliminary solution.

niheaven: > #5851 is not going to be included in v0.4.0 while this PR will be, so there should be a preliminary solution.

btw, #5851 is ready for review. I've been using it for several days now and everything is working well.

niheaven: Done. And I'll check if there's a more efficient way.

niheaven: Aha, okay, I could check all paths in PATH that start with $scoopdir. Will fix it later.

chawyehsu: Oh you probably misunderstood my point. I didn't mean to force converting all previously added PATH. I meant it should check both `$env:PATH` and `$env:SCOOP_PATH` when doing uninstallation so there won't be paths leftover when users toggle the config back and forth.

niheaven: All done. Now switching won't waste any more time and uninstalling will remove paths from `$env:PATH` and `$env:SCOOP_PATH`.

lib/core.ps1
```diff
@@ -142,11 +142,63 @@ function set_config {
...
+        [string]
+        $Value
+    )
+
+    if ($Name -eq 'use_isolated_path') {
+        . "$PSScriptRoot\..\lib\system.ps1"
+
+        if ($Value -eq $true) {
+            info 'Switching to isolated path... This may take a while, please wait.'
+            $movedPath = Remove-Path -Path "$scoopdir\apps\*" -Quiet -PassThru
```
chawyehsu, line 170: `apps\scoop`?

niheaven: That dir is not in PATH, or I'm missing sth?

niheaven: `scoop\shims` is there, and we only shim `scoop.ps1`.

lib/install.ps1
```diff
@@ -906,19 +906,13 @@ function env_add_path($manifest, $dir, $global, $arch) {
...
-        # GH-3785: Add path in ascending order.
-        [Array]::Reverse($env_add_path)
-        $env_add_path | Where-Object { $_ } | ForEach-Object {
-            if ($_ -eq '.') {
-                $path_dir = $dir
-            } else {
-                $path_dir = Join-Path $dir $_
-            }
-            if (!(is_in_dir $dir $path_dir)) {
-                abort "Error in manifest: env_add_path '$_' is outside the app directory."
```
chawyehsu, line 918: This error report is useful but removed, is it expected?

lib/install.ps1
```diff
@@ -906,19 +906,13 @@ function env_add_path($manifest, $dir, $global, $arch) {
...
-            }
-            Add-Path -Path $path_dir -Global:$global -Force
+        if (get_config USE_ISOLATED_PATH) {
+            Add-Path -Path '%SCOOP_PATH%' -Global:$global
+            $target_path = 'SCOOP_PATH'
+        } else {
+            $target_path = 'PATH'
         }
+        Add-Path -Path ($env_add_path.Where({ $_ -and !$_.Contains(':') }).ForEach({ Join-Path $dir $_ })) -TargetEnvVar $target_path -Global:$global -Force
```
chawyehsu, line 915: What's the reason of filtering entry having `:`?

lib/install.ps1
```diff
@@ -906,19 +906,13 @@ function env_add_path($manifest, $dir, $global, $arch) {
...
-            if (!(is_in_dir $dir $path_dir)) {
-                abort "Error in manifest: env_add_path '$_' is outside the app directory."
-            }
-            Add-Path -Path $path_dir -Global:$global -Force
+        if (get_config USE_ISOLATED_PATH) {
+            Add-Path -Path '%SCOOP_PATH%' -Global:$global
+            $target_path = 'SCOOP_PATH'
+        } else {
+            $target_path = 'PATH'
         }
```
chawyehsu, line 914:
```suggestion
        $target_path = 'PATH'
        if (get_config USE_ISOLATED_PATH) {
            Add-Path -Path '%SCOOP_PATH%' -Global:$global
            $target_path = 'SCOOP_PATH'
        }
```

lib/install.ps1
niheaven: It's clear that `$path_dir` is `Join-Path $dir $_` and is in `$dir`, so the test here could never be run.

lib/install.ps1
niheaven: Hmm, is there any difference between these two implements?

chawyehsu: hoisting variable, less loc

lib/install.ps1
chawyehsu: Then what about relative path like `../<path>` that escapes from it? Before this it was checked by `is_in_dir` but you removed it in https://github.com/ScoopInstaller/Scoop/pull/5840/files#r1559002809

niheaven: Wow, you're right. I'll add it back.

niheaven: Fix it now and add some tests.

rashil2000:
I think we should use isolated path just by default (instead of making it a config option), seeing as it doesn't seem to have any disadvantages. Or does it, maybe I missed something?

CEbbinghaus:
This limits the path to SCOOP_PATH correct? Wouldn't it be better to allow the user to choose which name the environment variable takes?

niheaven:
> This limits the path to SCOOP_PATH correct? Wouldn't it be better to allow the user to choose which name the environment variable takes?

For convenience, a fixed name is better for users.

> I think we should use isolated path just by default (instead of making it a config option), seeing as it doesn't seem to have any disadvantages. Or does it, maybe I missed something?

Yes, it is, and we still need a migration. @ScoopInstaller/maintainers How about this feature?

rashil2000:
> For convenience, a fixed name is better for users.

I agree, it's sufficient and self-explanatory.

> Yes, it is, and we still need a migration. @ScoopInstaller/maintainers How about this feature?

Migration will be needed. We can even update the Readme and the website intro, saying that PATH pollution is now handled better because SCOOP_PATH is modified now, instead of directly modifying PATH.

chawyehsu:
> I think we should use isolated path just by default (instead of making it a config option)

We need a transition period.

rashil2000:
What for, if I may ask? 
The `scoop update` command can handle the environment change without user intervention, right? It doesn't seem to be a breaking change.

niheaven:
By setting `SCOOP_PATH` as default, I will make a commit and do migration as `Complete-ConfigChange()`.

lib/install.ps1
```diff
@@ -906,34 +906,23 @@ function env_add_path($manifest, $dir, $global, $arch) {
...
-                $path_dir = $dir
-            } else {
-                $path_dir = Join-Path $dir $_
-            }
-            if (!(is_in_dir $dir $path_dir)) {
-                abort "Error in manifest: env_add_path '$_' is outside the app directory."
-            }
-            Add-Path -Path $path_dir -Global:$global -Force
+        $target_path = 'PATH'
+        if (get_config USE_ISOLATED_PATH) {
```
rashil2000, line 910: This config is still there - intentional?

niheaven: By mistake 🤣

niheaven:
Got the idea. So it's better to offer config now and notify users that we'll change the default behavior and change it a few months later.

What's your opinion? @chawyehsu @rashil2000

rashil2000:
I don't have an issue with providing a config option per se - I just don't see purpose in involving user interaction in an internal implementation detail.

niheaven:
So choose an implementation here and then we can start bumping the version.

If config, I'll revert the last commit and wait for several months to submit a PR again, if default, then directly merge all commits.

lib/core.ps1
```diff
@@ -142,11 +142,28 @@ function set_config {
...
+    param (
+        [Parameter(Mandatory, Position = 0)]
+        [string]
+        $Name,
+        [Parameter(Mandatory, Position = 1)]
+        [AllowEmptyString()]
+        [string]
+        $Value
+    )
+    # Add needed config change logic here
```
rashil2000, line 164: An empty function?

libexec/scoop-update.ps1
```diff
@@ -57,6 +57,22 @@ if(($PSVersionTable.PSVersion.Major) -lt 5) {
 }
 $show_update_log = get_config SHOW_UPDATE_LOG $true
 
+# Migrate PATH EnvVar to SCOOP_PATH
+$movedPath = Remove-Path -Path "$scoopdir\apps\*" -Quiet -PassThru
```
rashil2000, line 61: Will this be run every time on `scoop update`?

lib/core.ps1
niheaven: Yes, empty function, for later usage.

libexec/scoop-update.ps1
niheaven: Yes, check every time since we don't know if the user has mixed PATH. But could be removed in next version.

rashil2000: I see, makes sense.

lib/core.ps1
rashil2000: Umm, so we'll add it when we have usage? Right now it doesn't fit here.

niheaven: Hmm, see usage in #5851. will remove from here now

rashil2000:
So, I'm facing a weird error: 
image

Did you face/rectify this?

rashil2000:
In addition to `$scoopdir\apps\*` paths, the SCOOP_PATH should also contain `$scoopdir\shims`, right? I think we missed it here...

niheaven:
No, putting `$scoopdir\shims` directly in PATH should be preferred. This will not break scoop's feature in case of `SCOOP_PATH`-related error.

CEbbinghaus:
> > For convenience, a fixed name is better for users.
>
> I agree, it's sufficient and self-explanatory.

I am not arguing that it is not self explanatory. I am saying that limiting it to 1 single environment variable breaks peoples environments who use that environment variable for other purposes (e.g setting the path to the scoop executable). It is fairly easy to allow configuration of the environment variable so its not a huge cost while the benefit to end users might be not having to fix their setup.

chawyehsu:
>use that environment variable for other purposes (e.g setting the path to the scoop executable)

`SCOOP_PATHS`/`SCOOP_APPPATHS` would be a better name if you concern about this, I don't think `SCOOP_PATHS`/`SCOOP_APPPATHS` has been used elsewhere. It is feasible to allow variable env var with something like `use_isolated_path: "MY_PERFECT_ENV_NAME"` but I'm not a fan of this kind of flexibility.

> What's your opinion?
>
> So choose an implementation here and then we can start bumping the version.
>
> If config, I'll revert the last commit and wait for several months to submit a PR again, if default, then directly merge all commits.

Abstention

chawyehsu (approved):

PR #5841, opened 2024-03-22
perf(shim): Update kiennq-shim to v3.1

niheaven:
#### Description

Download from https://github.com/kiennq/scoop-better-shimexe/releases/latest

VirusTotal:

https://www.virustotal.com/gui/file/554aac610eae2cd61cae3dea0c66fd8283e95eb9c4af5dd9e591197c6a28e119?nocache=1

#### Checklist:


supporting/shims/kiennq/version.txt
```diff
@@ -1 +0,0 @@
-2.2.1
```
chawyehsu, line 1: Would be better to keep the file indicating what version is vendored.

niheaven: Readd it

chawyehsu (changes requested):
@kiennq Don't know if you have any plan of making a new release of the shim but we are going update the vendored one. Currently v3.0 is the planned but it was released in 2022. Are there any critical commits that need to be considered since then?

I'm blocking this PR to wait for response. Let's wait for 48 hrs at most.

kiennq:
> Currently v3.0 is the planned but it was released in 2022. Are there any critical commits that need to be considered since then?

There's a commit that prevents the shim from requiring elevated launch unexpectedly when the shim name contains `update`, `install` string.
kiennq/scoop-better-shimexe@e0fb559a9a9e4c4c09b00d65d131519f4c9781d8

I've released a new version.

niheaven:
Updated to v3.1

chawyehsu:
Thanks for the quick response.

chawyehsu (approved):

PR #5872, opened 2024-04-09
fix(shim): Run JAR file from app's root directory

niheaven:
#### Description

For Java apps, the working directory should be apps' root directory since config files are normally located beside `.jar`.

#### Checklist:


niheaven:
I'm not sure if `.py` or other executives need such tweaks.

rashil2000:
Why WSL check?

niheaven:
WSL couldn't recognize `D:/aaa/bbb`-type path, since it uses `/mnt/D/aaa/bbb`, while Cygwin could access `D:/aaa/bbb` correctly.

niheaven:
Some use cases: https://github.com/ScoopInstaller/Main/pull/5702, https://github.com/ScoopInstaller/Main/pull/5704

rashil2000:
I recognize the usefulness of this feature, but still don't get the WSL part - do we really need to handle the use case where bash scripts (meant to run in Git Bash etc.) would be run inside WSL Bash? Wouldn't that lead to additional problems?

niheaven:
WSL shares PATH with the host machine, so if you install Java and can run in pwsh, you can run it in WSL too. That's why we should consider such cases.

rashil2000 (approved):

PR #5921, opened 2024-04-24
fix(json): Serialize jsonpath return

### LINKED ISSUE #5920: [Bug] `json_path` returning string literal `\n` when using union operator
#### Current Behavior
Currently, `json_path` would return the string literal `\n` when using union operator (`[,]`).

#### Expected Behavior
Instead, `json_path` should return the values with new lines, separated by commas, and contained in brackets (`[]`).

I am trying to extract the version and UUID from this JSON <https://api.azul.com/metadata/v1/zulu/packages?java_version=17&os=windows&arch=x86&archive_type=zip&java_package_type=jdk&javafx_bundled=false&crac_supported=false&latest=true&page_size=2>.

Using this JSON path should returns 4 different values
```
$..['distro_version','package_uuid']
```

> Note that <https://jsonpath.com/> only accept this syntax
> ```
> $..distro_version,package_uuid
> ```

Currently, the returned value is
```
[
  17,
  48,
  15,
  0
]\ncef00b50-77db-400a-afdd-b9e391e44631\n[
  17,
  48,
  15,
  0
]\n25351721-5e7e-4eef-a0ef-0c3686b38631
```

However, I don't think this is a feature and trying to the match regex for it would break if this is changed/fixed.

### System details

- **Windows version:** 11
- **OS architecture:** 64bit
- **PowerShell version:** 7.4.2

chawyehsu:
This may break checkver of some manifests if they leverage the old format of return. We can update those manifests once found.

#### Motivation and Context

Closes #5920

Relates to #XXXX

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,15 @@
 
 - **scoop-search:** Use SQLite for caching apps to speed up local search ([#5851](https://github.com/ScoopInstaller/Scoop/issues/5851), [#5918](https://github.com/ScoopInstaller/Scoop/issues/5918))
 
+### Bug Fixes
+
+- **json:** Serialize jsonpath return ([#5921](https://github.com/ScoopInstaller/Scoop/issues/5921))
+
 ## [v0.4.1](https://github.com/ScoopInstaller/Scoop/compare/v0.4.0...v0.4.1) - 2024-04-25
 
 ### Bug Fixes
 
 - **core:** Fix `Invoke-ExternalCommand` regression ([#5923](https://github.com/ScoopInstaller/Scoop/issues/5923))
-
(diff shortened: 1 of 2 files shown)


niheaven:
In the given example, `json_path` will yield `[[17,50,19,0],"87c7b863-4be0-4324-ab84-dfde19bccc18",[17,50,19,0],"12ad915b-781e-4859-aedb-2d1e51f9d384"]`. Can this be easily handled subsequently?

chawyehsu:
The output will be handled by the specified regex next to the jsonpath, so it's up to the manifest author. I believe it'd help to write simpler regex by yielding such clean json output.

niheaven (approved):

PR #5929, opened 2024-04-28
feat(core): New cache filename format

### LINKED ISSUE #4327: cached files with URL and filename > 260 characters issue and possible solution

chawyehsu:
#### Description

Replaces #4349, with sha1 for shorter output. For a better migration experience and implicit bandwidth saving, old cache files will still be used when available but this check may be removed sometime in the future.

#### Motivation and Context

Fixes #4327
Closes #4349

unit tests added

#### Checklist:

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -419,7 +419,7 @@ function cache_path($app, $version, $url) {
     }
 
     $urlStream = [System.IO.MemoryStream]::new([System.Text.Encoding]::UTF8.GetBytes($url))
-    $sha = (Get-FileHash -Algorithm SHA1 -InputStream $urlStream).Hash.ToLower()
+    $sha = (Get-FileHash -Algorithm SHA256 -InputStream $urlStream).Hash.ToLower().Substring(0, 7)
     $extension = [System.IO.Path]::GetExtension($url)
     $filePath = $filePath -replace "$underscoredUrl", "$sha$extension"
 
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -413,7 +413,7 @@ function cache_path($app, $version, $url) {
     $underscoredUrl = $url -replace '[^\w\.\-]+', '_'
     $filePath = "$cachedir\$app#$version#$underscoredUrl"
 
-    # NOTE: Scoop cache files migration. Remove this 6 months after it ships.
+    # NOTE: Scoop cache files migration. Remove this 6 months after the feature ships.
     if (Test-Path $filePath) {
         return $filePath
     }
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -409,7 +409,22 @@ function currentdir($app, $global) {
 function persistdir($app, $global) { "$(basedir $global)\persist\$app" }
 function usermanifestsdir { "$(basedir)\workspace" }
 function usermanifest($app) { "$(usermanifestsdir)\$app.json" }
-function cache_path($app, $version, $url) { "$cachedir\$app#$version#$($url -replace '[^\w\.\-]+', '_')" }
+function cache_path($app, $version, $url) {
+    $underscoredUrl = $url -replace '[^\w\.\-]+', '_'
+    $filePath = "$cachedir\$app#$version#$underscoredUrl"
+
+    # NOTE: Scoop cache files migration. Remove this 6 months after it ships.
+    if (Test-Path $filePath) {
+        return $filePath
+    }
+
+    $urlStream = [System.IO.MemoryStream]::new([System.Text.Encoding]::UTF8.GetBytes($url))
+    $sha = (Get-FileHash -Algorithm SHA1 -InputStream $urlStream).Hash.ToLower()
+    $extension = [System.IO.Path]::GetExtension($url)
+    $filePath = $filePath -replace "$underscoredUrl", "$sha$extension"
+
+    return $filePath
+}
 
 # apps
 function sanitary_path($path) { return [regex]::replace($path, "[/\\?:*<>|]", "") }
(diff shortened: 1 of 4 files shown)


lib/core.ps1
```diff
@@ -409,7 +409,22 @@ function currentdir($app, $global) {
...
+    $underscoredUrl = $url -replace '[^\w\.\-]+', '_'
+    $filePath = "$cachedir\$app#$version#$underscoredUrl"
+
+    # NOTE: Scoop cache files migration. Remove this 6 months after the feature ships.
+    if (Test-Path $filePath) {
+        return $filePath
+    }
+
+    $urlStream = [System.IO.MemoryStream]::new([System.Text.Encoding]::UTF8.GetBytes($url))
+    $sha = (Get-FileHash -Algorithm SHA1 -InputStream $urlStream).Hash.ToLower()
```
niheaven, line 422:
```suggestion
    $sha = (Get-FileHash -Algorithm SHA256 -InputStream $urlStream).Hash.ToLower().Substring(0, 7)
```

We could utilize the first seven digits of the SHA-256 hash, similar to GitHub's commit reference, as it is concise and sufficiently unique.

test/Scoop-Core.Tests.ps1
```diff
@@ -259,6 +259,29 @@ Describe 'get_app_name_from_shim' -Tag 'Scoop', 'Windows' {
     }
 }
 
+Describe 'cache_path' -Tag 'Scoop' {
+    BeforeAll {
+        $cachedir = "$([IO.Path]::GetTempPath())ScoopTestFixtures\cache"
+    }
+
```
niheaven, line 266:
```suggestion
```

Use `Mock` below to avoid `BeforeAll`

test/Scoop-Core.Tests.ps1
```diff
@@ -259,6 +259,29 @@ Describe 'get_app_name_from_shim' -Tag 'Scoop', 'Windows' {
...
+    It 'returns the correct cache path for a given input' {
+        $url = 'https://example.com/git.zip'
+        $ret = cache_path 'git' '2.44.0' $url
+        $inputStream = [System.IO.MemoryStream]::new([System.Text.Encoding]::UTF8.GetBytes($url))
+        $sha = (Get-FileHash -Algorithm SHA1 -InputStream $inputStream).Hash.ToLower()
```
niheaven, line 271:
```suggestion
        $sha = (Get-FileHash -Algorithm SHA256 -InputStream $inputStream).Hash.ToLower().Substring(0, 7)
```

test/Scoop-Core.Tests.ps1
```diff
@@ -259,6 +259,29 @@ Describe 'get_app_name_from_shim' -Tag 'Scoop', 'Windows' {
...
+        $ret | Should -Be "$cachedir\git#2.44.0#$sha.zip"
+    }
+
+    # # NOTE: Remove this 6 months after the feature ships.
+    It 'returns the old format cache path for a given input' {
+        $fixture = "$cachedir\git#2.44.0#https_example.com_git.zip"
+        New-Item -ItemType File -Path $fixture -Force
```
niheaven, line 278:
```suggestion
        Mock Test-Path { $true }
```

Use `Mock` here.

test/Scoop-Core.Tests.ps1
```diff
@@ -259,6 +259,29 @@ Describe 'get_app_name_from_shim' -Tag 'Scoop', 'Windows' {
...
+        $ret = cache_path 'git' '2.44.0' 'https://example.com/git.zip'
+        $ret | Should -Be $fixture
+        Remove-Item -Path $fixture -Force
```
niheaven, line 281:
```suggestion
        $ret | Should -Be "$cachedir\git#2.44.0#https_example.com_git.zip"
```

niheaven (commented):
btw, I believe the migration period should be extended, as many apps are not updated for years.

chawyehsu:
I have no idea how frequently our users cleaning and rebuiding their caches, but the max extended support period I would do would be 1 year. It is however just an expectation and it may be there for a much longer time - _we may forget to update it, [most of the time](https://github.com/ScoopInstaller/Scoop/pull/5715) ;)_

niheaven (approved):

PR #5945, opened 2024-05-08
fix(core): Fix "Invoke-ExternalCommand" quoting rules

niheaven:
#### Description

Quoting rules...

There's a bug in PS5 if `extract_dir` has spaces when using 7z to extract the archive, and this PR tries to fix it.

#### Motivation and Context

- Closes https://github.com/ScoopInstaller/Extras/issues/13174
- Closes https://github.com/ScoopInstaller/Extras/issues/13221
- Closes https://github.com/ScoopInstaller/Extras/issues/13227
- Closes https://github.com/ScoopInstaller/Extras/issues/13255

There's a known bug that NSIS's `/D` param doesn't work if the path has spaces in PS7, if we provide `/D=xxx xxx` in `ArgumentList`, the NSIS installer doesn't recognize it, don't know why.

Local install affected apps in PS5 and add some test cases.

#### Checklist:


chawyehsu:
Could you please add unit tests for this type of situation to the Decompress testsuite?

---

The [NSIS Document](https://nsis.sourceforge.io/Docs/Chapter3.html) says that there are two constrains of the `/D` switch.
- **NO** quotes (even if the path contains spaces)
- It **MUST** be the last parameter used in the command line

> There's a known bug that NSIS's `/D` param doesn't work if the path has spaces in PS7, if we provide `/D=xxx xxx` in `ArgumentList`, the NSIS installer doesn't recognize it, don't know why.

Probably the second constrain is the cause in this case?

lib/core.ps1
```diff
@@ -771,7 +771,8 @@ function Invoke-ExternalCommand {
         $Process.StartInfo.WindowStyle = [System.Diagnostics.ProcessWindowStyle]::Hidden
     }
     if ($ArgumentList.Length -gt 0) {
-        $ArgumentList = $ArgumentList | ForEach-Object { [regex]::Split($_.Replace('"', ''), '(?<=(?<![:\w])[/-]\w+) | (?=[/-])') }
+        # Remove existing double quotes and split arguments
+        $ArgumentList = $ArgumentList.ForEach({ $_ -replace '"' -split '(?<=(?<![:\w])[/-]\w+) | (?=[/-])' })
```
chawyehsu, line 775: I would like to have some extra anotations explaining this sophisicated regex pattern.

niheaven: Will add it.

niheaven:
Alright, I will implement some test cases and refactor the decompression test structure accordingly.

niheaven:
Test cases added, all done.

chawyehsu:
I think I found out the cause of the `/D` switch getting ignored in PS7 @niheaven

image

```pwsh
$Process = New-Object System.Diagnostics.Process
$Process.StartInfo.FileName = '...exe'
$Process.StartInfo.ArgumentList.Add('/S')
$Process.StartInfo.ArgumentList.Add('/D=C:\temp\miniconda space space')
```

If an arguement that is added to `ProcessStartInfo.ArgumentList` contains whitespaces, it will be wrapped with quotes before being passed to the native command, this is an expected behavoir for most cases but not for the NSIS installer.

Probably we have to use the legacy argument escaping necessarily for NSIS installers, whether in PS5 or PS7, i.e. there needs to be one more condition of testing NSIS installers in 
https://github.com/ScoopInstaller/Scoop/blob/36026f1804090b1726fe8a7df699ca739fdb9cbc/lib/core.ps1#L782

chawyehsu:
Regarding to the TestCases, perhaps there should also be test nsis installers to validate such a situation.

niheaven:
> Probably we have to use the legacy argument escaping necessarily for NSIS installers

Indeed, that was my initial thought, but I was unable to find an efficient method to verify it. PEiD or Exeinfo should be able to accomplish the task, yet I was unable to discern their methodology. (Using a hex editor, I could find `Nullsoft` in PE, but couldn't with PS)

chawyehsu:
Reading hex to identify NSIS installers requires extra IO. What about picking the inactive #3502 up? That may take much work though, and a workaround (by just checking if `/D=<path>` exists?) should be released to mitigate existing issues.

niheaven:
I am currently addressing issue #3502 by dividing it into multiple individual PRs, with some already merged, the most recent being #5951. I will also review any existing manifest that utilizes `/D=`, as I recall that other installers may use this as well.

niheaven:
Check NSIS installer by its arguments (should have `/S` and `/D=xxx`), and add test cases.

chawyehsu:
> Check NSIS installer by its arguments (should have /S and /D=xxx)

That's the workaround I was thinking of.

test/Scoop-Decompress.Tests.ps1
```diff
@@ -94,21 +113,39 @@ Describe 'Decompression function' -Tag 'Scoop', 'Windows', 'Decompress' {
...
 
+        It 'extract NSIS installer' {
+            $to = test_extract 'Expand-7zipArchive' $test7
+            $to | Should -Exist
+            "$to\empty" | Should -Exist
+            (Get-ChildItem $to).Count | Should -Be 1
+        }
+
+        It 'self-extract NSIS installer' {
+            $to = "$working_dir\NSIS Test"
```
chawyehsu, line 124: 👍

chawyehsu (approved):

RaphaelTarita:
also closes https://github.com/ScoopInstaller/Extras/issues/13160. When is this fix expected to be released?

niheaven:
Although I believe the SQLite cache feature is stable enough for the end user, cherry-picking some fixes to the main branch is a decision that requires careful consideration. I will proceed with it now and create a pull request.

PR #5956, opened 2024-05-10
fix(scoop-download|install|update): Use consistent options

lewis-yeung:
### Description

~This PR is to fix a bug: the shorthand of the option `--no-hash-check` for the command `scoop download` should be `-s` instead of `-h` to be consistent with `scoop install` and `scoop update`, since the flag `-h` always means printing help text for a command.~

~BTW, the option `--no-hash-check` for `scoop download` is also inconsistent with `--skip` for `scoop install` or `scoop update`, even though they work the same way. For backwards compatibility, I did not adjust it. Should we?~

**EDITED:** Use consistent options (`-s/--skip-hash-check`) for `scoop download`, `scoop install` and `scoop update`.

### How Has This Been Tested?

~- Run `scoop download -h`, and it prints the help text for us.~
~- Run `scoop download -s <app>`, and it skips the hash verification after the download, which is the expected behavior.~

1. Run the following to check available CLI options in the help text:

   - `scoop download -h`
   - `scoop install -h`
   - `scoop update -h`

2. Run the following to skip hash validation:

   - `scoop download -s <app>`
   - `scoop download --skip-hash-check <app>`
   - `scoop install -s <app>`
   - `scoop install --skip-hash-check <app>`
   - `scoop update -s <app>`
   - `scoop update --skip-hash-check <app>`

### Checklist:


niheaven:
Prefer renaming all args to `--skip-hash-check` w/ `-s`.

lewis-yeung:
> Prefer renaming all args to `--skip-hash-check` w/ `-s`.

@niheaven Agreed. Will update later.

niheaven (dismissed):
LGTM, wait for other maintainers

niheaven (approved):

niheaven (approved):

PR #5983, opened 2024-05-21
fix(decompress): Match `extract_dir`/`extract_to` and archives

lewis-yeung:
### Description

This PR is to fix a regression introduced in the refactoring of #5951 which broke the match between entries of `extract_dir`/`extract_to` lists and files that require unpacking:

https://github.com/ScoopInstaller/Scoop/blob/a5bd2297c6227e5b09869acafaa12cbdc73f6132/lib/decompress.ps1#L22-L26

The directories specified by `extract_dir`/`extract_to` fields in an app manifest are only meaningful for download files that require unpacking, if a download URL of an unpacking-free file is placed before those of archive files, the latter will not be unpacked from/to the specified paths.

### How Has This Been Tested?

Tested by running:

```powershell
scoop install 'https://raw.githubusercontent.com/lewis-yeung/scoop-bucket/a9b9df6f918995d3c6f1323a6133b20a0d7b4c6d/bucket/oh-my-posh.json'
```

It works properly after the fix.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - **scoop-download|install|update:** Use consistent options ([#5956](https://github.com/ScoopInstaller/Scoop/issues/5956))
 - **core:** Fix "Invoke-ExternalCommand" quoting rules ([#5945](https://github.com/ScoopInstaller/Scoop/issues/5945))
 - **scoop-info:** Fix download size estimating ([#5958](https://github.com/ScoopInstaller/Scoop/issues/5958))
+- **decompress:** Match `extract_dir`/`extract_to` and archives ([#5983](https://github.com/ScoopInstaller/Scoop/issues/5983))
 
 ### Code Refactoring
 
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -13,20 +13,16 @@ function Invoke-Extraction {
         $ProcessorArchitecture
     )
 
-    # 'url', 'extract_dir' and 'extract_to' are paired
     $uri = @(url $Manifest $ProcessorArchitecture)
+    # 'extract_dir' and 'extract_to' are paired
     $extractDir = @(extract_dir $Manifest $ProcessorArchitecture)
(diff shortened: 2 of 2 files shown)


niheaven:
The optimal approach is to move these files to the end, or prefix them with a `.` in `extract_dir`/`extract_to`.

In this PR, `extracted` is unused and should be removed.

lewis-yeung:
@niheaven The code must work without modifying any existing app manifests. Right?

> The optimal approach is to move these files to the end,

Which kind of files are you referring to? Files that do not need to be unpacked?

> or prefix them with a `.` in `extract_dir`/`extract_to`.

Do you mean filtering out non-archive files and adding placeholders for them in the `$extractDir`/`$extractTo` list? This approach is not that straightforward compared to the previous code.

> In this PR, `extracted` is unused and should be removed.

Is it? I mean, it counts unpackable files here and ensures that they are correctly paired with entries of `extract_dir`/`extract_to`:

https://github.com/ScoopInstaller/Scoop/blob/d693af1ef259892fb1cb429f2b7c86799f8a06ea/lib/decompress.ps1#L60

which is exactly what the previous code did.

niheaven:
https://github.com/ScoopInstaller/Scoop/blob/5819b5a1ef6ee8740c4310f0d0ba06a88c39917f/lib/install.ps1#L590-L595

The term `extracted` was previously used to identify `extract_dir`/`extract_to` because the old method employed a `foreach` routine. However, since I've switched to a `for` loop, the enumerator now serves to identify each item.

lewis-yeung:
@niheaven

> The term `extracted` was previously used to identify `extract_dir`/`extract_to` because the old method employed a `foreach` routine. However, since I've switched to a `for` loop, the enumerator now serves to identify each item.

The fact is that `extract_dir`/`extract_to` entries are for archive files only and do not always align with URLs. The `for` loop here broke the previously working cases. Without further filtering logic, the `$extracted` counter works and is pretty straightforward. Why not?

niheaven:
You're right, I misunderstood what you meant.

niheaven (approved):

PR #5984, opened 2024-05-22
added the ability to specify @ version when installing from url

### LINKED ISSUE #5977: Autoupdate to manifest url
Hello everyone!

I was wondering if there was any way i could use the `autoupdate`  name@version feature with a manifest specified as a path and not from a bucket.

For example lets say i have a `manifest.json` that has an `autoupdate` field. If I were to run:

`scoop install ./manifest.json@20`

I would receive the error: `Couldn't find manifest for 'manifestjson@20'`

Is there any workaround to this? Any flags i could set?

insertokername:
added the ability to use the @ symbol when using `scoop install` with a url

#### Description

Simple change to the regex that matches the name of the app to not require the whole string to be ending with .json

fixes #5977 

This has been tested with all installation methods that have been mentioned in the `scoop install --help` 

#### Checklist:

- [X] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -1192,7 +1192,7 @@ function applist($apps, $global) {
 }
 
 function parse_app([string]$app) {
-    if ($app -match '^(?:(?<bucket>[a-zA-Z0-9-_.]+)/)?(?<app>.*\.json$|[a-zA-Z0-9-_.]+)(?:@(?<version>.*))?$') {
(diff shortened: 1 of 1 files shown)


insertokername:
Closed so i can make a new pull request that has the signed commit.

PR #6003, opened 2024-06-05
fix(scoop-alias): Pass options correctly

### LINKED ISSUE #5767: [Bug] `scoop-alias.ps1` `list` Verbose Flag Doesn't Produce `Summary` Output
#### Current Behavior

`Summary` property is not ever displayed at this piece of code. I've tried various forms of `scoop alias list -v` or `scoop alias list --verbose` or `scoop alias list -verbose:$true`

https://github.com/ScoopInstaller/Scoop/blob/f93028001fbe5c78cc41f59e3814d2ac8e595724/libexec/scoop-alias.ps1#L100-L104

#### Expected Behavior

I expect to see the `Summary` output along with `Name` and `Command`.

N/A

#### Possible Solution

~~No actual full solution yet.~~ The smallest possible solution is to change the below line to `if ($name -in @('-v', '--verbose')) {`. The splatting of `@arguments` upstream passes the `-v` or `--verbose` as plaintext to `$name` by position so the `$verbose` parameter isn't required. Not sure if this is the *best* or most *robust* solution.

https://github.com/ScoopInstaller/Scoop/blob/f93028001fbe5c78cc41f59e3814d2ac8e595724/libexec/scoop-alias.ps1#L100

~~#### Attempts To Fix~~
~~I've tried adding `$verbose` as a simple _parameter_ to `list_aliases` function but that didn't work.~~

~~After debugging a little, there is something going on with this line with the `@arguments` _splat_. It doesn't appear to work. When I pass the raw parameters in like `$cmd_path list -v` will produce the `Summary` output.~~

~~https://github.com/ScoopInstaller/Scoop/blob/f93028001fbe5c78cc41f59e3814d2ac8e595724/lib/commands.ps1#L35~~

#### References
#2163 PR

### System details

_Windows Version 22H2 (OS Build 19045.3803)_

_64 Bit_

```powershell
Name                           Value
----                           -----
PSVersion                      7.3.10
PSEdition                      Core
GitCommitId                    7.3.10
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

_No Additional Software_

#### Scoop Configuration

```json
{
  "last_update": "2023-12-22T14:31:06.1549275-06:00",
  "alias": {
    "describe": "scoop-describe",
    "fsearch": "scoop-fsearch"
  },
  "scoop_branch": "master",
  "scoop_repo": "https://github.com/ScoopInstaller/Scoop"
}
```

lewis-yeung:
### Description

This PR has fixed a bug that the option `-v`/`--verbose` cannot be passed to the command `scoop alias list`.

### Motivation and Context

Closes #5767.

### How Has This Been Tested?

Suppose we have an alias with `Summary` properly set, run:

```powershell
scoop alias list -v | Format-List # the property `Summary` is showed
scoop alias list | Format-List # the property `Summary` is not showed
```

### Checklist:


mattcargile:
Nice. This looks like it fixes my issue.
https://github.com/ScoopInstaller/Scoop/issues/5767

lib/alias.ps1
```diff
@@ -0,0 +1,83 @@
+. "$PSScriptRoot\..\lib\install.ps1" # shim related
```
niheaven, line 1: Don't load library insides another library, move it to `scoop-alias.ps1` and add `install.ps1` to test file.

lib/alias.ps1
```diff
@@ -0,0 +1,83 @@
...
+
+$script:config_alias = 'alias'
```
niheaven, line 3: Why create a script var?

lib/alias.ps1
```diff
@@ -0,0 +1,83 @@
...
+
+function init_alias_config {
+    $aliases = get_config $script:config_alias
+    if ($aliases) {
+        $aliases
+    } else {
+        New-Object -TypeName PSObject
```
niheaven, line 10: Just create an object and doesn't assign it?

lib/alias.ps1
lewis-yeung: Uh oh, this was moved from the original code and left unchanged. Will change it.

lib/alias.ps1
lewis-yeung: Same as above. The necessary refactoring will be done.

niheaven:
There're some bugs in the code, and I push some tweaks:

- Move alias-related code to `lib/commands.ps1`
- Pipe string into `Out-UTF8File()`
- Remove `get_aliases_from_config()` and use `get_config()` directly

niheaven (approved):

PR #6024, opened 2024-06-21
fix(checkver): Correct error messages

lewis-yeung:
### Description

When checking update and the `$script` is used, an error message should mention `the output of script` instead of `$url`.

### Checklist:

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -294,8 +294,10 @@ while ($in_progress -gt 0) {
             }
             $page = (New-Object System.IO.StreamReader($ms, (Get-Encoding $wc))).ReadToEnd()
         }
+        $source = $url
         if ($script) {
             $page = Invoke-Command ([scriptblock]::Create($script -join "`r`n"))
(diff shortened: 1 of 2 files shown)


bin/checkver.ps1
```diff
@@ -285,7 +285,11 @@ while ($in_progress -gt 0) {
             continue
         }
 
-        if ($url) {
+        if ($script) {
+            $source = 'the output of script'
+            $page = Invoke-Command ([scriptblock]::Create($script -join "`r`n"))
+        } elseif ($url) {
+            $source = $url
```
niheaven, line 292: The `$page` parameter can be utilized within the `script`. Please restore the original sequence and seek an alternative method to rectify the output.

lewis-yeung: Sure. I'll fix it later.

lewis-yeung: @niheaven Updated.

niheaven (approved):

PR #6030, opened 2024-06-25
fix(shim): Restore original path for JAR cmd

### LINKED ISSUE #6026: [Bug] .jar shims change the working directory when running from cmd
#### Current Behavior
`.jar` shims change the current working directory to the package directory when running from cmd.

#### Expected Behavior
shims should not change the working directory

for example, install apktool: `scoop install main/apktool` and then run `apktool` in cmd:

```
C:\>apktool --version
2.9.3

C:\Users\User\scoop\apps\apktool\current>
```
When running from powershell this doesn't happen for some reason.

The cause of this is 
https://github.com/ScoopInstaller/Scoop/blob/d337bb1fa69a56637101254a1ac9e0ad13a12647/lib/core.ps1#L1012-L1018

I think `cd` shouldn't be used anywhere in shims.

### System details

**Windows version:**  11

**OS architecture:** 64bit

**PowerShell version:** 5.1.22621.3672

#### Scoop Configuration

```json
{
    "last_update":  "2024-06-23T08:09:07.4455297+03:00",
    "scoop_branch":  "master",
    "scoop_repo":  "https://github.com/ScoopInstaller/Scoop"
}

```

niheaven:
#### Description

Use `pushd` and `popd` to restore cmd path

#### Motivation and Context

- Closes #6026

- Relates to #5872

image

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - **core:** Use correct path in 'bash' ([#6006](https://github.com/ScoopInstaller/Scoop/issues/6006))
 - **core:** Limit the number of commands to get when search for git executable ([#6013](https://github.com/ScoopInstaller/Scoop/pull/6013))
 - **checkver:** Correct error messages ([#6024](https://github.com/ScoopInstaller/Scoop/issues/6024))
+- **shim:** Restore original path for JAR cmd ([#6030](https://github.com/ScoopInstaller/Scoop/issues/6030))
 
 ### Code Refactoring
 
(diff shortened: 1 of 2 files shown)


davidpatel0:
This does not resolve the issue, `cd` shouldn't be used at all. Another problem with the current state (that this PR doesn't fix) is that using relative paths with `.jar` shims is broken (as they are being resolved from `scoop\apps\X` directory and not from the real working directory.
I will add this problem to the original issue.

niheaven:
Yes, there are some drawbacks to this fix and using `pushd/popd`, but the working directory is necessary for some jar applications.

Perhaps there is a better solution, but for now, use this as a temporary bug fix.

PR #6031, opened 2024-06-27
Option to not add shortcuts to startmenu when installing a package

insertokername:
#### Description
Added an option to skip creating the startmenu shortcuts when installing a package.

#### Motivation and Context
This is really useful in a fer specific use cases, for example i have two scoop installations on my pc for various purposes. This flag really helps avoid conflicts between installs and also because the paths of my installs are manually managed this, again, helps a lot.

Ran the flag on multiple packages installs.
The code is really quite simple and only interacts with a small portion of the codebase.

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - **core:** New cache filename format ([#5929](https://github.com/ScoopInstaller/Scoop/issues/5929))
 - **install:** Added the ability to install specific version of app from URL/file link ([#5988](https://github.com/ScoopInstaller/Scoop/issues/5988))
 - **decompress:** Use innounp-unicode as default Inno Setup Unpacker ([#6028](https://github.com/ScoopInstaller/Scoop/issues/6028))
+- **scoop-install** Option to not add shortcuts to startmenu when installing a package
 
 ### Bug Fixes
 
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -5,7 +5,7 @@ function nightly_version($quiet = $false) {
     return "nightly-$(Get-Date -Format 'yyyyMMdd')"
 }
 
-function install_app($app, $architecture, $global, $suggested, $use_cache = $true, $check_hash = $true) {
+function install_app($app, $architecture, $global, $suggested, $use_cache = $true, $check_hash = $true, $add_startmenu = $true) {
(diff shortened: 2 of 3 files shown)


qwertyhjklxyz:
Running `scoop update|reset <app>` still creates the shortcut.
Running `scoop uninstall <app>` removes the shortcut, even if it was not created.
Maybe add this option to `install.json`.

insertokername:
Thank you for the help! I made some additions to hopefully fix my oversights!

After making these changes i observed some strange behaviour, even if there is no startmenu shortcut in the actuall folder windows still adds the program to the startmenu directly from the scoop/apps folder. I suspect this might be from register tomfoolery that is made by the installer but i am not sure. If you have any ideas they would be greatly appreciated!

Another thing i would like to mention is that shims also get added to the startmenu, i intend to not interfere with this since you can already manage shims through scoop shim and this option is intended to only affect the startmenu folder.

qwertyhjklxyz:
> After making these changes i observed some strange behavior, even if there is no startmenu shortcut in the actual folder windows still adds the program to the startmenu directly from the scoop/apps folder. I suspect this might be from registry tomfoolery that is made by the installer but i am not sure. If you have any ideas they would be greatly appreciated!

Do you mean apps in the `nonportable` repository, scoop just calls installer and doesn't control the behavior of the installer, but it is possible to remove shortcuts in the `post_install`.
> Another thing i would like to mention is that shims also get added to the startmenu, i intend to not interfere with this since you can already manage shims through scoop shim and this option is intended to only affect the startmenu folder.

Scoop doesn't add shim to the startmenu.

By the way, you didn't update the `scoop-update.ps1`.
`scoop-export.ps1` and `scoop-import.ps1` seem to need updating as well.

insertokername:
>Do you mean apps in the nonportable repository, scoop just calls installer and doesn't control the behavior of the installer, but it is possible to remove shortcuts in the post_install.

Noted, thank you for the help!

> Scoop doesn't add shim to the startmenu.

I accidentally confused the startmenu for the search, sorry! 

I pushed a fix for the update and the import, hope it addresses all issues! Thank you for your patience with me, this is my second ever contribution with actual code so sorry for the mistakes.

insertokername:
@niheaven is there any way you could take a quick look at this?

PR #6090, opened 2024-08-05
fix: Homepage no longer downloaded on script

### LINKED ISSUE #5704: [Bug] homepage should not be downloaded when checkver script is specified
#### Current Behavior

When `checkver.script` is specified in a manifest and `checkver.url` is not, checkver defaults to requesting the `homepage` url before running the script. If that request errors, the script is never run and checkver fails.
#### Expected Behavior

When `checkver.script` is specified, `homepage` and `checkver.url` should be ignored and the script should run as expected.

I changed the [homepage](https://github.com/kazzarin/bucket/blob/22fb2662e9143a1712b5d806abfc2b9947b65e60/bucket/oracle-graalvm.json#L4) for a manifest in my own bucket, which produced the following error:
```
> .\bin\checkver.ps1 oracle-graalvm
oracle-graalvm: The operation has timed out.
URL https://www.oracle.com/java/graalvm/ is not valid
```
Reverting the [change](https://github.com/kazzarin/bucket/blob/56732b8d74f4babfbd3400ff43829dd2b08ec36d/bucket/oracle-graalvm.json#L4) fixed it. The same behaviour was observed in my bucket's excavator, so I'm sure it's not a problem with my local machine. The wiki docs on the [script property](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate#properties-of-checkver) don't mention if it can be used with `checkver.url` or `homepage`, so I assume based on the manifests in the given examples that `checkver.script` is intended to be used exclusively when specified.

Unrelated, the site seems to timeout because of the link in scoop's user agent.
### System details

**Windows version:** 11

**OS architecture:** 64bit

**PowerShell version:** 7.3.9

**Additional software:** Git

#### Scoop Configuration

```json
{
    "aria2-enabled":  false,
    "aria2-warning-enabled":  false,
    "cat_style":  "plain",
    "debug":  true,
    "gh_token":  "redacted",
    "last_update":  "2023-10-27T22:05:56.4511315-04:00",
    "scoop_branch":  "develop",
    "scoop_repo":  "https://github.com/ScoopInstaller/Scoop",
    "show_update_log":  false,
    "use_external_7zip":  true
}

```

notPlancha:
Currently `checkver.ps1` GETs the homepage, even if `script` is specified, and does nothing with it. This becomes an issue when the homepage doesn't return a 2xx, like a timeout, or a `403 Forbiden` (like cloudflare does). I don't see any reason this is the case, so I wrapped this logic in an if/else. I am not sure if this is the best option, or if this is breaking at all (I don't think it is but I'm not completly sure)

fixes #5704.

PS: I'm not sure if I'm supposed to create the PR to the master branch or the develop branch (CONTRIBUTING.md isn't really clear on that). Right now they're synced so it shouldn't be an issue.


xrgzs (approved):

xrgzs:
I encountered this problem recently. I don't understand how this is designed, which means that the app will not be updated when its website runs away? Hope to merge this conmmit as soon as possible.

xrgzs:
@niheaven   @chawyehsu

notPlancha:
This was not targeting develop, #6230 now is.

PR #6092, opened 2024-08-06
fix (decompress): `Expand-7zipArchive` only delete temp dir / `$extractDir` if it is empty

### LINKED ISSUE #6011: [Bug] Missing folders after installing packages making applications behave incorrectly
## Bug Report
There's possibly an issue with Scoop installation process extracting from `.exe` installer files. As reported in [here](https://github.com/ScoopInstaller/Extras/issues/13324) and later figuring out in the discussion, for some reason installing the `tor-browser` package there's a folder missing (`Browser\browser`) that causes the error I experienced.

#### Current Behavior

After installing the `extras/tor-browser`, there's a folder missing that causes the app to not behave correctly. When installing the application downloading it manually, either through the installer or extracting using 7-zip, the issue does not occur. Copying the missing folder also seems to fix the issue.

#### Expected Behavior

Tor Browser should open and execute normally.

I think it's important to note that the issue appearing coincides with Scoop updating to version 0.4.2 and  `tor-browser` updating to 13.0.15, as noted by someone in the discussion.

#### Possible Solution

It's only a wild guess, but maybe there's something up with having one folder inside another with the same name only differing by the first letter being uppercase or not?

### System details

**Windows version:** 11 Pro 23H2

**OS architecture:** 64bit

**PowerShell version:** system is 5.1.22621.3672 and Microsoft store version is 7.4.2

**Additional software:** 7-zip is 24.06, git is 2.45.2

#### Scoop Configuration

```json
{
  "last_update": "2024-06-11T19:15:24.7094978-03:00",
  "aria2-warning-enabled": false,
  "aria2-disabled": true,
  "scoop_branch": "master"
}
```

o-l-a-v:
#### Description

`decompress.ps1` deletes `$extractDir` even if it's not empty, which can cause applications to not work after installation.

More context and how I got to this solution in following comment:

* <https://github.com/ScoopInstaller/Scoop/issues/6011#issuecomment-2271639656>

#### Motivation and Context

Closes #6011, <https://github.com/ScoopInstaller/Extras/issues/13324>

#### How Has This Been 

Tested locally with `scoop install extras/tor-browser`.

#### Checklist:


o-l-a-v:
If tests should be updated (tell me): Should be sufficient to add a `.7z` to <https://github.com/ScoopInstaller/Scoop/blob/develop/test/fixtures/decompress/TestCases.zip> with:

```
root
├─ $PLUGINSDIR
└─  keep
       └─ keep
             └─  file.txt
```

Then check that `"$DestinationPath\keep"` exists?

niheaven:
Could you please sign your commit? And the logic should be tweaked by only check if `$ExtractDir` is empty.

o-l-a-v:
> Could you please sign your commit?

I've never done that before, I'll se if I figure it out.

> And the logic should be tweaked by only check if $ExtractDir is empty.

Isn't that what I do? If no child items inside it => delete.

lib/decompress.ps1
```diff
@@ -123,8 +123,11 @@ function Expand-7zipArchive {
     }
     if (!$IsTar -and $ExtractDir) {
         movedir "$DestinationPath\$ExtractDir" $DestinationPath | Out-Null
-        # Remove temporary directory
-        Remove-Item "$DestinationPath\$($ExtractDir -replace '[\\/].*')" -Recurse -Force -ErrorAction Ignore
+        # Remove temporary directory if empty
+        $ExtractDirFullPath = [string] "$DestinationPath\$($ExtractDir -replace '[\\/].*')"
+        if ((Get-ChildItem -Path $ExtractDirFullPath -Force).Count -le 0) {
+            Remove-Item -Path $ExtractDirFullPath -Recurse -Force -ErrorAction Ignore
+        }
```
niheaven, line 130:
```suggestion
        if ((Get-ChildItem "$DestinationPath\$ExtractDir" -Force -ErrorAction Ignore).Count -eq 0) {
            Remove-Item "$DestinationPath\$($ExtractDir -replace '[\\/].*')" -Recurse -Force -ErrorAction Ignore
        }
```

This is what I mean.

niheaven:
For commit signing, you could ref https://docs.github.com/authentication/managing-commit-signature-verification

lib/decompress.ps1
niheaven: Since `extract_dir` could be a subdirectory such as `aaa\bbb`, `Expand-xxx` will delete `aaa` after extraction. Other functions that use a rarely used temporary directory may not encounter this bug, in my opinion.

o-l-a-v: `-le` (less than or equal) will return true if count is 0 too, but sure, I can change that.

I see you also removed `-Path`, why?

niheaven:
Hmm, you should try to pass the tests now :)

o-l-a-v:
I think I've fixed the test errors now @niheaven: Can't do `Get-ChildItem` if `-Path` does not exist.

lib/decompress.ps1
```diff
@@ -123,15 +123,21 @@ function Expand-7zipArchive {
...
-        # Remove temporary directory
-        Remove-Item "$DestinationPath\$($ExtractDir -replace '[\\/].*')" -Recurse -Force -ErrorAction Ignore
+        # Remove temporary directory if it still exists and is empty
+        $ExtractDirFullPath = [string] "$DestinationPath\$($ExtractDir -replace '[\\/].*')"
+        if (
+            (Test-Path -Path $ExtractDirFullPath -PathType 'Container') -and
+            (Get-ChildItem -Path $ExtractDirFullPath -Force).Count -eq 0
+        ) {
+            Remove-Item -Path $ExtractDirFullPath -Recurse -Force -ErrorAction Ignore
+        }
```
niheaven, line 133:
```suggestion
        $ExtractDirTopPath = "$DestinationPath\$($ExtractDir -replace '[\\/].*')"
        if ((Get-ChildItem -Path $ExtractDirTopPath -Force -ErrorAction Ignore).Count -eq 0) {
            Remove-Item -Path $ExtractDirTopPath -Recurse -Force -ErrorAction Ignore
        }
```

Change param name, and using `-ErrorAction Ignore` in `Get-ChildItem` will solve the prob (return $null that `Count` is 0 if Path is not existed).

olavrb-innovasjonnorge: That's just up to preference, but I'll comply with whatever Scoop wants.

lib/decompress.ps1
```diff
@@ -123,15 +123,18 @@ function Expand-7zipArchive {
     }
     if (!$IsTar -and $ExtractDir) {
         movedir "$DestinationPath\$ExtractDir" $DestinationPath | Out-Null
-        # Remove temporary directory
-        Remove-Item "$DestinationPath\$($ExtractDir -replace '[\\/].*')" -Recurse -Force -ErrorAction Ignore
+        # Remove temporary directory if it is empty
+        $ExtractDirTopPath = [string] "$DestinationPath\$($ExtractDir -replace '[\\/].*')"
+        if ((Get-ChildItem -Path $ExtractDirTopPath -Force -ErrorAction Ignore).Count -eq 0) {
+            Remove-Item -Path $ExtractDirTopPath -Recurse -Force -ErrorAction Ignore
+        }
```
z-Fng, line 130: While testing manifests, I happened to run into an issue here.

When the `$ExtractDir` depth is greater than or equal to three, like the one below (`$ExtractDir` : `keep\sub\sub-sub`), the folder 'keep' will not be removed. The return value of Get-ChildItem for folder 'keep' is always greater than 0 as there exists a sub-folder.

```
└─ keep
   └─ sub
      └─  sub-sub
         └─  empty
   └─ othersub
```

o-l-a-v: So one should look for files specifically? Add `-File -Recurse` to `Get-ChildItem`?

* <https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-childitem?view=powershell-7.5#-file>

z-Fng: > Add -File -Recurse to Get-ChildItem?

When extraction is done, we cannot tell whether the folder is newly created or already existed.
If the folder already exists before extraction, even if it is an empty folder, we should not remove it.

Perhaps we could check if the folder exists recursively before 7-Zip actually starts extracting?
```
keep
   └─ sub
      └─  sub-sub
```
1. Check if the folder `$DestinationPath\keep` exists.
    - If it does not exist, we can safely remove `$DestinationPath\keep`.
    - If it does exist, proceed to step 2.
2. Check if the folder `$DestinationPath\keep\sub` exists.
     - If it does not exist, we can safely remove `$DestinationPath\keep\sub`.
    - If it does exist, proceed to step 3.
3. ...

o-l-a-v: Can you create a new issue with an example manifest where this problem occurs?

There is also this feature request which would make this logic unnecessary:

* <https://github.com/ScoopInstaller/Scoop/issues/6093>

z-Fng: > Can you create a new issue with an example manifest where this problem occurs?

OK. I'm just about to do it.

> There is also this feature request which would make this logic unnecessary

It would be even better if this new feature fixes the issue.

z-Fng: > Can you create a new issue with an example manifest where this problem occurs?

- #6515

PR #6229, opened 2024-11-13
feat(cache): Add "tidy" as older cache removal command

### LINKED ISSUE #6068: [Feature] Remove older files from cache after installation
#### Problem to solve 
Files in the cache folder are never automatically removed. It makes it grow in size and many (most?) do not know about it or or how to clean it.

#### Solution I would like
I have already created a function to be called at the end of an installation. It removes cache files older than one month, but keeps at least the two most recent ones. I could also add a flag to skip this step (e.g. `—keep-cache`). We may also reverse this behavior by adding a flag that removes the older cache after installation, but it would not help cache discoverability.
See https://github.com/flodavid/Scoop/tree/feature/remove-old-cache

#### Alternatives I have  considered
There is no alternative today other than knowing about the cache and manually removing it.
We could also add a message at the end of an installation telling how much space is used by the cache and giving the command to clean it.

flodavid:
- Keep two items in cache for every app
- Files that are less than one month old are not deleted
- Automatically remove older cache of the app after an update
- Short option to do not use cache changed form `-k` to `-d` (for "download")
- Add option to keep older cache, with `-k` (for "keep") as short option to `update` command
- Add long option `--keep-cache` to `update` command

#### Description
The new `cache tidy` command enables to clean all packages older than one month, except the two most recent ones, of all the apps (with "*"), or a given one. This cache "tidying" is automatically ran after an app is updated, for this app only.
No option to change the one month period or the minimum number to keep is present at the moment.

#### Motivation and Context
Files in the cache folder were never automatically removed. It makes it grow in size and many (most?) do not know about it or or how to clean it. Besides, the only way to manage the cache was by removing all of it. 

* Closes [[Feature] Remove older files from cache after installation  #6068](https://github.com/ScoopInstaller/Scoop/issues/6068)

I have overwritten my local scoop scripts with the ones I modified in the repository.
I modified the date of a copy of a recent "which" package in cache with `(Get-Item which#2.16#607562b.zip).lastwritetime=$(Get-Date -Year 2020 -Month 12 -Day 31)`. Then I used the new tidy command with `scoop tidy which`. Only the package that is supposed to be older was removed.
Then I modified the version in the manifest of the "current" package of which, so that scoop suppose it is outdated. I did the same procedure of copying a recent package and editing its date. Then I ran `scoop update which`. "which" was reinstalled. Again, only the package that is supposed to be older was removed.

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 **autoupdate:** GitHub predefined hashes support ([#6416](https://github.com/ScoopInstaller/Scoop/issues/6416), [#6435](https://github.com/ScoopInstaller/Scoop/issues/6435))
 
+### BREAKING CHANGE
+
+- **scoop-install|update**: `-k`option that skipped cache and forced redownload is replaced by `-d`. `-k` now keeps older cache after app update
+
+### Features
+
+- **scoop-cache|update**: New tidy command that removes older packages from cache. Automatically run it after app update ([#6068](https://github.com/ScoopInstaller/Scoop/issues/6068))
+
 ### Bug Fixes
 
 - **scoop-download|install|update:** Fallback to default downloader when aria2 fails ([#4292](https://github.com/ScoopInstaller/Scoop/issues/4292))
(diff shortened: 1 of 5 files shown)


flodavid:
rebased onto develop

LukasWillin:
I was looking for a command like this. It would be nice to have a command which keeps 1 or 2 previous versions around in case some problems crop up with the current version after a little more use.

Thanks for the PR!

Issue #6103, opened 2024-08-14
[Bug] scoop status error


Jordan-Haidee:
#### Current Behavior

After running `scoop status`, it raise an error:
```txt
Where-Object: F:\scoop\apps\scoop\current\libexec\scoop-status.ps1:58
Line |
  58 |      Get-ChildItem $dir | Where-Object name -NE 'scoop' | ForEach-Obje …
     |                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | The 'Ine' operator failed: The term 'Get-UserAgent' is not recognized as a name of a cmdlet,
     | function, script file, or executable program. Check the spelling of the name, or if a path
     | was included, verify that the path is correct and try again..
```
#### Expected Behavior

No error message should raise, and status of installed softwares should be displayed normally.

#### Additional context/output

```
Where-Object: F:\scoop\apps\scoop\current\libexec\scoop-status.ps1:58
Line |
  58 |      Get-ChildItem $dir | Where-Object name -NE 'scoop' | ForEach-Obje …
     |                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | The 'Ine' operator failed: The term 'Get-UserAgent' is not recognized as a name of a cmdlet,
     | function, script file, or executable program. Check the spelling of the name, or if a path
     | was included, verify that the path is correct and try again..

Name         Installed Version           Latest Version           Missing Dependencies Info
----         -----------------           --------------           -------------------- ----
7zip         24.07                       24.08
bruno        1.23.1                      1.24.0
cmake        3.30.1                      3.30.2
everything   1.4.1.1024                  1.4.1.1026
eza          0.18.16                     0.19.0
ffmpeg       7.0                         7.0.2
flutter      3.22.3                      3.24.0
frp          0.58.0                      0.59.0
ghostscript  10.03.0                     10.03.1
git          2.45.2                      2.46.0
gitbutler    0.12.12                     0.12.19
imagemagick  7.1.1-32                    7.1.1-36
inkscape     1.3.1_2023-11-16_91b66b0783 1.3.2_2023-11-25_091e20e
just         1.29.1                      1.34.0
koodo-reader 1.6.6                       1.6.9
latexindent  3.24.1                      3.24.4
lua          5.4.2                       5.4.7-2
maven        3.9.7                       3.9.8
miniconda3   24.4.0-0                    24.5.0-0
mysql        8.4.0                       9.0.1
[...]

PR #6252, opened 2024-12-20
fix(scoop-status): do not loop through empty array; do not pass empty…

### LINKED ISSUE #6251: [Bug] Lack of argument checks leads to undefined behavior; Poor exception handling causes incorrect stack trace affecting `scoop status`
#### Current Behavior
This is a 2-in-1 issue. A lack of argument sanitation led to empty strings being passed around and a related error was being discard with an empty error thrown instead.

#### Expected Behavior
Arguements should be checks for validity and errors should not be caught if they'll be rethrown or discarded and replaced.

https://github.com/ScoopInstaller/Scoop/issues/6103#issuecomment-2556430131

#### Possible Solution

PR incoming...

- If the return value of Get-ChildItem is a zero-length array, do not loop through the array.
- Filter null, empty, and whitespace strings out of Get-ChildItem's return value
- Do not rethrow or discard caught errors.

### System details

**Windows version:** 11

**OS architecture:** 64bit

**PowerShell version:** 7.4.6

**Additional software:** git version 2.45.2.windows.1

#### Scoop Configuration

```json
{
  "last_update": "2024-12-19T22:08:00.6529901-08:00",
  "alias": {
    "upgrade": "scoop-upgrade",
    "outdated": "scoop-outdated"
  },
  "scoop_branch": "develop",
  "scoop_repo": "https://github.com/ScoopInstaller/Scoop"
}
```

BinToss:
#### Description

The first commit improves PowerShell's ability to trace a failed call to Get-UserAgent. It had noted Where-Object or Get-ChildItem calls in scoop-status.ps1 as the cause despite the error being thrown by manifest.ps1's `url_manifest` function.

Later commits improve error handling across the call chains and the last commit imports the `Get-UserAgent` function when it isn't already imported.

#### Motivation and Context

A lack of argument sanitation and an instance of bad error handling caused undefined behavior an error message that didn't make sense. As a result, the root cause was difficult to trace.

Fixes #6251, https://github.com/ScoopInstaller/Scoop/issues/6103

#### Checklist:


BinToss:
`scoop-status.sp1` does not yet have any tests and thorough testing would require a larger refactor. For now, I've added the test suite and one test for its one public function.

BinToss:
Is there a conventional-changelog config I'm unaware of? The changelog looks similar to the output of the Angular and Conventional Commits configs except for the inclusion commits that don't bump the version e.g. non-breaking refactors, docs, build, et cetera.

lib/manifest.ps1
```diff
@@ -14,14 +14,15 @@ function parse_json($path) {
 function url_manifest($url) {
     $str = $null
     try {
+        if (!(Get-Command Get-UserAgent -ErrorAction SilentlyContinue)) {
```
BinToss, line 17: Note: this `if` statement imports Get-UserAgent if a \*suspected\* race condition causes the function to be called before it's imported by a separate script. I don't know why so many of Scoop's scripts call functions without explicitly importing them beforehand. That's a refactor for another day.

lib/manifest.ps1
```diff
@@ -14,14 +14,15 @@ function parse_json($path) {
...
+            Import-Module "$PSScriptRoot/download.ps1" -Function 'Get-UserAgent'
+        }
```
BinToss, line 19: Note: this `if` statement imports Get-UserAgent if a \*suspected\* race condition causes the function to be called before it's imported by a separate script. I don't know why so many of Scoop's scripts call functions without explicitly importing them beforehand. That's a refactor for another day.

libexec/scoop-status.ps1
```diff
@@ -55,8 +55,10 @@ $true, $false | ForEach-Object { # local and global apps
     $dir = appsdir $global
     if (!(Test-Path $dir)) { return }
 
-    Get-ChildItem $dir | Where-Object name -NE 'scoop' | ForEach-Object {
-        $app = $_.name
+    [string[]] $appNames = Get-ChildItem $dir -Directory -Name -Exclude 'scoop' | Where-Object Length -GT 0
+    if ($appNames.Length -eq 0) { return }
+
+    foreach ($app in $appNames) {
```
BinToss, line 61: *   fix: refactor `Where-Object name -NE 'scoop'` to `Get-ChildItem`'s `Exclude` parameter. If the input object didn't have a `Name` parameter, an error would occur. This would happen when `Get-ChildItem` returns a zero-length array such as in the case of an empty or non-existent Global app dir.
*   fix: add `Where-Object Length -GT 0` to exclude zero-length strings from `$appNames`.
*   fix: assign app names array to `[string[]] $appNames`. Ensures the script halts if any other type is assigned to `$appNames` and prevents unexpected behavior caused by a type mismatch.
*   style: change `ForEach-Object` to `foreach`. A stylistic preference for the `$appName in $appNames` syntax over `$appName = $_`.

BinToss:
While refactoring loops, I realized PowerShell can be too forgiving with `return` vs `continue` in loops. Specifically, when the loop body is provided as a ScriptBlock. This led to bad practice on my part and I've identified places I should have used `continue`  used instead of `return` (which was breaking loops). Oops. I'll push the fix in a few minutes.

BinToss:
Current issue: local tests succeed whereas CI tests fail. `$ENV:SCOOP` is defined locally, but not in CI. This environment variable is necessary for bucket-related tests.

>  Describing Test-UpdateStatus
> Error: [-] Describe Test-UpdateStatus failed
> Message
>   ItemNotFoundException: Cannot find path 'D:\buckets' because it does not exist.
>   at Get-LocalBucket, D:\a\ScoopInstaller.Scoop\ScoopInstaller.Scoop\lib\buckets.ps1:64
>   at <ScriptBlock>, D:\a\ScoopInstaller.Scoop\ScoopInstaller.Scoop\libexec\scoop-status.ps1:38
>   at <ScriptBlock>, D:\a\ScoopInstaller.Scoop\ScoopInstaller.Scoop\test\Scoop-Status.Tests.ps1:20

I've decided to add a TODO note and a Skip condition. Note: this Skip condition is ignored by unsupported, legacy versions of Pester (e.g. the version included with Windows Powershell)

BinToss:
I've noticed some error exit codes after successful `scoop update *` ops. The console logs don't indicate any errors occurred. I'll investigate these error codes before this is merged.

BinToss:
This is on hold until I can come back to it. There's only a few edge-case errors I'd like to address before updating the change log.

One of the edge-case errors requires one or more of the following conditions:

*   One of the local buckets becomes corrupted (my `main` was broken. Probably caused by a BSOD occurring during a `git pull`).
*   The user tries to install/upgrade an application, but misspells the name

Issue #2124, opened 2018-03-14
Windows 10 does not display nirlauncher/sysinternals shortcuts in folder


rasa:
In Windows 7, the  nirlauncher and sysinternals packages nicely appear as folders in the "Scoop Apps" folder. On Windows 10, the folder names are ignored, resulting, and each package's entries appear intermixed with the other non-folder apps. I spent a couple of hours trying to fix this, but eventually gave up. Here are three possible solutions:

1. Instead of creating the shortcut folder under `Scoop Apps`, we create it in the same folder the `Scoop Apps` folder is created in. Elegant, but will users expect this, since other apps get installed in `Scoop Apps`. Perhaps we should simply do away with `Scoop Apps`?

2. We put the app's shortcuts in `%userprofile%\scoop\apps\app\current\shortcuts`, and create a single shortcut in `Scoop Apps` that opens an Explorer window on that directory. Kludgy.

3. Instead of creating `Nirlauncher\app` shortcuts, we create `zNirlaucher - app` shortcuts. This forces these multiple-app packages to appear at the end of the `Scoop Apps` list. Kludgy.

Feedback?

stkb:
Personally I've never really found it useful that the shortcuts are all put in the 'Scoop Apps` folder, so I'd do away with that.

However we'd then need to be careful that the scoop installer doesn't overwrite existing (non-scoop) shortcuts.

r15ch13:
@stkb That's why they get put in the 'Scoop Apps' folder 😁
But does it really matter? Do you really search a program in the menu manually? Typing in a programs name is much faster.

stkb:
@r15ch13 Yeah 99% of the time I type the name too rather than browse (so I'm not *too* bothered by this issue either) but I can see it being handy in some cases, especially with these packages.

To fix the collision issue, we could add a flag (probably in the description field) to shortcuts we create. Then if a shortcut already exists without our flag, we add `" (Scoop)"` to the name of our shortcut.

rasa:
How about we set the folder name as a config option:
```
scoop config startmenu_folder “Scoop Apps”
```
and default to “” for new installs?

Stanzilla:
Sounds good to me. The new convention is to not use folders in the start menu anyway.

rasa:
The Windows 10 April 2018 update hides the `Scoop Apps` folder. See https://github.com/lukesampson/scoop/issues/2233

kodybrown:
[...]

PR #6329, opened 2025-04-18
feat(config): Add configuration option for start menu shortcut location

dingdang66686:
#### Description

Set `start_menu_root` to `$true` for easier access to Scoop apps without navigating through a subfolder.

#### Motivation and Context
- By default, shortcuts are placed in the 'Scoop Apps' subfolder under 'Programs'. 
- This change allows users to access Scoop applications directly from the root of the 'Programs' menu, improving convenience for those who prefer a flatter menu structure.
- Users no longer need to drill down into a secondary folder, streamlining the app-launching experience.

Relates to #2124

I tested this change on my local machine and it works as expected.

#### Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added `start_menu_root` configuration option allowing users to choose whether shortcuts are created in the root Programs folder or the nested "Scoop Apps" subfolder, providing more flexibility in shortcut placement.


steawth:
bump, I would utilize this option lots

elisimpson:
I was just thinking of doing a PR for this issue. This solution would be adequate for my use case, but perhaps another way to do it would be instead of a boolean `start_menu_root` have a string `start_menu_folder_name` with the default being "Scoop Apps". A user wanting to use the main root folder would set the config string to "", which I've confirmed works with the current `Convert-Path` function.

Either way is fine with me as I just want to remove the folder entirely, but the string solution would allow someone to rename the folder to something other than "Scoop Apps" if they wanted.

One way or another, I'd love to have this change sooner rather than later. Right now I have to manually hack `lib/shortcuts.ps1` to get it the way I want.

PR #6373, opened 2025-05-27
fix(decompress): Wrap `$ExtractDir` in quotes

### LINKED ISSUE #6372: [Bug] `extract_dir` doesn't support spaces in path
#### Current Behavior
Spaces in `extract_dir` split a single argument into multiple arguments passed to 7z, which causes 7z to fail.

For example, `highlight` 4.16 requires `"extract_dir": "highlight-4.16 -x64"`:

```
"path\to\7z.exe"
"x"
"C:\Users\galaxysnail\scoop\apps\highlight\4.16\highlight-4.16-x64.zip"
"-oC:\Users\galaxysnail\scoop\apps\highlight\4.16"
"-xr!*.nsis"
"-y"
"-ir!highlight-4.16"
"-x64\*"

Command Line Error:
Incorrect wildcard type marker
64\*
```

#### Expected Behavior
The directory should be extracted successfully.

#### Possible Solution
Wrap the `extract_dir` argument in quotes.

### System details

**Windows version:** 10

**OS architecture:** 64bit

**PowerShell version:** `5.1.19041.5848`

**Additional software:** 7z

GalaxySnail:
#### Description
Support spaces in `extract_dir`.

#### Motivation and Context

Fix #6372

I tested it locally.

#### Checklist:

### DIFF
--- a/lib/decompress.ps1
+++ b/lib/decompress.ps1
@@ -96,7 +96,7 @@ function Expand-7zipArchive {
     $ArgList = @('x', $Path, "-o$DestinationPath", '-xr!*.nsis', '-y')
     $IsTar = ((strip_ext $Path) -match '\.tar$') -or ($Path -match '\.t[abgpx]z2?$')
     if (!$IsTar -and $ExtractDir) {
-        $ArgList += "-ir!$ExtractDir\*"
+        $ArgList += '"-ir!$ExtractDir\*"'
     }
     if ($Switches) {
         $ArgList += (-split $Switches)


o-l-a-v:
Shouldn't one put the actual path into quotes? Do something like this instead:

```pwsh
$ArgList += '-ir!"{0}\*"' -f $ExtractDir
```

Vs. what this PR suggests:

```pwsh
$ArgList += '"-ir!$ExtractDir\*"'
```

^ I also think this will not work, as `$variable` inside single quote `'` will not get the value of the variable.
 
* <https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-7.5#single-quoted-strings>

> A string enclosed in single quotation marks is a verbatim string. The string is passed to the command exactly as you type it. No substitution is performed.

z-Fng:
> I also think this will not work, as $variable inside single quote ' will not get the value of the variable.

Yes, a simple test can tell that this doesn't work.

```
scoop config scoop_repo https://github.com/GalaxySnail/Scoop.git
scoop config scoop_branch quote-extract_dir
scoop update
```

```
scoop install https://github.com/ScoopInstaller/Main/raw/0267d4145cdc0ba05fc3f4b5f5532cd682063cca/bucket/highlight.json
```

z-Fng:
Double quotes don’t work here either. All existing double quotes will be removed in Invoke-ExternalCommand.
https://github.com/ScoopInstaller/Scoop/blob/b588a06e41d920d2123ec70aee682bae14935939/lib/core.ps1#L708-L714

GalaxySnail:
> ^ I also think this will not work, as `$variable` inside single quote `'` will not get the value of the variable.
> 
> * [learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-7.5#single-quoted-strings](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-7.5#single-quoted-strings)
> 
> > A string enclosed in single quotation marks is a verbatim string. The string is passed to the command exactly as you type it. No substitution is performed.

You're right, this patch doesn't work. I must have mixed things up when I was trying different quoting methods and ended up submitting the wrong patch. I'm not very familiar with PowerShell; the key point here is figuring out how to properly quote arbitrary strings in PowerShell command arguments. It's much less obvious to me than in POSIX shells. 😢

z-Fng:
This PR should be closed, as it isn't working as expected. See: https://github.com/ScoopInstaller/Scoop/pull/6373#issuecomment-3397355326 & https://github.com/ScoopInstaller/Scoop/issues/6372#issuecomment-3397382813

PR #6383, opened 2025-06-08
fix(autoupdate): support non hash update manifest

cscnk52:
#### Motivation and Context

In current version of scoop, if one manifest want to bypass hash check, it should have `nightly` version, but in most cases, some manifest have proper version information, and it should bypass hash check too for some reason.

But current version of scoop will show an error when encounter this kind of situation like below:

```powershell
ERROR You cannot call a method on a null-valued expression.
```

This PR aim to fix this issue.

my test manifest:

https://github.com/cscnk52/cetacea/blob/53b0ffe93ed8d83f4b20d2af4a574d2e436040a1/bucket/todesk.json#L6-L10

Before:

```
Autoupdating todesk
Computed hash: 78bacfba23438526553960a45ef7dae817dab185421b00be33c8a70d155c1237
ERROR You cannot call a method on a null-valued expression.
```

After:

```
Autoupdating todesk
WARNING: Manifest for todesk does not contain a 'hash' field, skipping hash update.
Writing updated todesk manifest
```

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - **commands**: Handling broken aliases ([#6141](https://github.com/ScoopInstaller/Scoop/issues/6141))
 - **shim:** Do not suppress `stderr`, properly check `wslpath`/`cygpath` command first ([#6114](https://github.com/ScoopInstaller/Scoop/issues/6114))
 - **scoop-bucket:** Add missing import for `no_junction` envs ([#6181](https://github.com/ScoopInstaller/Scoop/issues/6181))
+- **autoupdate:** support non hash update manifest
 
 ### Code Refactoring
 
--- a/lib/autoupdate.ps1
+++ b/lib/autoupdate.ps1
@@ -366,7 +366,7 @@ function Update-ManifestProperty {
                     $newHash = HashHelper -AppName $AppName -Version $Version -HashExtraction $Manifest.autoupdate.hash -URL $newURL -Substitutions $Substitutions
                     $Manifest.hash, $hasPropertyChanged = PropertyHelper -Property $Manifest.hash -Value $newHash
                     $hasManifestChanged = $hasManifestChanged -or $hasPropertyChanged
-                } else {
+                } elseif ($Manifest.architecture.hash) {
                     # Arch-spec
                     $Manifest.architecture | Get-Member -MemberType NoteProperty | ForEach-Object {
                         $arch = $_.Name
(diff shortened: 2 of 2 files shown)


AkariiinMKII:
This is kind of risky, how about adding an `allow-non-hash` switch to scoop config in advance?

cscnk52:
@AkariiinMKII I don't think it's necessary to add a config for this option.
And there have proper warning when updating non hash manifest.

AkariiinMKII:
> @AkariiinMKII I don't think it's necessary to add a config for this option. And there have proper warning when updating non hash manifest.

Considering users may meet CDN pollutions, and this is what hash checking for, they need a way to interrupt the process, or confirmation before continue, not a message

AkariiinMKII:
> > Considering users may meet CDN pollutions, and this is what hash checking for, they need a way to interrupt the process, or confirmation before continue, not a message
> 
> Yes, you are right. But it is not broken to exist manifest with hash check, it is just valid when a manifest creator explicit remove `hash` in manifest.

Of course it is a good idea to allow skipping hash check and there already be situations that need this function, like https://github.com/ScoopInstaller/Nonportable/issues/52 
However, the manifest author could not guarantee what file you actually download without hash check, means **attacks may come from manifests by trusted author**, users should exactly know what they are doing and use this function at their own risk, according to their environments.

cscnk52:
> Of course it is a good idea to allow skipping hash check and there already be situations that need this function, like [ScoopInstaller/Nonportable#52](https://github.com/ScoopInstaller/Nonportable/issues/52) However, the manifest author could not guarantee what file you actually download without hash check, means **attacks may come from manifests by trusted author**, users should exactly know what they are doing and use this function at their own risk, according to their environments. It should not be decided by manifest author.

Yes, thanks for your reminder, in most cases, my bucket only used by me, I wasn’t careful enough. I will close this PR, thanks!

AkariiinMKII:
> > Of course it is a good idea to allow skipping hash check and there already be situations that need this function, like [ScoopInstaller/Nonportable#52](https://github.com/ScoopInstaller/Nonportable/issues/52) However, the manifest author could not guarantee what file you actually download without hash check, means **attacks may come from manifests by trusted author**, users should exactly know what they are doing and use this function at their own risk, according to their environments. It should not be decided by manifest author.
> 
> Yes, thanks for your reminder, in most cases, my bucket only used by me, I wasn’t careful enough. I will close this PR, thanks!

As I said it is actually a good idea that I believe many users need it.
Seems there needs a switch like `"non_hash": "true"` in `autoupdate` field to prevent generating hash value in auto-update process, and finally, let users decide whether to use by editing scoop config.
And there are several actions could be chosen like silently allow all, or ask for confirmation in each installation, or save the preference for each app in `install.json`...
Multiple function modules would be involved to implement this function, how about opening an issue first?

PR #6390, opened 2025-06-12
Update autoupdate.ps1 to drop the `sha256:` prefix of GitHub asset.digest

shenbo:
GitHub currently generates a corresponding SHA-256 hash for uploaded assets.

We can get the `asset.digest` (sha256 hash) by JSONPath fuction,  but the digest  has a `sha256:` prefix.

the function `find_hash_in_json([String] $url, [Hashtable] $substitutions, [String] $jsonpath)` in  `lib\autoupdate.ps1`  does not contain the regex function.

this pr  to  drop the `sha256:` prefix.


HUMORCE:
#6416

PR #6418, opened 2025-07-12
feat(install): Add separator at the end of notes, highlight suggestions

### LINKED ISSUE #6417: [Feature] Add separation between suggestions and notes, highlight suggestions
#### Command
`scoop install`
#### Problem
The suggestion and note sections lack proper visual separation, and suggestion section isn't highlighted, making it easy to be ignored.
#### Possible Solution
1. Add the same separator after the note section as before the note section.
2. Use different colors to display the text in the suggestion section.
#### Example
```console
scoop install games/prismlauncher
```
- Before: Image
- After: Image

z-Fng:
#### Description
- `feat(install)`: Add separator at the end of notes, highlight suggestions.

#### Motivation and Context
- Closes #6417 
The suggestion and note sections lack proper visual separation, and suggestion section isn't highlighted, making it easy to be ignored.
    - Before: Image
    - After: Image

#### Checklist:

## Summary by CodeRabbit

- New Features
  - Installer now adds a separator after displayed notes for clearer readability.
  - Unmet feature suggestions are highlighted in DarkYellow for improved visibility.

- Documentation
  - Updated changelog under Unreleased to reflect the new installer display enhancements.

### DIFF
--- a/lib/install.ps1
+++ b/lib/install.ps1
@@ -359,6 +359,7 @@ function show_notes($manifest, $dir, $original_dir, $persist_dir) {
         Write-Output 'Notes'
         Write-Output '-----'
         Write-Output (wraptext (substitute $manifest.notes @{ '$dir' = $dir; '$original_dir' = $original_dir; '$persist_dir' = $persist_dir }))
+        Write-Output '-----'
     }
 }
 
@@ -419,7 +420,7 @@ function show_suggestions($suggested) {
             }
 
             if (!$fulfilled) {
-                Write-Host "'$app' suggests installing '$([string]::join("' or '", $feature_suggestions))'."
+                Write-Host "'$app' suggests installing '$([string]::join("' or '", $feature_suggestions))'." -ForegroundColor DarkYellow
             }
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## [Unreleased](https://github.com/ScoopInstaller/Scoop/compare/v0.5.3...develop)
 
+### Features
+
+- **install:** Add separator at the end of notes, highlight suggestions ([#6418](https://github.com/ScoopInstaller/Scoop/issues/6418))
+
 ### Bug Fixes
 
 - **scoop-download:** Fix function `nightly_version` not defined error ([#6386](https://github.com/ScoopInstaller/Scoop/issues/6386))


HUMORCE:
The separator should only output when suggestion existed instead always, if we have to do this.

z-Fng:
Maybe my title is a bit misleading. I think the separator at the end of notes should always be displayed if possible.

This is because it not only separates notes from suggestions, but also separates notes from the output (or warnings) of other apps installed at the same time.

Here are some examples:
- Single app with only notes: `scoop install 1password-cli`
    - Before: image
    - After: image
- Single app with both notes and suggestions: `scoop install 1password-cli`
    - Before: image
    - After: image
- Multiple apps: `scoop install bind btop-lhm`
    - Before: image
    - After: image

The above are just my suggestions. (: There may be better ways to separate them.

> The separator should only output when suggestion existed instead always, if we have to do this.

For the situation with only notes, it is also acceptable, right?
image

HUMORCE:
We could remove `Notes` character, when it are wrapped by separator. In some cases, it does not a 'note'.

z-Fng:
> We could remove `Notes` character, when it are wrapped by separator. In some cases, it does not a 'note'.

That makes sense to me. Simply highlighting suggestions would be helpful as well.

lib/install.ps1
```diff
@@ -419,7 +420,7 @@ function show_suggestions($suggested) {
             }
 
             if (!$fulfilled) {
-                Write-Host "'$app' suggests installing '$([string]::join("' or '", $feature_suggestions))'."
+                Write-Host "'$app' suggests installing '$([string]::join("' or '", $feature_suggestions))'." -f DarkYellow
```
niheaven: Use fullname of args, and add a changelog entry.

z-Fng: Got it. I'll handle it right away.

z-Fng: Done.

z-Fng:
> HUMORCE:
> We could remove Notes character, when it are wrapped by separator. In some cases, it does not a 'note'.

> coderabbitai:
> 358-364: Optional: drop the “Notes” label when wrapped by separators
Per review thread, removing the header reduces redundancy when the block is delimited above and below.

As mentioned above, is it better to remove 'Notes' characters when they are wrapped by separators? Like:
```diff
-Write-Output 'Notes'
Write-Output '-----'
Write-Output (wraptext (substitute $manifest.notes @{ '$dir' = $dir; '$original_dir' = $original_dir; '$persist_dir' = $persist_dir }))
Write-Output '-----'
```

Or make 'Notes' characters wrapped by separators too? Like this:

```powershell
 Write-Output '-----'
 Write-Output 'Notes:'
 Write-Output (wraptext (substitute $manifest.notes @{ '$dir' = $dir; '$original_dir' = $original_dir; '$persist_dir' = $persist_dir }))
 Write-Output '-----'
```

niheaven:
Keep it as is now imo. A `---` without title looks strange, right?

z-Fng:
> Keep it as is now imo. A --- without title looks strange, right?

Agree. This looks more natural imo.
- Keep it as is now
image

- Remove 'Notes' characters
image

- Wrap 'Notes' by separators
image

niheaven (approved):

PR #6439, opened 2025-08-13
fix(scoop-update): Force sync tags w/ remote branch while scoop update

### LINKED ISSUE #6438: [Bug] Re-tagging release breaks scoop
#### Current Behavior

If a Scoop release tag is mutated it blocks the update of Scoop until the tag is manually removed from the local machine. This isn't obvious as the error only says _Update failed._.

#### Expected Behavior

I'd expect Scoop release tags to be immutable for all of the supply chain reasons as well as to not break end-users. I'd also expect the error message in the case where the tag is blocking a Scoop update to be meaningful. If Git is intended to replace the tags this needs to be explicitly configured for the Scoop Git repo and shouldn't rely on Git global defaults.

#### Possible Solution

Manually remove the tag from Scoop.

niheaven:
#### Description

Force sync tags while running `scoop update` to update Scoop Core.

#### Motivation and Context

- Closes #6438

#### Checklist:

## Summary by CodeRabbit

- Bug Fixes
  - scoop update now force-syncs tags with the remote, ensuring all tags are fetched and updated during the operation.

- Documentation
  - Added an Unreleased changelog entry under Bug Fixes noting the tag force-sync behavior in scoop update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [Unreleased](https://github.com/ScoopInstaller/Scoop/compare/v0.5.3...develop)
+
+### Bug Fixes
+
+- **scoop-update**: Force sync tags w/ remote branch while scoop update ([#6439](https://github.com/ScoopInstaller/Scoop/issues/6439))
+
 ## [v0.5.3](https://github.com/ScoopInstaller/Scoop/compare/v0.5.2...v0.5.3) - 2025-08-11
 
 ### Features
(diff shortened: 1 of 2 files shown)


HUMORCE:
Should we implement similar changes to bucket updates?

niheaven:
A bucket should not have `tag`, imo.

HUMORCE:
I mean something like hard reset when conflicting.

niheaven:
Stash? If someone made their own change.

z-Fng:
Would it be possible to merge these hotfixes (#6439, #6454) and release a new version soon?

It’s been a month since the last release.

niheaven:
I'll check for #6454 soon.

Issue #6277, opened 2025-01-21
[Bug] Scoop warns about overwriting shim scoop.ps1


w4tchdoge:
#### Current Behavior

I am getting the following `WARN` messages when updating scoop using `scoop update`
```console
WARN  Overwriting shim ('scoop.ps1' -> 'scoop.ps1') installed from scoop
WARN  Overwriting shim ('scoop.cmd' -> 'scoop.ps1') installed from scoop
WARN  Overwriting shim ('scoop' -> 'scoop.ps1') installed from scoop
```

#### Expected Behavior

I expect to not get the `WARN` messages when updating scoop

#### Additional context/output

I have recently reinstalled scoop on this device using the advanced install method with `-ScoopDir 'D:\Scoop'`. The ScoopDir of the previous installation was `C:\-0PFiles\Scoop` which has since been deleted.

### System details

**Windows version:** 11

**OS architecture:** 64bit

**PowerShell version:** 7.4.6

#### Scoop Configuration

```json
{
  "last_update": "2025-01-22T06:20:47.9371099+08:00",
  "scoop_repo": "https://github.com/ScoopInstaller/Scoop",
  "scoop_branch": "master",
  "root_path": "D:\\Scoop\\",
  "alias": {
    "infov": "scoop-infov",
    "ssearch": "scoop-ssearch"
  },
}
```

david-bakin:
Looks like this on my machine - a brand new out-of-the-box (but updated) Windows 11 `[Version 10.0.26100.3194]` where I just installed Scoop with the local dir as `C:\Scoop\Local` and the global dir as `C:\Scoop\Global` and this was the first update about an hour later:

```
C:\>scoop update
Updating Scoop...
Remove-Item : An object at the specified path C:\Scoop\Local\shims\scoop.ps1. does not exist.
At C:\Scoop\Local\apps\scoop\current\lib\core.ps1:943 char:13
+             Remove-Item -Path "$shim.$path_app" -Force -ErrorAction S ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Remove-Item], PSArgumentException
    + FullyQualifiedErrorId : Argument,Microsoft.PowerShell.Commands.RemoveItemCommand

WARN  Overwriting shim ('scoop.ps1' -> 'scoop.ps1') installed from scoop
[...]

Issue #6194, opened 2024-10-17
[Bug] An object at the specified path ... ... does not exist.


ETime-Github:
#### Current Behavior

When update scoop saied scoop.ps1 does not exist. But it actually exists.
I update about once a week and it prompts this error.
If I update a few times in a short period of time, it won't get an error, but I haven't tested how long the intervals are.
#### Expected Behavior

No error.
#### Additional context/output

```powershell
 $ scoop update *
Updating Scoop...
Remove-Item: Scoop\apps\scoop\current\lib\core.ps1:943
Line |
 943 |              Remove-Item -Path "$shim.$path_app" -Force -ErrorAction S …
     |              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | An object at the specified path Scoop\shims\scoop.ps1. does not exist.
WARN  Overwriting shim ('scoop.ps1' -> 'scoop.ps1') installed from scoop
Remove-Item: Scoop\apps\scoop\current\lib\core.ps1:943
Line |
 943 |              Remove-Item -Path "$shim.$path_app" -Force -ErrorAction S …
     |              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | An object at the specified path Scoop\shims\scoop.cmd. does not exist.
WARN  Overwriting shim ('scoop.cmd' -> 'scoop.ps1') installed from scoop
Remove-Item: Scoop\apps\scoop\current\lib\core.ps1:943
Line |
 943 |              Remove-Item -Path "$shim.$path_app" -Force -ErrorAction S …
     |              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | An object at the specified path Scoop\shims\scoop. does not exist.
WARN  Overwriting shim ('scoop' -> 'scoop.ps1') installed from scoop
```
![image](https://github.com/user-attachments/assets/bcf365da-e25e-4829-859d-7f5833e39485)
![image](https://github.com/user-attachments/assets/1f6590be-9693-41d6-bbf4-c7f5722404eb)

### System details

**Windows version:** [e.g. 7, 8, 10, 11]
11
**OS architecture:** [e.g. 32bit, 64bit, arm64]
64bit
[...]

PR #6485, opened 2025-09-16
fix(scoop-update): normalize `$currentdir` to prevent unnecessary overwrite warning ScoopInstaller/Scoop#6277

zodiacg:
To fix unnecessary overwrite warning mentioned in ScoopInstaller/Scoop#6277

#### Description

The `$currentdir` variable in `scoop-update.ps1` is not normalized. Under certain settings (e.g. path with a trailing '\' in root_path config), it will cause unnecessary warnings about overwriting scoop shims during update.

For example:
* If the `root_path` config has a trailing `\`, the `$currentdir` variable will carry a double `\` when passed to `shim` function.
* The `warn_on_overwrite` checks the shim file and path to compare app names. Due to the double slashes, the `get_app_name` function will failed to match the correct app name (scoop in this case), and return a `''`. Thus causing the warning.

#### Motivation and Context

Relates to #6277 or #6194.

It fixes my errors but I'm not quite sure about other possible causes.

Anyway, normalizing a path variable should be considered good practice regardless.

#### Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path handling in the update process by using absolute paths, preventing failures caused by relative path resolution.
  * Ensures reliable detection of the current installation and associated repository, regardless of the working directory.
  * Reduces errors in file operations during updates (e.g., checking repository state and related actions).

### DIFF
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -75,7 +75,7 @@ function Sync-Scoop {
     if (!(Test-GitAvailable)) { abort "Scoop uses Git to update itself. Run 'scoop install git' and try again." }
 
     Write-Host 'Updating Scoop...'
-    $currentdir = versiondir 'scoop' 'current'
+    $currentdir = Convert-Path (versiondir 'scoop' 'current')
     if (!(Test-Path "$currentdir\.git")) {
         $newdir = "$currentdir\..\new"
         $olddir = "$currentdir\..\old"


niheaven:
Thanks for your contribution~ But I found the bug is caused by `Get-AbsolutePath()` and was fixed in #6501, please test that one.

PR #6490, opened 2025-09-18
fix(checkver): Allow script to run when URL fetch fails but script exists

### LINKED ISSUE #5704: [Bug] homepage should not be downloaded when checkver script is specified

### LINKED ISSUE #6480: [Bug] For the autoupdate of the manifest, it should not be interrupted just because the homepage is invalid.
#### Current Behavior

- Currently, for the autoupdate of the manifest, it will first check if the homepage is valid. Only if it is valid will the autoupdate logic proceed.
- Is this really reasonable?
- Let's take an example:

  - If the software is hosted on GitHub but has its own homepage: a.com
  - The bucket maintainer uses a.com as the homepage
  - Later, a.com may have access issues (such as DNS resolution problems, domain name changes, etc.)
  - Because the homepage becomes invalid, the automatic update logic will be interrupted
  - Unless the bucket maintainer updates it in a timely manner, the manifest will never be able to complete the automatic update
  - This is entirely dependent on the timely update by the bucket maintainer. Is this reasonable?

#### Possible Solution

- Ignore the case where the homepage is invalid and continue executing the autoupdate logic.

z-Fng:
#### Description
- `fix(checkver)`: Allow script to run when URL fetch fails but script exists.

#### Motivation and Context
- Closes #5704
- Closes #6480
- Closes #6274

Failing to fetch homepage/checkver.url prevents checkver.script from running.

Sometimes, the homepage is unreachable duo to anti-bot mechanisms or IP blocks targeting GitHub's hosted runner networks.

For this situation, we have to change the homepage, or use a fake url as checkver.url.
e.g.
- https://github.com/abgox/abyss/commit/604f294e6e9b6b12a5d4606e2faafdef2a9c1269
- https://github.com/Calinou/scoop-games/blob/master/bucket/epic-games-launcher.json
```json
    "checkver": {
        "url": "https://scoop.sh",
        "useragent": "$app/$version",
        "script": [
            "$redirUrl = [Net.HttpWebRequest]::Create('https://launcher-public-service-prod06.ol.epicgames.com/launcher/api/installer/download/EpicGamesLauncherInstaller.msi')",
            "$redirUrl.method = 'head'",
            "[Web.HttpUtility]::ParseQueryString($redirUrl.GetResponse().ResponseUri.Query).Get('launcherfilename')"
        ],
        "regex": "\\AEpicInstaller-([\\d.]+)\\.msi\\Z"
    }
```

Tested it locally with [Calinou/scoop-games/epic-games-launcher.json](https://github.com/Calinou/scoop-games/blob/master/bucket/epic-games-launcher.json)(checkver.url removed).
- Before: image
- After: image

#### Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * When a URL fetch fails, provided fallback scripts are now run so checks continue instead of being skipped; downloaded content is only processed on successful fetches.
  * Logs and source labels now reflect when content comes from fallback output to avoid misleading results.

* **New Features**
  * Added a fallback flow that uses script-derived content on download failures while preserving normal behavior otherwise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - **buckets|scoop-info:** Switch git log date format to ISO 8601 to avoid locale issues ([#6446](https://github.com/ScoopInstaller/Scoop/issues/6446))
 - **scoop-version:** Fix logic error caused by missing brackets ([#6463](https://github.com/ScoopInstaller/Scoop/issues/6463))
 - **core|manifest:** Avoid error messages when searching non-existent 'deprecated' directory ([#6471](https://github.com/ScoopInstaller/Scoop/issues/6471))
+- **checkver:** Allow script to run when URL fetch fails but script exists ([#6490](https://github.com/ScoopInstaller/Scoop/issues/6490))
 
 ### Code Refactoring
 
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -281,11 +281,16 @@ while ($in_progress -gt 0) {
         }
         $err = $ev.SourceEventArgs.Error
         if ($err) {
-            next "$($err.message)`r`nURL $url is not valid"
-            continue
+            if (!$script) {
+                next "$($err.message)`r`nURL $url is not valid"
+                continue
+            } else {
+                # Run script despite URL download failure
+                Write-Host "$($err.message)`r`nURL $url is not valid. Falling back to checkver.script ..."
+            }
         }
 
-        if ($url) {
+        if ($url -and !$err) {
             $ms = New-Object System.IO.MemoryStream
             $ms.Write($result, 0, $result.Length)
             $ms.Seek(0, 0) | Out-Null


z-Fng:
cc: @niheaven Sorry to ping you here, requesting a code review for this PR.

niheaven (dismissed):
I'd like to add a new `method` in checkver and unify such situation. But as a temp solution, it's okay.

Please add a changelog entry.

z-Fng:
> I'd like to add a new `method` in checkver and unify such situation.

That's a great idea! Maybe we can make use of this in a future refactor.

> Please add a changelog entry.

OK. I'll update it right away.

z-Fng:
> Please add a changelog entry.

Done.

niheaven (approved):

PR #6494, opened 2025-09-19
`formatjson.ps1` - Sort manifest root keys by `schema.json` and child keys alphabetically

### LINKED ISSUE #6491: [Feature] `formatjson.ps1`: Sort manifest root properties
### Context

We now have linting on manifest PRs by running `formatjson.ps1` and comparing the result.

* <https://github.com/ScoopInstaller/GithubActions/pull/61>

The next step in automating more and putting less load on maintainers and reviewers is to sort the object of root elements / first level elements in the manifest JSON files. This was first discussed here:

* Issue: <https://github.com/ScoopInstaller/GithubActions/issues/47>
* PR: <https://github.com/ScoopInstaller/GithubActions/pull/50>

Discussions resulted in Scoop wanting this functionality in [formatjson.ps1](https://github.com/ScoopInstaller/Scoop/blob/master/bin/formatjson.ps1) instead. That is what this feature request is about.

### Is your feature request related to a problem? Please describe

Scoop has a preferred order for JSON root elements in manifests, but it is not enforced or automated.

### Describe the solution you'd like

* A "master" file containing all manifest properties **and** the wanted order of those.
  * This file can then be referenced by other logic.
  * [schema.json](https://github.com/ScoopInstaller/Scoop/blob/master/schema.json) does not adhere to the wanted order of elements.
      <details><summary>Click to expand</summary>

      ```pwsh
      PS > (Invoke-RestMethod -Method 'Get' -Uri 'https://github.com/ScoopInstaller/Scoop/raw/refs/heads/master/schema.json').'properties'

      $schema        : @{type=string; format=uri}
      _comment       : @{description=Deprecated. Use ## instead.; $ref=#/definitions/stringOrArrayOfStrings}
      ##             : @{description=A comment.; $ref=#/definitions/stringOrArrayOfStrings}
      architecture   : @{additionalProperties=False; properties=; type=object}
      autoupdate     : @{$ref=#/definitions/autoupdate}
      bin            : @{$ref=#/definitions/stringOrArrayOfStringsOrAnArrayOfArrayOfStrings}
      persist        : @{$ref=#/definitions/stringOrArrayOfStringsOrAnArrayOfArrayOfStrings}
      checkver       : @{$ref=#/definitions/checkver}
      cookie         : @{description=Undocumented: Found at https://github.com/se35710/scoop-java/search?l=JSON&q=cookie; type=object}
      depends        : @{$ref=#/definitions/stringOrArrayOfStrings}
      description    : @{type=string}
      env_add_path   : @{$ref=#/definitions/stringOrArrayOfStrings}
      env_set        : @{type=object}
      extract_dir    : @{$ref=#/definitions/stringOrArrayOfStrings}
      extract_to     : @{$ref=#/definitions/stringOrArrayOfStrings}
      hash           : @{$ref=#/definitions/hash}
      homepage       : @{format=uri; type=string}
      innosetup      : @{description=True if the installer InnoSetup based. Found in https://github.com/ScoopInstaller/Main/search?l=JSON&q=innosetup; type=boolean}
      installer      : @{$ref=#/definitions/installer}
      license        : @{$ref=#/definitions/license}
      notes          : @{$ref=#/definitions/stringOrArrayOfStrings}
      post_install   : @{$ref=#/definitions/stringOrArrayOfStrings}
      post_uninstall : @{$ref=#/definitions/stringOrArrayOfStrings}
      pre_install    : @{$ref=#/definitions/stringOrArrayOfStrings}
      pre_uninstall  : @{$ref=#/definitions/stringOrArrayOfStrings}
      psmodule       : @{additionalProperties=False; properties=; type=object}
      shortcuts      : @{$ref=#/definitions/shortcutsArray}
      suggest        : @{additionalProperties=False; patternProperties=; type=object}
      uninstaller    : @{$ref=#/definitions/uninstaller}
      url            : @{$ref=#/definitions/uriOrArrayOfUris}
      version        : @{pattern=^[\w\.\-+_]+$; type=string}

      PS >
      ```

      </details>
  * I think this is the properties and wanted order?
      <details><summary>Click to expand</summary>

      ```text
      $schema
      _comment
      ##
      version
      description
      homepage
      license
      notes
      depends
      suggest
      url
      hash
      architecture
      innosetup
      extract_dir
      extract_to
      pre_install
      installer
      post_install
      env_add_path
      env_set
      bin
      shortcuts
      persist
      pre_uninstall
      uninstaller
      post_uninstall
      psmodule
      checkver
      autoupdate
      cookie
      ```

      </details>

* A new function in [json.ps1](https://github.com/ScoopInstaller/Scoop/blob/master/lib/json.ps1), maybe `Sort-ScoopManifestRootProperties`
* Call new function from [formatjson.ps1](https://github.com/ScoopInstaller/Scoop/blob/master/bin/formatjson.ps1).

Example logic for sorting JSON by wanted order, works with Windows PowerShell 5.1 and PowerShell 7.4, 7.5.

```pwsh
# Wanted order
$WantedOrder = [string[]](
    'key3',
    'key1',
    'key2'
)

# Sample JSON with wrong order
$SampleJson = [string](
    ConvertTo-Json -InputObject (
        [PSCustomObject]@{
            'key2' = [string] 'value2'
            'key3' = [bool] $false
            'key1' = [PSCustomObject]@{
                'key11' = [byte] 1
            }
        }
    )
)

# Read sample JSON
$From = [PSCustomObject](ConvertFrom-Json -InputObject $SampleJson)

# Generate sorted output
$To = [PSCustomObject]::new()
$From.'PSObject'.'Properties'.'Name' | Sort-Object -Property @{'Expression' = {[byte]($WantedOrder.IndexOf($_))}} | ForEach-Object -Process {
    $null = Add-Member -InputObject $To -NotePropertyName $_ -NotePropertyValue $From.$_
}
ConvertTo-Json -InputObject $To

```

o-l-a-v:
#### Description

1. Reordered root properties in `schema.json` so it matches the wanted order.
1. Created new functions in `manifest.ps1`
   * `Sort-ScoopManifestProperties`:
       * Sorts root properties according to `schema.json` properties.
       * Sorts one level deep child properties alphabetically if the parent property is of type `PSCustomObject`.
       * Called by `formatjson.ps1`.
   * `Sort-PSCustomObjectKeysRecursively`:
      * Uses recursion to sort all child keys alphabetically, does not touch other values.
      * Called by `Sort-ScoopManifestProperties` for manifest 1 level child properties that are of type `[PSCustomObject]`.
1. Made `formatjson.ps1` use new function `Sort-ScoopManifestProperties`.

#### Motivation and Context

Closes #6491.

```pwsh
# Path to formatjson.ps1 in branch "formatjson-sort-root-properties"
$FunctionPath = [string] 'D:\git\ScoopInstaller--Scoop\bin\formatjson.ps1'

# Path to main bucket in branch "master" pulled 2025-09-19T12:25:00+02:00
$Directory = [string] 'D:\git\ScoopInstaller--Main\bucket'

# Run format json, look at git afterwards
$ErrorActionPreference = 'Stop'
. $FunctionPath -Dir $Directory -App '*'
```

image

#### Checklist:

   * Info about `formatjson.ps1` was added with <https://github.com/ScoopInstaller/GithubActions/pull/61>: <https://github.com/ScoopInstaller/GithubActions/wiki/Pull-Request-Checks>

## Summary by CodeRabbit

* **New Features**
  * Manifest formatter now deterministically orders root and first-level properties; added two manifest utilities to support recursive and root-level ordering.
  * Schema expanded with new top-level fields and updated required fields (version, homepage, license).

* **Refactor**
  * More robust manifest parsing and safer formatting pipeline that writes back only when output is valid.

* **Tests**
  * Added test to ensure empty-path JSON parsing does not throw.

* **Documentation**
  * CHANGELOG updated.


o-l-a-v:
@deevus @HUMORCE @rashil2000 @z-Fng: What do you guys think?

o-l-a-v:
@coderabbitai review

bin/formatjson.ps1
```diff
@@ -34,11 +34,21 @@ param(
...
+    $json = [PSCustomObject](Sort-ScoopManifestRootProperties -JsonAsObject $json)
+
+    # Beautify
+    $json = [string](ConvertToPrettyJson -data $json)
+
+    # Convert to 4 spaces
+    $json = [string]($json -replace "`t", '    ')
+
+    # Overwrite file content
     [System.IO.File]::WriteAllLines($file, $json)
```
z-Fng: When JSON parsing fails, it will clear the manifest content. Should this behavior be adjusted to prevent contributors from losing their work?

bin/formatjson.ps1
```diff
@@ -34,11 +34,21 @@ param(
...
+    # Path of file
+    $file = [string] $_.'FullName'
 
-    # convert to 4 spaces
-    $json = $json -replace "`t", '    '
+    # Parse JSON
+    $json = [PSCustomObject](parse_json -path $file)
+
+    # Sort JSON root properties
+    $json = [PSCustomObject](Sort-ScoopManifestRootProperties -JsonAsObject $json)
```
z-Fng: Is it possible to sort the keys recursively with global/local order?

bin/formatjson.ps1
```diff
@@ -34,11 +34,21 @@ param(
...
+
+    # Beautify
+    $json = [string](ConvertToPrettyJson -data $json)
```
z-Fng, line 47: I am concerned whether the sort is a stable sort algorithm for keys that are not present in the index list.

Could multiple executions yield different results?

o-l-a-v: IMO there could and should be a check for keys we don't want to be present. Doesn't this exist already?

To make formatjson.ps1 safer we could do: If key is not present in the schema.json

* Optian a) fail execution
* Option b) remove key

bin/formatjson.ps1
o-l-a-v: Yup, maybe we should add some error handling.

bin/formatjson.ps1
o-l-a-v: I don't fully understand what you're asking for here, about the local/global order.

Recursice ordering keys by the schema.json should be possible. Or just sort child keys alphabetically.

z-Fng: > local/global order

My description might be redundant. Jsut ignore it. Focus on recursive sorting.

I thought that when sorting keys like architecture.64bit/32bit , local order are used because the configuration file contains architecture.64bit/32bit. However, when sorting keys like architecture.64bit.url/hash, the configuration file lacks this entry, causing it to fall back to the global order.

As I'm not sure how to describe this, I called it  local/global order.

https://github.com/ScoopInstaller/Scoop/blob/37aa4596d5f1e5b8d914786b6f038e597ac52960/schema.json#L217-L231

bin/formatjson.ps1
z-Fng: > Doesn't this exist already?

Does it? It will fail execution currently.
```
Sort-Object: D:\Scoop\apps\scoop\current\lib\json.ps1:241:9
Line |
 241 |          Sort-Object -Property @{
     |          ~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot convert value "-1" to type "System.Byte". Error: "Value was either too large or too small for an unsigned byte."
```

o-l-a-v: I was thinking more about that other parts of the validation should catch manifest errors before it gets to `formatjson.ps1`. Thus validating that the manifest doesn't contain invalid keys is a task that `formatjson.ps1` shouldn't try to solve.

o-l-a-v: Seems that's what the validator is for?

<https://github.com/ScoopInstaller/Scoop/tree/master/supporting/validator>

Thus I don't think the new function `Sort-ScoopManifestRootProperties` should have a lot of validation logic. It's a feature that it throws if manifest does not comply with `schema.json` IMO.

bin/formatjson.ps1
o-l-a-v: I added sorting of first level child keys if parent is of type `PSCustomObject`. Going even deeper will be more complex. It's doable, but I think what we got now is a good start.

bin/formatjson.ps1
o-l-a-v: I added a check for parent keys not present in `schema.json` inside `Sort-ScoopManifestRootProperties`. Then users will get a better error message than what you posted in <https://github.com/ScoopInstaller/Scoop/pull/6494#discussion_r2365560938>.

bin/formatjson.ps1
o-l-a-v: Maybe there are existing PowerShell or C# projects that can sort a JSON file by `schema.json` and handle all the complexity of going deeper we can use?

In Python there is: <https://github.com/ikonst/jschon-sort>.

But as I already said, I think this serves as a very good start for getting more uniform manifests.

bin/formatjson.ps1
o-l-a-v: Or we could write a warning, and just return the object as is / not try to sort the object at all. Instead of throwing.

z-Fng: > Seems that's what the validator is for?

Yes, thanks for reminding.


> I added a check for parent keys not present in `schema.json` inside `Sort-ScoopManifestRootProperties`. Then users will get a better error message than what you posted.

Totally agree. Great.

z-Fng:
@coderabbitai review

lib/json.ps1
```diff
@@ -212,3 +212,65 @@ function normalize_values([psobject] $json) {
...
+    <#
+        .SYNOPSIS
+            Sort JSON root properties according to schema.json, and level one child properties alphabetically.
+    #>
+    [OutputType([PSCustomObject])]
+
+    Param(
+        [Parameter(Mandatory)]
+        [PSCustomObject] $JsonAsObject
+    )
```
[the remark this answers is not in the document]
o-l-a-v: Added validation as `[ValidateScript({})]`.

o-l-a-v:
I implemented some of the nitpicks from CodeRabbit.

* CHANGELOG: Fixed.

> bin/formatjson.ps1 (2): Use WriteAllText for a single string payload

I tested this, this does not write the final newline. So I kept it to use WriteAllLines.

> lib/json.ps1: Cache wanted order; avoid hidden dependency on parse_json

Scoop does this many other places, uses it's own functions. I can change this if maintainers want it, but that would not fit with what existing code already does.

> Avoid [byte] cast; use int to prevent overflow and past -1 issues

It's highly unlikely that the Scoop manifest will ever get more than 255 unique root properties. But I changed it to uint16 just in case, which will do 65535.

lib/manifest.ps1
```diff
@@ -212,3 +218,106 @@ function uninstaller($manifest, $arch) { arch_specific 'uninstaller' $manifest $
...
+    # Get wanted order from Scoop manifest schema
+    if ([string]::IsNullOrWhiteSpace($Script:WantedOrder)) {
+        $Script:WantedOrder = [string[]](
+            (
+                parse_json -path (
+                    '{0}\..\schema.json' -f $PSScriptRoot
+                )
+            ).'properties'.'PSObject'.'Properties'.'Name'
+        )
+    }
```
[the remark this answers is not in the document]
o-l-a-v: > IndexOf on a string[] isn’t an instance method; use [Array]::IndexOf(...)

This is not true. `[string[]]` also has the `.IndexOf()` method.

```pwsh
PS > $([string[]]('0','1','2')).IndexOf('1')
1
PS >
```

lib/manifest.ps1
```diff
@@ -2,10 +2,16 @@ function manifest_path($app, $bucket) {
     (Get-ChildItem (Find-BucketDirectory $bucket) -Filter "$(sanitary_path $app).json" -Recurse).FullName
 }
 
-function parse_json($path) {
-    if ($null -eq $path -or !(Test-Path $path)) { return $null }
+function parse_json {
+    Param(
+        [Parameter(Mandatory)]
```
z-Fng: Adding `[Parameter(Mandatory)]` changes the original logic. I encountered the following error while testing.
```console
> scoop download uninstalled-app
WARNING: Searching manifest: uninstalled-app, main
parse_json: D:\Scoop\apps\scoop\current\lib\manifest.ps1:123
Line |
 123 |      parse_json (manifest_path $app $bucket)
     |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot bind argument to parameter 'path' because it is an empty string.
WARNING: Searching manifest: uninstalled-app, extras
parse_json: D:\Scoop\apps\scoop\current\lib\manifest.ps1:123
Line |
 123 |      parse_json (manifest_path $app $bucket)
     |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot bind argument to parameter 'path' because it is an empty string.
```

o-l-a-v: Good find, I'll revert.

Can this be added as an automated test?

z-Fng: Scoop seems to support unit tests only and does not support integration tests currently.

Perhaps we can add tests for `parse_json` in the unit tests?

bin/formatjson.ps1
z-Fng: > Going even deeper will be more complex. It's doable, but I think what we got now is a good start.

Yes. It is a good start. The changes made so far are getting a bit complex. Thankfully, it's just a tool.

> Maybe there are existing PowerShell or C# projects that can sort a JSON file by schema.json and handle all the complexity of going deeper we can use?

Unfortunately, I haven't been able to find any yet.

z-Fng:
> I thought that when sorting keys like architecture.64bit/32bit , local order are used because the configuration file contains architecture.64bit/32bit. However, when sorting keys like architecture.64bit.url/hash, the configuration file lacks this entry, causing it to fall back to the global order.

As I mentioned earlier, this is the code implementation of my previous idea. Can this help us go further in a simpler way?

Just a thought.

```powershell
function Get-RootPropertiesObject {
    [OutputType([PSCustomObject])]
    param()

    if (-not $Script:RootPropertiesObject) {
        $Script:RootPropertiesObject = [PSCustomObject](
            (
                parse_json -path (
                    '{0}\..\schema.json' -f $PSScriptRoot
                )
            ).'properties'
        )
    }

    return $Script:RootPropertiesObject
}

function Sort-Properties {
    param(
        [Parameter(Mandatory)]
        $Properties,
        [Parameter(Mandatory = $false)]
        [PSCustomObject] $PropertiesObjectDefault = (Get-RootPropertiesObject),
        [Parameter(Mandatory = $false)]
        [PSCustomObject] $PropertiesObjectOverride = $null
    )

    if ($Properties -is [PSCustomObject]) {

        $OrderDefault = $PropertiesObjectDefault.'PSObject'.'Properties'.'Name'
        $OrderOverride = $PropertiesObjectOverride.'PSObject'.'Properties'.'Name'

        [PSCustomObject] $PropertiesObject = if ($PropertiesObjectOverride) {
            $PropertiesObjectOverride
        } else {
            $PropertiesObjectDefault
        }

        $PropertyNames = $Properties.'PSObject'.'Properties'.'Name'

        $SortedProperties = [PSCustomObject]::new()

        $SortedPropertyNames = @($PropertyNames.Where{ $_ -in $OrderOverride } | Sort-Object { $OrderOverride.IndexOf($_) }) +
                               @($PropertyNames.Where{ ($_ -notin $OrderOverride) -and ($_ -in $OrderDefault)} | Sort-Object { $OrderDefault.IndexOf($_) }) +
                               @($PropertyNames.Where{ ($_ -notin $OrderOverride) -and ($_ -notin $OrderDefault)})

        $SortedPropertyNames | ForEach-Object {
            $SortedProperty = Sort-Properties -Properties $Properties.$_ -PropertiesObjectDefault $PropertiesObjectDefault -PropertiesObjectOverride $PropertiesObject.$_.'properties'

            if ($SortedProperty) {
                Add-Member -InputObject $SortedProperties -NotePropertyName $_ -NotePropertyValue $SortedProperty | Out-Null
            }
        }

        if ((-not $SortedProperties.PSObject.Properties) -or ($SortedProperties.PSObject.Properties.Count -eq 0)) {
            $SortedProperties = $null
        }

        return $SortedProperties
    }

    if ($Properties -is [array]) {
        $SortedProperties = @()

        for($i = 0; $i -lt $Properties.Count; ++$i) {
            $SortedProperties += @(Sort-Properties -Properties $Properties[$i] -PropertiesObjectDefault $PropertiesObjectDefault)
        }

        return ,$SortedProperties
    }

    return $Properties
}

function Get-SortedManifest {
    [OutputType([PSCustomObject])]

    param(
        [Parameter(Mandatory)]
        [ValidateScript({ $null -ne $_ -and $_ -ne [PSCustomObject]::new() })]
        [PSCustomObject] $Manifest
    )

    Sort-Properties -Properties $Manifest -PropertiesObjectDefault (Get-RootPropertiesObject)
}
```

o-l-a-v:
<https://github.com/ScoopInstaller/Scoop/pull/6494#issuecomment-3329194281>

I don't see how that is simpler. Also the first function won't, as PSCustomObject.properties won't output anything. Is it LLM generated?

z-Fng:
> I don't see how that is simpler.

Possibly more comprehensive.

> PSCustomObject.properties won't output anything

It's actually 'properties' in schema.json.

> Is it LLM generated?

Yes, after editing and modifying it, it runs fine on my local machine.

o-l-a-v:
> > I don't see how that is simpler.

> Possibly more comprehensive.

That's subjective. I think not using aliases, using explicit types, and using named parameters with functions are more readable, maintainable. But I'm down with whatever Scoop maintainers want, as long as we can get the functionality we want here. 😊

>> PSCustomObject.properties won't output anything

> It's actually 'properties' in schema.json.

Isn't it `<output_from_parse_json>.PSObject.Properties.Name` you want?

>> Is it LLM generated?

> Yes, after editing and modifying it, it runs fine on my local machine.

Ok.

abgox:
- There is a similar implementation in my [abyss](https://github.com/abgox/abyss) bucket: https://github.com/abgox/abyss/blob/main/script/sort-json.ps1
- It has already been working in workflows and performing well.
- It is not associated with the Schema file but instead defines the order directly in the script, which reduces some complexity.
- Maybe it can serve as a reference.

Issue #6425, opened 2025-08-03
[Bug] checkver: Incorrect version returned when script fails without output


z-Fng:
#### Current Behavior
When performing version detection (_checkver.ps1_) for multiple manifests, if checkver.script **produces no output**, it may return an incorrect version.

#### How to Reproduce
```powershell
$manifests = @(
    @{
        script = "Invoke-WebRequest -Uri 'https://www.qt.io/offline-installers' -UseBasicParsing| Select-Object -ExpandProperty Content"
        regex  = 'Qt Creator\s+([\d.]+)\s+for Windows'
    }
    @{
        script = '1/0'
        regex  = 'blender-([\w.]+)-windows-x64\.zip'
    }
)

foreach ($manifest in $manifests) {
    $re = New-Object System.Text.RegularExpressions.Regex($manifest.regex)

    $page = Invoke-Command ([scriptblock]::Create($manifest.script))

    $match = $re.Matches($page) | Select-Object -First 1

    Write-Host "Found: $match `r`n"
}
```

#### Cause Analysis
PowerShell appears to have the following behavior:
> Even when a function produces no output, PowerShell returns an empty PSObject as a placeholder to keep the return value consistent.

This can be observed by running:
```powershell
function Test-NotReturn { }

function Test-ReturnNULL { return $null }

foreach ($result in @($null, (Test-NotReturn), (Test-ReturnNULL))) {
    Write-Host "Result is `$null: $($null -eq $result)"
    Write-Host "Result is string: $($result -is [string])"
    Write-Host "Result is PSObject: $($result -is [PSObject])"
    Write-Host ''
}
```

<img width="1220" height="451" alt="Image" src="https://github.com/user-attachments/assets/31e6c6af-04de-435f-81db-75105ce717d7" />

<img width="1220" height="587" alt="Image" src="https://github.com/user-attachments/assets/98d6cefa-5553-4dc6-a983-19f431d15dd5" />

Errors inside `Invoke-Command` may produce no standard output, causing `$page` to be assigned an empty `PSObject`.
Calling `$re.Matches($page)` later throws an exception since `$page` is an object, leaving `$match` from the previous iteration intact.

#### Possible Solution
1. Properly handle the outputs of `Invoke-Command`.
2. Ensure the `$match` variable is initialized before use.

z-Fng:
Reproduce with simple scripts:
```powershell
$testcases = @(
    @{
        app   = "QTCreater"
[...]

PR #6547, opened 2025-11-12
fix(checkver): Fix incorrect version returned when script fails without output

z-Fng:
## Description
Migrate from #6489.

#### Motivation and Context
Closes:
- #6425
- https://github.com/ScoopInstaller/GithubActions/issues/36

Relats to:
- #4941
- #6489

#### Changes
- `fix(checkver)`: Fix incorrect version returned when checkver.script fails without output.

#### Tests
- Before: https://github.com/z-Fng/Test-Scoop-checkver/actions/runs/19346533804/job/55348214134
> other-pkg: 18.0.0 (scoop version is 1.0.0) autoupdate available
- After: https://github.com/z-Fng/Test-Scoop-checkver/actions/runs/19365821642/job/55408575924
> other-pkg: couldn't retrieve content from the output of script

#### Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where version detection could return an incorrect value when a probe or script failed without producing output.
  * Improved isolation between individual checks so results from prior attempts no longer affect subsequent version detection, reducing false positives after errors.

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -274,6 +274,8 @@ while ($in_progress -gt 0) {
     $expected_ver = $json.version
     $ver = $Version
 
+    $matchesHashtable = @{}
+
     if (!$ver) {
         if (!$regexp -and $replace) {
             next "'replace' requires 're' or 'regex'"
@@ -303,6 +305,11 @@ while ($in_progress -gt 0) {
         if ($script) {
             $page = Invoke-Command ([scriptblock]::Create($script -join "`r`n"))
             $source = 'the output of script'
+        }
+
+        if ($null -eq $page) {
+            next "couldn't retrieve content from $source"
+            continue
         }
 
         if ($jsonpath) {
@@ -363,7 +370,6 @@ while ($in_progress -gt 0) {
             }
 
             if ($match -and $match.Success) {
-                $matchesHashtable = @{}
                 $re.GetGroupNames() | ForEach-Object { $matchesHashtable.Add($_, $match.Groups[$_].Value) }
                 $ver = $matchesHashtable['1']
                 if ($replace) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - **core|manifest:** Avoid error messages when searching non-existent 'deprecated' directory ([#6471](https://github.com/ScoopInstaller/Scoop/issues/6471))
 - **path:** Trim ending slash when initializing paths ([#6501](https://github.com/ScoopInstaller/Scoop/issues/6501))
 - **checkver:** Allow script to run when URL fetch fails but script exists ([#6490](https://github.com/ScoopInstaller/Scoop/issues/6490))
+- **checkver:** Fix incorrect version returned when script fails without output ([#6547](https://github.com/ScoopInstaller/Scoop/issues/6547))
 - **uninstall:** Import `url_filename` from `download.ps1` ([#6530](https://github.com/ScoopInstaller/Scoop/issues/6530))
 - **schema:** Add missing `hash.mode` value `github` ([#6533](https://github.com/ScoopInstaller/Scoop/issues/6533))
 - **core:** Skip NO_JUNCTION logic when $app is 'scoop' in `currentdir` function ([#6541](https://github.com/ScoopInstaller/Scoop/issues/6541))


bin/checkver.ps1
```diff
@@ -302,6 +307,12 @@ while ($in_progress -gt 0) {
         $source = $url
         if ($script) {
             $page = Invoke-Command ([scriptblock]::Create($script -join "`r`n"))
+
+            # When Invoke-Command produces no output, $page is assigned an empty PSObject. Assign it to an empty string instead.
+            if ([string]::IsNullOrWhiteSpace($page)) {
+                $page = [string]::Empty
+            }
+
```
z-Fng: I'm not sure if these lines of code are necessary. 

Theoretically, all functions that don't produce output could have this problem.

The important part is to initialize `$match` so that leftover values from the previous iteration aren't accidentally used if an exception occurs during matching.

Alternatively, we could have exception handling for matching, though I think initializing the variable is a better approach.

niheaven: Does `$matchHashtable` or `$page` introduce the bug? I think only one should be initialed in each loop.

Oh, `$match` is initialed below too...

z-Fng: > Does `$matchHashtable` or `$page` introduce the bug?
1. It's actually caused by `$match` not being initialized. Maybe I should simplify the changes a bit.
  https://github.com/ScoopInstaller/Scoop/blob/b588a06e41d920d2123ec70aee682bae14935939/bin/checkver.ps1#L299
  When `Invoke-Command` produce no output, `$page` will be assigned an empty `PSObject` instead of `$null`.
    https://github.com/ScoopInstaller/Scoop/blob/b588a06e41d920d2123ec70aee682bae14935939/bin/checkver.ps1#L355
  Calling `$re.Matches($page)` later throws an exception since `$page` is an object, leaving `$match` from the previous iteration intact.

2. It looks like `$matchesHashtable` also has the risk of being reused across iterations, so I updated it as well.
https://github.com/ScoopInstaller/Scoop/blob/b588a06e41d920d2123ec70aee682bae14935939/bin/checkver.ps1#L352-L374
https://github.com/ScoopInstaller/Scoop/blob/b588a06e41d920d2123ec70aee682bae14935939/bin/checkver.ps1#L411

niheaven: If `$match` is initialized in each loop, the `PSObject` type of `$page` won't matter, and the same goes for `$matchHashtable`, right?

z-Fng: Yes, exactly. I'll minimize the changes. 

Edit: Completed and tests have passed.

> #### Tests
> - Before: https://github.com/z-Fng/Test-Scoop-checkver/actions/runs/19346533804/job/55348214134
> > other-pkg: 18.0.0 (scoop version is 1.0.0) autoupdate available
> - After: https://github.com/z-Fng/Test-Scoop-checkver/actions/runs/19346631777/job/55348541521
> > 'scoop_repo' has been set to 'https://github.com/z-Fng/Scoop.git'
> > 'scoop_branch' has been set to 'fix-checkver'
> 
> > other-pkg: couldn't match 'Qt Creator\s+([\d.]+)\s+for Windows' in the output of script

niheaven:
The problem is all about a `$null` value of `$page`, so an early return when `$page` is `$null` should be more proper.

Add the following script near line 309 should fix the bug.

```powershell
        if ($null -eq $page) {
            next "Couldn't download $url or execute script"
            continue
        }
```

z-Fng:
> The problem is all about a `$null` value of `$page`, so an early return when `$page` is `$null` should be more proper.

Yes, I totally overlooked that. Changes applied.

> #### Tests
> - After: https://github.com/z-Fng/Test-Scoop-checkver/actions/runs/19365821642/job/55408575924
> > 'scoop_repo' has been set to 'https://github.com/z-Fng/Scoop.git'
> > 'scoop_branch' has been set to 'fix-checkver'
> 
> > other-pkg: couldn't retrieve content from the output of script


I think initializing `$match` and `$matchesHashtable` is still useful, too. Is it okay if we keep that part as well?

niheaven:
> I think initializing $match and $matchesHashtable is still useful, too. Is it okay if we keep that part as well?

The `$matchesHashtable` might end up with the wrong value during the loop, but `$match` should be fine, it gets reassigned (even if the reassignment fails) before it's actually used.

z-Fng:
Yes, that's right. Thanks for pointing it out. Changes applied.

niheaven (approved):

PR #6556, opened 2025-11-17
fix(checkver): Prevent $page from carrying over between iterations

z-Fng:
#### Description
- fix(checkver): Prevent $page from carrying over between iterations.

#### Motivation and Context
This PR serves as a patch for PR #6490.

If `$url` fails to fetch, the `$page` value may persist across iterations and be used by `checkver.script`.

- Relates to: #5080
- Relates to: #6547
- Relates to: #6490

I've tested this locally. I'm not sure if adding a test case is necessary.

#### Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved download vs. script fallback flow for more reliable content retrieval
  * Clarified source and error messages to better reflect whether content came from a URL or script
  * Enhanced error-checking to reduce false failures during retrieval

* **Chores**
  * Updated changelog entry with an additional cross-reference

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 - **getopt:** Teach getopt to respect the `--%` token ([#6477](https://github.com/ScoopInstaller/Scoop/issues/6477))
 - **core|manifest:** Avoid error messages when searching non-existent 'deprecated' directory ([#6471](https://github.com/ScoopInstaller/Scoop/issues/6471))
 - **path:** Trim ending slash when initializing paths ([#6501](https://github.com/ScoopInstaller/Scoop/issues/6501))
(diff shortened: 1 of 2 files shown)


niheaven:
The operators `-not` and `!` are equivalent in PowerShell, however, `!` is more intuitive to use.

z-Fng:
> The operators `-not` and `!` are equivalent in PowerShell, however, `!` is more intuitive to use.

I feel that `-not` is more intuitive, aligns better with `-eq`, and matches the general PowerShell style, which is why I changed it initially. But that's fine -- keeping the changes minimal is also the right approach for a PR, I think.

Changes have been applied as suggested.

niheaven (approved):

PR #6557, opened 2025-11-17
Add file lock to prevent concurrent executions

KoltesDigital:
#### Description

This patch adds a file lock to ensure only one process uses Scoop. Extra processes wait 1 second and retry, so dequeuing is not handled in order.

The diff shows a lot of modified lines, but it's actually just an indent increase due to the try-finally block.

#### Motivation and Context

I've created a scheduled task to automate `scoop update *` every morning or at every startup. Whether it's good or bad is another topic :) But I've noticed some conflicts. For instance, `scoop list` shows `install failed` while the app is being installed. I suspect that I got corrupted git-buckets due to parallel updates too.

This is also in line with my expectations of a service. IMHO Scoop acts like a session-wide service that several processes can access, so it must ensure it's safe to do so. The fact that Scoop exclusively works with PS1 scripts is an implementation detail.

Note that global installs made by different process owners could still conflicts. A better version would be to pinpoint which commands are subjects to lock, and in these cases whether the lock should be global. This would also have the benefit not to lock for things like displaying help. I'd love to have that, but there's more effort to put, so please tell me what you think first.

Locally with multiple shells.

#### Checklist:

- [X] I have added an entry in the CHANGELOG.

## Summary by CodeRabbit

* **New Features**
  * Adds a file locking mechanism to prevent concurrent executions; callers see a one-time waiting message if a run is already active.

* **Bug Fixes**
  * Ensures the lock is always released after execution.
  * Unknown or unsupported commands now emit a clear warning and exit with a non-zero status.

* **Chores**
  * Updated ignore rules to exclude lock files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.log
 .DS_Store
 ._.DS_Store
+.lock
 scoop.sublime-workspace
 test/installer/tmp/*
 test/tmp/*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - **install:** Add separator at the end of notes, highlight suggestions ([#6418](https://github.com/ScoopInstaller/Scoop/issues/6418))
 - **download|scoop-download:** Add GitHub issue prompt when the default downloader fails ([#6539](https://github.com/ScoopInstaller/Scoop/issues/6539))
 - **download|scoop-config:** Allow disabling automatic fallback to the default downloader when Aria2c download fails ([#6538](https://github.com/ScoopInstaller/Scoop/issues/6538))
+- **core:** Add file lock to prevent concurrent executions ([#6557](https://github.com/ScoopInstaller/Scoop/issues/6557))
 
 ### Bug Fixes
 
--- a/bin/scoop.ps1
+++ b/bin/scoop.ps1
@@ -6,48 +6,74 @@ Set-StrictMode -Off
 . "$PSScriptRoot\..\lib\commands.ps1"
 . "$PSScriptRoot\..\lib\help.ps1"
 
-$subCommand = $Args[0]
+$lock_file_path = "$PSScriptRoot\..\.lock"
+$lock_stream = $null
+$lock_show_waiting_message = $true
 
-# for aliases where there's a local function, re-alias so the function takes precedence
-$aliases = Get-Alias | Where-Object { $_.Options -notmatch 'ReadOnly|AllScope' } | ForEach-Object { $_.Name }
-Get-ChildItem Function: | Where-Object -Property Name -In -Value $aliases | ForEach-Object {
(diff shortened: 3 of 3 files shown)


rashil2000:
Question: is there a shared persistent state that Scoop reads/writes when updating or installing apps? As far as I know Scoop was one of the few apps that supported doing multiple installs/updates (of different apps of course) at the same time. 

Running a scheduled task for update is fine, but what's the use case of running `list` or `status` while the apps are getting updated?

KoltesDigital:
@rashil2000 more on the context, to be honest... and also to get feedback from you :)

Thought I'm a developer, I'm dealing with IT for a small company, so I'm trying to use my dev habits to manage PCs. Workstations run Windows, and we have a couple of in-house apps. I've chosen to install them and keep them updated through Scoop. Installation itself is automatized through Intune depending on groups the user or machine belong to, but it can't handle updates. So the first thing Intune does is install Scoop, and then Intune only install apps through Scoop. This allows user to install extra apps. Intune also installs a scheduled task as described, to get things up-to-date. We're not big enough to get fixed, verified versions of apps... so I prefer to always have the latest versions, with the tiny risk that an app may break.

Last detail: Intune can install its things as device or as user, and it's kind of a nightmare regarding permissions. When as device, install process belongs to SYSTEM account, and is elevated. When as user, install process belongs to the user, and is automatically elevated iff the user is admin. This last point is really painful: user context is not the same for all users. For instance, when Intune installs Scoop for admin users, buckets (Git repos) somehow get wrong permissions, and non-elevated shells cannot use Scoop anymore... so there are some tricks to get back on track.

Because Scoop takes care of updates, I don't have to deal with that on my apps, and that's great. Whenever CICD pushes a new version, `scoop update ...` does the job. However, my users won't run Scoop. That's why the scheduled task runs it every morning. However, we can have several updates per day, most of them are bugfixes to the previous one, so we can't wait the next morning. So in the main app, there's a regular check to its manifest (directly through https) to see whether there's a new version, and if so, offers to the user to run `scoop update <app>` under the hood. Well I know, it's not far away from reimplementing a whole update process... but at least Scoop does a part of that.

Many users reported many times that the app couldn't launch anymore. My assumption is that they launched the app right at startup, the app tells them to update it, so they do it, but the scheduled task is also doing it, and boom. Thankfully, the fix is simply uninstall and reinstall.

`list` or `status` were just examples, they show an incorrect state but it's harmless. Running concurrent `update`s is not. If you're willing to consider a dedicated lock for each app, that would be awesome. Please tell me and I can try to do it. BTW I've put the lock at the repo's root, but if we're getting a bunch of them, another location surely is better, please tell me an advice.

rashil2000:
This should ideally be done on a per-app basis. The `SCOOP/apps/<app-name>/` is one location to place the lock in. 

We can also show the status 'Updating' in the `list` or `status` command if the lock file exists for that app.

Issue #6588, opened 2026-01-23
[Bug] difficulty downloading 7zip


jennerboi:
# Bug Report

#### Current Behavior
every time i try to download 7zip it gives me a message that "a politic for program control has blocked this file"

#### Expected Behavior
it would just download 7zip

#### Additional context/output
WARN  Purging previous failed installation of lessmsi.
ERROR 'lessmsi' isn't installed correctly.
'lessmsi' was uninstalled.
Installing 'lessmsi' (2.11.7) [64bit] from 'main' bucket
Loading lessmsi-v2.11.7.zip from cache
Checking hash of lessmsi-v2.11.7.zip ... ok.
Extracting lessmsi-v2.11.7.zip ... done.
Linking ~\scoop\apps\lessmsi\current => ~\scoop\apps\lessmsi\2.11.7
Creating shim for 'lessmsi'.
Creating shortcut for Less MSIérables (lessmsi-gui.exe)
'lessmsi' (2.11.7) was installed successfully!
Installing '7zip' (25.01) [64bit] from 'main' bucket
Loading 7z2501-x64.msi from cache
Checking hash of 7z2501-x64.msi ... ok.
Extracting 7z2501-x64.msi ... ERROR Exception calling "Start" with "0" argument(s): "En politik for programkontrol har blokeret denne fil"
The expression after '&' in a pipeline element produced an object that was not valid. It must result in a command name,
 a script block, or a CommandInfo object.
At C:\Users\jens116p\scoop\apps\scoop\current\lib\core.ps1:242 char:18
+         return & $git @ArgumentList
+                  ~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : BadExpression

The expression after '&' in a pipeline element produced an object that was not valid. It must result in a command name,
 a script block, or a CommandInfo object.
At C:\Users\jens116p\scoop\apps\scoop\current\lib\core.ps1:260 char:14
+     return & $git @ArgumentList
+              ~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : BadExpression

Failed to extract files from C:\Users\jens116p\scoop\apps\7zip\25.01\7z2501-x64.msi.
Log file:
  ~\scoop\apps\7zip\25.01\msi.log

Please try again or create a new issue by using the following link and paste your console output:
https:////
PS C:\Users\jens116p>
### System details
[...]

Issue #6562, opened 2025-11-28
[Bug] Scoop fails to install 7zip with error 1601 - Windows 2025 Server


oswcab:
#### Current Behavior
When trying to install 7zip as a regular user it fails with this error:

```
scoop install git
Installing '7zip' (25.01) [64bit] from 'main' bucket
7z2501-x64.msi (1.9 MB) [=============================================================================================] 100%
Checking hash of 7z2501-x64.msi ... ok.
Extracting 7z2501-x64.msi ... ERROR Exit code was 1601!
The expression after '&' in a pipeline element produced an object that was not valid. It must result in a command name, a script block, or a CommandInfo object.
At C:\Users\konflux-builder\scoop\apps\scoop\current\lib\core.ps1:242 char:18
+         return & $git @ArgumentList
+                  ~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : BadExpression

Failed to extract files from C:\Users\konflux-builder\scoop\apps\7zip\25.01\7z2501-x64.msi.
Log file:
  ~\scoop\apps\7zip\25.01\msi.log

Please try again or create a new issue by using the following link and paste your console output:
https:////
```

Referred log, `scoop\apps\7zip\25.01\msi.log`, is empty.

If scoop is installed as Administrator, it works without a problem, but we're trying scoop exactly because we want to give an user the right to install some dependencies for his build system.

Running `scoop checkup` as a regular user showed two warnings: Long paths and developer mode.
As an admin, run these two commands to fix that:

```
Set-ItemProperty 'HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem' -Name 'LongPathsEnabled' -Value 1
```
and
```
$RegPath = "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\AppModelUnlock"
Set-ItemProperty -Path $RegPath -Name "AllowDevelopmentWithoutDevLicense" -Value 1 -Type DWORD -Force
```

but this still does not help.

To avoid any possible permission issue with the Temp folder:

```
$TempPath = "C:\Users\konflux-builder\Temp"
New-Item -ItemType Directory -Force -Path $TempPath
[Environment]::SetEnvironmentVariable("TEMP", "$TempPath", "User")
[Environment]::SetEnvironmentVariable("TMP", "$TempPath", "User")
```

Now `scoop checkup` only reports:

```
[...]

Issue #4735, opened 2022-02-12
[Bug] Cannot install 7zip and git


gsantovena:
#### Current Behavior
When installing git, scoop is trying to install 7zip but fails because git is needed.

#### Expected Behavior
Install 7zip and git via scoop without issues.

#### Additional context/output
```
PS C:\Windows\system32> sudo scoop install 7zip git --global
Installing '7zip' (21.07) [64bit]
Loading 7z2107-x64.msi from cache
Checking hash of 7z2107-x64.msi ... ok.
Extracting 7z2107-x64.msi ... ERROR Exit code was 1603!
git : The term 'git' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try
again.
At line:1 char:1
+ git config --get remote.origin.url
+ ~~~
    + CategoryInfo          : ObjectNotFound: (git:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

Failed to extract files from C:\ProgramData\scoop\apps\7zip\21.07\7z2107-x64.msi.
Log file:
  C:\ProgramData\scoop\apps\7zip\21.07\msi.log

Please try again or create a new issue by using the following link and paste your console output:
https:////
```

### System details

**Windows version:** 10

**OS architecture:** 64bit

**PowerShell version:** 5.1.14393.4583

**Additional software:** [(optional) e.g. ConEmu, Git]

#### Scoop Configuration

```json
PS C:\Windows\System32\config\systemprofile\.config\scoop> cat .\config.json
{
    "lastupdate":  "2022-02-12T05:16:05.5201691+00:00",
    "7ZIPEXTRACT_USE_EXTERNAL":  false
}```

**Comment by rashil2000** [comment]
Paste the contents of log file

**Comment by gsantovena** [comment]
> Paste the contents of log file

```
PS C:\> cat C:\Windows\system32\config\systemprofile\scoop\apps\7zip\21.07\msi.log
=== Logging started: 2/12/2022  5:55:32 ===
The installer has encountered an unexpected error installing this package. This may indicate a problem with this package. The error code is 2103. The arguments are: 26, ,
=== Logging stopped: 2/12/2022  5:55:33 ===
```

**Comment by rashil2000** [comment]
Can you try doing a local (non-global) install?

**Comment by gsantovena** [comment]
Same error:
```
PS C:\> scoop install 7zip
[...]

PR #6591, opened 2026-01-29
fix(buckets): Skip Git invocation if unavailable in `new_issue_msg`

z-Fng:
#### Motivation and Context
  Fix issues that occur when Git is not installed:
  - `new_issue_msg` throws a error when trying to invoke Git commands, leading to confusing error messages and incorrect issue links.
    e.g. https://github.com/ScoopInstaller/Scoop/issues/6588
    https://github.com/ScoopInstaller/Scoop/issues/4735
    https://github.com/ScoopInstaller/Scoop/issues/6562

    > Please try again or create a new issue by using the following link and paste your console output:
    > https:////
  - Before:image
  - After: image

#### Changes:
  - `lib/buckets.ps1`: Check `Test-GitAvailable` before invoking Git in `new_issue_msg`.

#### Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced robustness by gracefully handling scenarios where Git is unavailable when creating issue reports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Bug Fixes
 
 - **core:** Fix the grep parameter in the `Invoke-GitLog` function ([#6407](https://github.com/ScoopInstaller/Scoop/issues/6407))
+- **buckets:** Skip Git invocation if unavailable in `new_issue_msg` ([#6591](https://github.com/ScoopInstaller/Scoop/issues/6591))
 - **buckets:** Fix the filtering condition when retrieving the number of manifests ([#6509](https://github.com/ScoopInstaller/Scoop/issues/6509))
 - **scoop-download:** Fix function `nightly_version` not defined error ([#6386](https://github.com/ScoopInstaller/Scoop/issues/6386))
 - **scoop-download:** Fix incorrect download success state ([#6473](https://github.com/ScoopInstaller/Scoop/issues/6473))
--- a/lib/buckets.ps1
+++ b/lib/buckets.ps1
@@ -190,7 +190,7 @@ function new_issue_msg($app, $bucket, $title, $body) {
     $url = known_bucket_repo $bucket
     $bucket_path = "$bucketsdir\$bucket"
 
-    if (Test-Path $bucket_path) {
+    if ((Test-Path $bucket_path) -and (Test-GitAvailable)) {
         $remote = Invoke-Git -Path $bucket_path -ArgumentList @('config', '--get', 'remote.origin.url')
         # Support ssh and http syntax
         # git@PROVIDER:USER/REPO.git


29039:
The "scoop install git" command works just fine. Because the main bucket is already included with scoop as a normal non-git folder, which in return contains the manifest to install git. Once the main bucket is updated after fit is installed, the main bucket also gets converted to git.

I'm not saying that is necessarily the best solution. I would prefer it if git (and 7zip) was always installed and the main bucket was synced with git in the first place.

But it's not broken by default when scoop is installed.

If your git it broken, better to reinstall scoop.

This proposed solution is pure AI slop which does nothing.

The only change I'd advocate for is for either git (and 7zip) to be installed by default, or git (and 7zip) to be installed automatically once a command requiring git is run, rather than throwing an error and having to install it manually.

z-Fng:
> The "scoop install git" command works just fine. Because the main bucket is already included with scoop as a normal non-git folder, which in return contains the manifest to install git. Once the main bucket is updated after fit is installed, the main bucket also gets converted to git.

> This proposed solution is pure AI slop which does nothing.

I think you missed the point of what I'm actually fixing here, and it's very rude of you to call this 'AI slop' before you've even figured out what problem I'm solving.

> I'm not saying that is necessarily the best solution. I would prefer it if git (and 7zip) was always installed and the main bucket was synced with git in the first place.
> 
> But it's not broken by default when scoop is installed.
> 
> If your git it broken, better to reinstall scoop.
> 

> The only change I'd advocate for is for either git (and 7zip) to be installed by default, or git (and 7zip) to be installed automatically once a command requiring git is run, rather than throwing an error and having to install it manually.

Scoop supports using external Git and 7zip installations. The assumption that 'git (and 7zip) was always installed by scoop' simply doesn't hold true for all users. Real-world scenarios are more complex than that—some errors even occur during installation, and the issue I'm addressing causes the error messages to be confusing. Reinstalling might clear the state, but it doesn't fix the logic gap that causes this confusing error in the first place.

29039:
My apologies, I completely misunderstood what you were trying to achieve, and for my tone.

I thought that you were actually trying to remediate the fact that git is not installed by default and that coderabbit make a fix which didn't actually do that.

Actually you were just making some edge case error messages more user friendly.

Which is a noble goal, but why not be more ambitious to actually provide a remediation for the elephant in the room that scoop doesn't install git automatically when needed?

Just seems like extra work for the user, and doesn't seem that hard to implement.

maybe something like this?

update the Test-GitAvailable in lib/core.ps1:

```
function Test-GitAvailable {
    # Check if git is available
    if ([Boolean](Get-HelperPath -Helper Git)) {
        return $true
    }

    # Git not found - attempt auto-install
    warn "Git is required but not installed. Installing git..."
    & "$PSScriptRoot\..\libexec\scoop-install.ps1" 'git'

    # Check again after install
    if ([Boolean](Get-HelperPath -Helper Git)) {
        success "Git installed successfully"
        return $true
    }

    # Install failed
    return $false
}
```

bin/scoop.ps1 lines 23 & 34 replace:

`Test-GitAvailable` 

with

`([Boolean](Get-HelperPath -Helper Git))
`

z-Fng:
> My apologies, I completely misunderstood what you were trying to achieve, and for my tone.

No worries. We've all seen enough so called ‘AI slop'.

> Which is a noble goal, but why not be more ambitious to actually provide a remediation for the elephant in the room that scoop doesn't install git automatically when needed?

As I mentioned, Scoop supports using external Git installations, so it doesn't install Git automatically. Whether Git should be installed automatically is out of scope for this PR. Furthermore, you'll find that even if Scoop did so, it wouldn't resolve the issue this PR aims to address, **as the problem occurs even during the Git installation itself**.

> image

PR #6600, opened 2026-02-13
move default config file from userprofile\.config to scoop\persist - Update core.ps1

29039:
#### Description
dumping a .config folder in the root user profile is annoying, so I propose moving it to be in scoop\persist\scoop to be consistent with other scoop apps

#### Motivation and Context

Annoyance

Companion to https://github.com/ScoopInstaller/Install/pull/126

tested on my PC works fine
tested to still accept .config as priority if it already exists

#### Checklist:

- [ x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

Sorry am noob

## Summary by CodeRabbit

* **Chores**
  * Enhanced configuration loading to intelligently search multiple paths including legacy and portable configurations. System now checks multiple standard locations in priority order while maintaining full backward compatibility with existing setups, ensuring configurations are correctly located regardless of user environment.

### DIFF
--- a/lib/core.ps1
+++ b/lib/core.ps1
@@ -1260,7 +1260,8 @@ function Out-UTF8File {
 Optimize-SecurityProtocol
 
 # Load Scoop config
-$configHome = $env:XDG_CONFIG_HOME, "$env:USERPROFILE\.config" | Select-Object -First 1
+$legacyConfig = if (Test-Path "$env:USERPROFILE\.config") { "$env:USERPROFILE\.config" }
+$configHome = $env:XDG_CONFIG_HOME, $legacyConfig, "$PSScriptRoot\..\..\..\..\persist" | Select-Object -First 1
 $configFile = "$configHome\scoop\config.json"
 # Check if it's the expected install path for scoop: <root>/apps/scoop/current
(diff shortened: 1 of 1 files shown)


29039:
There are documentation considerations in the following files but maybe it would be excessive to be doing pull requests for each one:

libexec/scoop-checkup.ps1

libexec/scoop-config.ps1

.github/ISSUE_TEMPLATE/Bug_report.md

Also I think that the coderabbit analysis is wrong - | Select-Object -First 1 skips over the null objects, that is how it was before, and also tested this to be the case.

abgox:
- Scoop already supports portable configuration files.
- You just need to move `config.json` into the root directory of scoop.
- However, It can be even better: When installing scoop, directly create `config.json` in the root directory of scoop.
- See ScoopInstaller/Install#119

29039:
Oh great! Can we please make this the default as to not litter the home directory with extra folders which don't serve much purpose?

To be honest, I already find it quite dubious that Scoop by default I installed to ~\scoop rather than %LocalAppData%\scoop like any other user app would do.

Issue #6639, opened 2026-04-17
[Bug] Incorrect version reported by `scoop search`


stevehipwell:
#### Current Behavior

When running `scoop search` the returned version doesn't match the value in the manifest or the version returned when using `scoop info`.

#### Expected Behavior

I'd expect the correct (and matching) version to be returned from all commands.

#### Additional context/output

```shell
~  scoop search copilot-cli
Results from local buckets...

Name        Version Source Binaries
----        ------- ------ --------
copilot-cli 1.0.7   main   copilot


~  scoop info copilot-cli

Name        : copilot-cli
Description : Brings the power of Copilot coding agent directly to your terminal
Version     : 1.0.31
Source      : main
Website     : https://github.com/features/copilot/cli
License     : GitHub Copilot CLI License
Updated at  : 17/04/2026 02:05:02
Updated by  : github-actions[bot]
Binaries    : copilot.exe


~  scoop install copilot-cli
Installing 'copilot-cli' (1.0.31) [64bit] from 'main' bucket
copilot-win32-x64.zip (51.4 MB) [===========================================================================================================] 100%
Checking hash of copilot-win32-x64.zip ... ok.
Extracting copilot-win32-x64.zip ... done.
Linking ~\scoop\apps\copilot-cli\current => ~\scoop\apps\copilot-cli\1.0.31
Creating shim for 'copilot'.
'copilot-cli' (1.0.31) was installed successfully!
```

### System details

**Windows version:** 11

**OS architecture:** 64bit

**PowerShell version:** 7.6.0

**Additional software:** n/a

#### Scoop Configuration

```json
{
  "last_update": "2026-04-17T10:53:19.4155995+01:00",
  "alias": {
    "upgrade": "scoop-upgrade",
    "ls": "scoop-ls"
  },
  "use_lessmsi": true,
  "scoop_branch": "master",
  "scoop_repo": "https://github.com/ScoopInstaller/Scoop",
  "use_sqlite_cache": true,
  "debug": false,
  "aria2-enabled": false,
  "cat_style": "auto",
  "gh_token": "*************"
}
```

z-Fng:
This seems to be a long-standing known issue.
- #6419

PR #6643, opened 2026-04-24
fix(sqlite): select latest search result semantically

### LINKED ISSUE #6419: [Bug] Scoop search won't get correct version when set use_sqlite_cache to true
https://github.com/ScoopInstaller/Scoop/blob/859d1db51bcc840903d5280567846ae2f7207ca2/lib/database.ps1#L256

Maybe the issue is that the SQL query uses simple string sorting with `ORDER BY version DESC`, which causes version numbers to be sorted lexicographically rather than semantically. In lexicographic order, "7.0.9" is indeed greater than "7.0.20" because string comparison is done character by character, and "9" > "2".

#### Current Behavior

```pwsh
❯ scoop search zotero                 
Results from local buckets...

Name        Version    Source   Binaries
----        -------    ------   --------
zotero      7.0.9      extras   zotero
zotero      7.0.9      he0119   zotero
zotero-beta 7.1-beta.8 he0119   zotero
zotero6     6.0.36     versions zotero
```

#### Expected Behavior

The latest zotero version should be 7.0.20.

```pwsh
❯ scoop search zotero                 
Results from local buckets...

Name        Version    Source   Binaries
----        -------    ------   --------
zotero      7.0.20      extras   zotero
zotero      7.0.20      he0119   zotero
zotero-beta 7.1-beta.8 he0119   zotero
zotero6     6.0.36     versions zotero
```

### System details

**Windows version:** [e.g. 7, 8, 10, 11]

11

**OS architecture:** [e.g. 32bit, 64bit, arm64]

64bit

7.5.2

#### Scoop Configuration

```json
{
  "last_update": "2025-07-12T17:13:23.9116762+08:00",
  "aria2-enabled": false,
  "scoop_branch": "master",
  "use_sqlite_cache": true
}
```

B67687:
## Summary
- keep SQLite as the fast search filter, but stop using lexicographic SQL ordering for version selection
- pick the latest matching version in PowerShell with `Compare-Version`
- add regression coverage for semantic version selection in cached search results

## Why
SQLite stores cached manifest versions as plain text, so `ORDER BY version DESC` can prefer `1.0.7` over `1.0.31` and `7.0.9` over `7.0.20` when `use_sqlite_cache` is enabled.

This keeps the cache fast for filtering, while moving the final "which version is latest" decision back to Scoop's existing semantic version logic.

Closes #6419.
Addresses #6639.

## Verification
- manually verified `Select-LatestScoopDBRows` returns `1.0.31` over `1.0.7`
- manually verified it returns `7.0.20` over `7.0.9`
- added focused regression tests in `test/Scoop-Database.Tests.ps1`
- local CI-style Pester run was not fully available here because the repo test script requires modules missing in this environment

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved semantic latest-version selection so package queries consistently return the correct newest semantic version per package and per package+source, increasing package resolution accuracy.

* **Tests**
  * Added automated tests that validate semantic-version selection in grouped (per package+source) and global scenarios to ensure a single correct latest version is chosen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 - **checkver:** Remove redundant always-true condition in GitHub checkver logic ([#6571](https://github.com/ScoopInstaller/Scoop/issues/6571))
 - **core:** Give `dark` higher priority when use `Extract-DarkArchive` ([#6637](https://github.com/ScoopInstaller/Scoop/issues/6637))
 - **checkver:** Harden github checkver ([#6641](https://github.com/ScoopInstaller/Scoop/issues/6641))
+- **scoop-search:** Select latest search result semantically ([#6643](https://github.com/ScoopInstaller/Scoop/issues/6643))
 
 ### Code Refactoring
 
--- a/lib/database.ps1
+++ b/lib/database.ps1
@@ -21,7 +21,7 @@ function Get-LatestScoopDBRow {
     return $latest
 }
 
-function Select-LatestScoopDBRows {
+function Select-LatestScoopDBRow {
     param(
         [Parameter(Mandatory)]
         [System.Data.DataTable]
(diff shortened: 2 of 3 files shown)


niheaven:
Thanks for the jobs! Please rebase the PR to develop branch.

PR #6641, opened 2026-04-24
fix(checkver): Harden github checkver

chawyehsu:
#### Description

**NOTE**: With this change, one has to use `"github":` instead of `"url":` to specify checkver urls in github url format, explicitly. This could be considered a breaking change and is by intention, to narrow down the use of GitHub token to only when the checkver is explicitly set to `github` mode.

#### Motivation and Context

Closes #XXXX

Relates to #XXXX

#### How Has This Been Tested?

Run `checkver.ps1` over local buckets.

#### Checklist:

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).
- [x] I have ensured that I am targeting the `develop` branch.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.
- [x] I have added an entry in the CHANGELOG.

Issue #6526, opened 2025-10-16
[Feature] Checkver - Add function to get redirect URLs that can be used in manifests


o-l-a-v:
#### Is your feature request related to a problem? Please describe.

In manifests for apps that don't have a straightforward way to get current version, `"checkver":{"script":[]}` is often used to get the metadata of some static URL pointing to latest release.

#### Describe the solution you'd like

Instead of having to handle the logic for getting a redirect URL, it'd be useful if Scoop added it as a function that can be referenced from checkver script.

Or maybe a new JSON property `"checkver":{"redirectUrl" = "<url_that_redirects>"}`, which regex later can get metadata out of?

Here is the easiest way I've found to get a redirect URL. Can be made a bit more robust if made into a Scoop function.

```pwsh
[System.Net.HttpWebRequest]::Create(
    '<url>'
).GetResponse().ResponseUri.AbsoluteUri
```

There is some logic for this in `download.ps1` already:

<https://github.com/ScoopInstaller/Scoop/blob/b588a06e41d920d2123ec70aee682bae14935939/lib/download.ps1#L115-L136>

o-l-a-v:
Related: <https://github.com/ScoopInstaller/Extras/issues/16379>

Idea for solution:

* Create an artifact for each Excavator run that holds key pairs:

```json
{
  "<package_name>": <failed_consecutive_excavator_runs>,
  "grafana": 4
}
```

* Each new run fetches the artifact from the previous run.
* If a package that failed previous run also fails this run, bump `<failed_consecutive_excavator_runs>` by one.
* If during a run `<failed_consecutive_excavator_runs>` for a package exceeds a threshold, say 3, create an issue in that repository.
  * One issue per package.
  * This could be a separate workflow running after Excavator.
* If Excavator suddenly works again for a package that got an open issue, close the issue.

PR #6652, opened 2026-05-01
fix(checkver)!: Harden checkver script evaluation

chawyehsu:
#### Description

**This PR is considered to be a breaking change to `checkver.script`.**

---

### `checkver` refinement series

- #6641 
- #6652 this pr **←**
- #6653

---

Prior attempts
- #5407

#### Motivation and Context

Enforce a safer checkver script evaluation in a contranined runtime. Limit side effects by restricting access to outer variables and functions.

Run `checkver.ps1` over local buckets and validate manifest `checkver.script` evaluation.

#### Checklist:

## Summary by CodeRabbit

* **Refactor**
  * Version-check execution now runs in a constrained, isolated runtime to limit side effects and enforce safer evaluation.

* **Bug Fixes**
  * Hardened script evaluation: rejects unsafe constructs, enforces strict error behavior, sanitizes environment, and treats any empty/falsy remote content as failure with clearer output for troubleshooting.

* **Changelog**
  * Added entry documenting the hardening of version-check evaluation.

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -274,6 +274,46 @@ function next($er) {
     Write-Host $er -ForegroundColor DarkRed
 }
 
+function Invoke-SanitizeScriptBlock {
+    param(
+        [Parameter(Mandatory = $true)]
+        [string] $ScriptBlock
+    )
+
+    $tokens = $null
+    $errors = $null
+    $ast = [System.Management.Automation.Language.Parser]::ParseInput($ScriptBlock, [ref]$tokens, [ref]$errors)
+
+    if ($errors.Count -gt 0 -or -not $ast) {
+        $message = if ($errors) { $errors[0].Message } else { 'unknown parse error' }
+        return $true, $message
+    }
+
+    # Restrict the capability of accessing outer variables via `$using:`
+    if ($ast.Find({ param($node) $node -is [System.Management.Automation.Language.UsingExpressionAst] }, $true)) {
+        return $true, "'`$using:' may not be used"
+    }
+
+    return $false, $null
+}
+
+function Get-CheckverScriptPrelude() {
+    return @(
+        '$ErrorActionPreference = "Stop"',
+        '$ProgressPreference = "SilentlyContinue"',
+        # Set allowed env vars, `SystemRoot` is required for irm/iwr
+        '$allowedEnvs = @("SystemRoot","TEMP","TMP")',
+        'Get-ChildItem Env: | Where-Object { $allowedEnvs -notcontains $_.Name } | Remove-Item',
+        # Use `UseBasicParsing` for Invoke-WebRequest by default (WindowsPowerShell)
+        '$PSDefaultParameterValues["Invoke-WebRequest:UseBasicParsing"] = $true',
+        # Use ConstrainedLanguage language mode to restrict checkver.script capabilities
+        '$ExecutionContext.SessionState.LanguageMode = "ConstrainedLanguage"',
+        # Available variables/functions for checkver.script
+        # - `$page`: the retrieved webpage content
+        '$page = $using:page'
+    )
+}
+
 # wait for all to complete
 while ($in_progress -gt 0) {
     $ev = Wait-Event
@@ -326,12 +366,26 @@ while ($in_progress -gt 0) {
             $page = (New-Object System.IO.StreamReader($ms, (Get-Encoding $wc))).ReadToEnd()
         }
 
+        # checkver.script evaluation
         if ($script) {
-            $page = Invoke-Command ([scriptblock]::Create($script -join "`r`n"))
-            $source = 'the output of script'
+            $script = @($script) -join "`n"
+            $isBadScript, $errorMessage = Invoke-SanitizeScriptBlock $script
+            if ($isBadScript) {
+                next "invalid checkver.script: $errorMessage"
+                continue
+            }
+
+            $prelude = Get-CheckverScriptPrelude
+
+            $script = $prelude + $script
+            $scriptBlock = [scriptblock]::Create($script -join "`n")
+
+            $page = [string](Start-Job -ScriptBlock $scriptBlock | Receive-Job -Wait -AutoRemoveJob)
+            debug "Output from evaluating checkver.script:`r`n$page"
+            $source = 'the output of checkver.script'
         }
 
-        if ($null -eq $page) {
+        if (!$page) {
             next "couldn't retrieve content from $source"
             continue
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - **core:** Give `dark` higher priority when use `Extract-DarkArchive` ([#6637](https://github.com/ScoopInstaller/Scoop/issues/6637))
 - **checkver:** Harden github checkver ([#6641](https://github.com/ScoopInstaller/Scoop/issues/6641))
 - **scoop-search:** Select latest search result semantically ([#6643](https://github.com/ScoopInstaller/Scoop/issues/6643))
+- **checkver:** Harden checkver script evaluation ([#6652](https://github.com/ScoopInstaller/Scoop/issues/6652))
 
 ### Code Refactoring
 


bin/checkver.ps1
```diff
@@ -266,6 +266,46 @@ function next($er) {
...
+    return @(
+        '$ErrorActionPreference = "Stop"',
+        '$ProgressPreference = "SilentlyContinue"',
+        # Set allowed env vars, `SystemRoot` is required for irm/iwr
+        '$allowedEnvs = @("SystemRoot","TEMP","TMP")',
+        'Get-ChildItem Env: | Where-Object { $allowedEnvs -notcontains $_.Name } | Remove-Item',
+        # Use `UseBasicParsing` for Invoke-WebRequest by default (WindowsPowerShell)
+        '$PSDefaultParameterValues["Invoke-WebRequest:UseBasicParsing"] = $true',
+        # Use ConstrainedLanguage language mode to restrict checkver.script capabilities
+        '$ExecutionContext.SessionState.LanguageMode = "ConstrainedLanguage"',
```
z-Fng, line 310: Isn't the restriction on `checkver.script` a bit too strict? I'm wondering if it's necessary to be this strict, considering that this script mostly runs in a CI environment. 

It adds  too many constraints. For example, it prevents both redirection-based logic and download-based version fetching. Currently, we have no better way to get the latest version for these two scenarios.

- Redirection-based version fetching:  https://github.com/ScoopInstaller/Extras/blob/4d67fee27762282036867bd0f625206e8d007408/bucket/xmind.json#L27-L32
- Download-based version fetching: https://github.com/ScoopInstaller/Extras/blob/4d67fee27762282036867bd0f625206e8d007408/bucket/spotify.json#L66-L74

bin/checkver.ps1
```diff
@@ -318,12 +358,26 @@ while ($in_progress -gt 0) {
...
+                next "invalid checkver.script: $errorMessage"
+                continue
+            }
+
+            $prelude = Get-CheckverScriptPrelude
+
+            $script = $prelude + $script
+            $scriptBlock = [scriptblock]::Create($script -join "`n")
+
+            $page = (Start-Job -ScriptBlock $scriptBlock | Receive-Job -Wait -AutoRemoveJob | Out-String).Trim()
```
z-Fng: Just a note, `OutString` changes the output format, so many scripts will need to rewrite the output script.
```
.\bin\checkver.ps1 -App elisa
Output from evaluating checkver.script:
26.04-4413
msvc2022
elisa: couldn't match '([\d.-]+) (?<lib>[\w]+)' in the output of checkver.script
```
```
.\bin\checkver.ps1 -App blender
Output from evaluating checkver.script:
Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      1      -1
blender: couldn't match '(\d+\.\d+\.\d+)' in the output of checkver.script
```

bin/checkver.ps1
chawyehsu: > Redirection-based version fetching: https://github.com/ScoopInstaller/Extras/blob/4d67fee27762282036867bd0f625206e8d007408/bucket/xmind.json#L27-L32

This can to be updated to use `Invoke-WebRequest` to get HTTP redirect URLs, e.g. [extras/rtss](https://github.com/ScoopInstaller/Extras/blob/cd8562f54c9bac4afe3af7c8f57e953bbe1881bc/bucket/rtss.json#L51-L56).

> Download-based version fetching: https://github.com/ScoopInstaller/Extras/blob/4d67fee27762282036867bd0f625206e8d007408/bucket/spotify.json#L66-L74

They should put the temporary files to `$env:TEMP`/`$env:TMP` (which is why these env vars are allowed in the implementation) instead of relying on the internal `cache_path` function and placing files expected to be in `$env:TEMP` to Scoop's cache dir.



> Isn't the restriction on checkver.script a bit too strict? I'm wondering if it's necessary to be this strict, considering that this script mostly runs in a CI environment.

From what I can tell, these implementations are all necessary.

z-Fng: > > Redirection-based version fetching: https://github.com/ScoopInstaller/Extras/blob/4d67fee27762282036867bd0f625206e8d007408/bucket/xmind.json#L27-L32
> 
> This can to be updated to use `Invoke-WebRequest` to get HTTP redirect URLs, e.g. [extras/rtss](https://github.com/ScoopInstaller/Extras/blob/cd8562f54c9bac4afe3af7c8f57e953bbe1881bc/bucket/rtss.json#L51-L56).

Yes, but having to support both PowerShell 5 and PowerShell 7 makes it more cumbersome (https://github.com/ScoopInstaller/Extras/pull/17052#issuecomment-3772494730) — not as clean as using HttpWebRequest.

But it's OK, maybe we can add a feature to get redirect URLs later (https://github.com/ScoopInstaller/Scoop/issues/6526).

> > Download-based version fetching: https://github.com/ScoopInstaller/Extras/blob/4d67fee27762282036867bd0f625206e8d007408/bucket/spotify.json#L66-L74
> 
> They should put the temporary files to `$env:TEMP`/`$env:TMP` (which is why these env vars are allowed in the implementation) instead of relying on the internal `cache_path` function and placing files expected to be in `$env:TEMP` to Scoop's cache dir.

You're right. I missed that point.

bin/checkver.ps1
```diff
@@ -318,12 +358,26 @@ while ($in_progress -gt 0) {
...
+            Write-Verbose "Output from evaluating checkver.script:`r`n$page"
```
z-Fng: Just to confirm, should we use `debug` rather than `Write-Verbose` for consistency?

bin/checkver.ps1
z-Fng: > From what I can tell, these implementations are all necessary.

Could you share the reasons?

The examples mentioned above are just common scenarios.

Is it possible to add an option that enables restrictions by default, but allows them to be disabled via configuration (both globally and in the manifest)? There may be still some special cases where workarounds have been implemented on checkver, and in such cases, without an option to disable restrictions, we would have to deprecate these manifests as they cannot be updated. e.g., https://github.com/ScoopInstaller/Extras/blob/e066b75375e385789965bc64a9c90cfca7e496be/bucket/qq-nt.json#L35

Additionally, should access to GitHub tokens also be permitted? There may be situations where a token is needed to access the GitHub API multiple times within the script.

bin/checkver.ps1
chawyehsu: I've noticed that when testing the implementation. The reason is that we want to ensure the ouptut from evaluating checkver.script, the `$page` variable to be `[string]` consistently rather than arbitrary types, as in subsequent operations on `$page`, it's expected to be `[string]`. But `Out-String` is indeed not robust, and I will try other possible solutions.

bin/checkver.ps1
chawyehsu: > Could you share the reasons?

Sorry, I'm unable to disclose more on this.

> Is it possible to add an option that enables restrictions by default, but allows them to be disabled via configuration

That would render the change meaningless.

> Additionally, should access to GitHub tokens also be permitted? There may be situations where a token is needed to access the GitHub API multiple times within the script.

I searched the entire org and found only [3 manifests](https://github.com/search?q=org%3AScoopInstaller+Get-GitHubToken+language%3AJSON&type=code&l=JSON) that explicitly use GitHub Tokens in the scriptblock. Is this truly an absolutely necessary action without alternatives? This involves the misuse of internal functions, which is a key focus of this PR and its subsequent refinements.

chawyehsu: > Yes, but having to support both PowerShell 5 and PowerShell 7 makes it more cumbersome (https://github.com/ScoopInstaller/Extras/pull/17052#issuecomment-3772494730) — not as clean as using HttpWebRequest.
> 
> But it's OK, maybe we can add a feature to get redirect URLs later (https://github.com/ScoopInstaller/Scoop/issues/6526).

Actually in my dev branch I've made a attempt to [provide an allowed builtin function](https://github.com/chawyehsu/scoop/commit/125437b48959364d275fdcf834d1d3ad1d670d03#diff-38a2fa93638321564fd5495333c30dafb0d9430bae53cd72109c24d10596437dR354) to get redirect URLs. This attempt is feasible, but it doesn't seem concise or intuitive enough to me, so it wasn't included in this PR. I believe we will have a more complete solution for #6526.

z-Fng: > > Is it possible to add an option that enables restrictions by default, but allows them to be disabled via configuration
> 
> That would render the change meaningless.

If we impose such strict restrictions without providing a way to disable them, we'll likely lose a great deal of flexibility. Once this change gets merged, some community buckets may consider bypassing these restrictions, and we'll only end up restricting ourselves. Some manifests may even be deprecated simply because they can no longer fetch the latest versions. I think just limiting this behavior to the official bucket should be fine. I've also consistently advocated for reducing the use of internal functions. Additionally, using the built-in GitHub Actions token instead of a PAT isn't so risky, is it?

> > Additionally, should access to GitHub tokens also be permitted? There may be situations where a token is needed to access the GitHub API multiple times within the script.
> 
> I searched the entire org and found only [3 manifests](https://github.com/search?q=org%3AScoopInstaller+Get-GitHubToken+language%3AJSON&type=code&l=JSON) that explicitly use GitHub Tokens in the scriptblock. Is this truly an absolutely necessary action without alternatives? This involves the misuse of internal functions, which is a key focus of this PR and its subsequent refinements.

If only a single GitHub API request is needed for version detection, we can place the URL in `checkver.github` and use the returned result via `$page`. However, if multiple requests are required, this approach won't work.

Perhaps we could ask the other maintainers for their thoughts?

z-Fng: I realize my previous comment might have sounded a bit too harsh -- I didn't mean to criticize this PR at all. The change is well-intentioned, and I completely agree that prioritizing security is essential. It's just that this could potentially be a breaking change, so we'll need a little more time to align/adjust.

ghost:
Ready to review this PR? Stage has broken it down into 3 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Define script sanitization and execution environment](https://stagereview.app/ScoopInstaller/Scoop/pull/6652/chapters/1) |
| 2 | [Execute checkver scripts in isolated jobs](https://stagereview.app/ScoopInstaller/Scoop/pull/6652/chapters/2) |
| 3 | [Document hardening changes in CHANGELOG](https://stagereview.app/ScoopInstaller/Scoop/pull/6652/chapters/3) |

<a href="https://stagereview.app/ScoopInstaller/Scoop/pull/6652">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 26e0ccbb68cd9e89c678e83c3f8e9743332c49c7 on Jun 22, 2026 9:16am UTC.</sub>

Issue #6559, opened 2025-11-22
[Feature] useGithubAPI by default when using github checkver


Lutra-Fs:
#### Is your feature request related to a problem? Please describe.
Scoop's GitHub checkver currently has inconsistent behavior that relies on fragile HTML scraping. When using the
  GitHub matcher with additional properties like regex, it falls back to parsing HTML pages instead of using the GitHub
   API, leading to:

  - Brittle regex patterns that match against HTML URLs instead of clean version strings
  - Performance issues due to downloading full HTML pages vs compact JSON API responses
  - Maintenance burden as manifest authors must understand GitHub's HTML structure
  - Reliability problems when GitHub changes their HTML layout, breaking existing manifests

#### Describe the solution you'd like
Modify Scoop's checkver logic to use GitHub API by default when using the GitHub matcher, regardless of additional
  properties. The proposed solution would:

  1. Enable `jsonpath` with GitHub matcher: Allow `jsonpath` to work with `github` property
  2. Default to API mode: Use GitHub API for all GitHub matcher usage, not just simple cases
  3. Simplify regex patterns: Extract version directly from tag_name field, then clean with simple regex

Examples of improved manifest structure:
```
  // Current (HTML scraping)
  "checkver": {
      "github": "https://github.com/wez/atomicparsley",
      "regex": "/releases/tag/(\\d+\\.\\d+\\.[a-f0-9]+)"
  }

  // Proposed (API + jsonpath)
  "checkver": {
      "github": "https://github.com/wez/atomicparsley",
      "jsonpath": "$.tag_name",
      "regex": "([\\d]+\\.[\\d]+\\.[\\d]+\\.[a-f0-9]+)"
  }

  // Simple cases remains the same
  "checkver": {
      "github": "https://github.com/owner/repo"
  }
```

We can also share the result between checkver and hash extraction, since it is using github API for github assets now.
#### Describe alternatives you've considered
One option could be introducing more controls within the checkver property itself. However, in most cases, it's primarily used to match a tag, so additional flexibility may not offer substantial benefit.

Lutra-Fs:
cc: @z-Fng @stevehipwell

stevehipwell:
@Lutra-Fs I've been thinking about this same problem and came to a very similar conclusion. I think my only concern would be backwards compatibility, I was thinking of a new primary key like `githubjson`. But with your proposal the initial implementation could keep the existing behaviour if a regex but no jsonpath was specified?

[...]

PR #6653, opened 2026-05-01
feat(checkver)!: Set GitHub mode default jsonpath and regex

chawyehsu:
#### Description

Set GitHub mode default jsonpath and regex

**This PR is considered to be a breaking change to `checkver.script`** as the default behavior of matching `regex`/`jsonpath` is changed as new default `jsonpath`/`regex` introduced.

---

### `checkver` refinement series

- #6641 
- #6652
- #6653 this pr **←**

#### Motivation and Context

Relates to #6559

Run `checkver.ps1` over local buckets.

#### Checklist:

## Summary by CodeRabbit

* **Refactor**
  * Improved version-check behavior so GitHub-based checks choose mode-appropriate defaults, yielding more consistent tag detection between API and web modes.

* **Documentation**
  * Added an "Unreleased" changelog entry describing the new default extraction behavior for GitHub-based version detection.

### DIFF
--- a/bin/checkver.ps1
+++ b/bin/checkver.ps1
@@ -122,23 +122,21 @@ $Queue | ForEach-Object {
     }
 
     # Not Specified
+    $url = $json.homepage
+    $regex = ''
+    $jsonpath = ''
+    $xpath = ''
+    $replace = ''
+
     if ($json.checkver.url) {
         $url = $json.checkver.url
-    } else {
-        $url = $json.homepage
     }
 
     if ($json.checkver.re) {
         $regex = $json.checkver.re
     } elseif ($json.checkver.regex) {
         $regex = $json.checkver.regex
-    } else {
-        $regex = ''
     }
-
-    $jsonpath = ''
-    $xpath = ''
-    $replace = ''
 
     ## GitHub
     #
@@ -156,7 +154,6 @@ $Queue | ForEach-Object {
     # ```
     if (($json.checkver -eq 'github') -or $json.checkver.github) {
         $githubUrlPattern = '^https://((www\.)?github\.com/[\w.-]+/[\w.-]+/?|api\.github\.com/repos/[\w.-]+/[\w.-]+/.+)$'
-        $regex = if ($regex) { $regex } else { '/releases/tag/(?:v|V)?([\d.]+)' }
 
         $inputGithubUrl = $json.homepage
         $fieldUsed = 'homepage'
@@ -178,6 +175,17 @@ $Queue | ForEach-Object {
         if ($GitHubToken) {
             $url = $url -replace '//(www\.)?github\.com/', '//api.github.com/repos/'
             $wc.Headers.Add('Authorization', "token $GitHubToken")
+        }
+
+        if ($url -like 'https://api.github.com*') {
+            # Default jsonpath/regex in GitHub API mode
+            $jsonpath = '$.tag_name'
+            if (-not $regex) {
+                $regex = '(?:v|V)?([\d.-]+)'
+            }
+        } elseif (-not $regex) {
+            # Default regex in GitHub HTML mode
+            $regex = '/releases/tag/(?:v|V)?([\d.-]+)'
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - **install:** Add separator at the end of notes, highlight suggestions ([#6418](https://github.com/ScoopInstaller/Scoop/issues/6418))
 - **download|scoop-download:** Add GitHub issue prompt when the default downloader fails ([#6539](https://github.com/ScoopInstaller/Scoop/issues/6539))
 - **download|scoop-config:** Allow disabling automatic fallback to the default downloader when Aria2c download fails ([#6538](https://github.com/ScoopInstaller/Scoop/issues/6538))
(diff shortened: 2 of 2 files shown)


bin/checkver.ps1
```diff
@@ -179,6 +176,17 @@ $Queue | ForEach-Object {
             $url = $url -replace '//(www\.)?github\.com/', '//api.github.com/repos/'
```
z-Fng, line 176: LGTM. Just to confirm, a few questions:
1. Do we need to support HTTP as well?
2. (This might be outside the scope) Is it limited to `api.github.com/repos/<owner>/<repo>` only? Should we also support other request types like `api.github.com/repositories/<id>`? ref: https://github.com/ScoopInstaller/Extras/pull/17382
3. Could we add debug some logs so it's clearer what's happening?
https://github.com/ScoopInstaller/Scoop/blob/b588a06e41d920d2123ec70aee682bae14935939/lib/core.ps1#L297-L325
```
debug $state
```

chawyehsu: > Do we need to support HTTP as well?

No reason to support HTTP.

> Should we also support other request types like api.github.com/repositories/<id>? 

It depends. I don't see the necessity currently. GitHub API already handles those scenarios.

> debug $state

Sure. Could be a separate PR for debuggability improvement

z-Fng: > 2\. (This might be outside the scope) Is it limited to `api.github.com/repos/<owner>/<repo>` only? Should we also support other request types like `api.github.com/repositories/<id>`?

> It depends. I don't see the necessity currently. GitHub API already handles those scenarios.

It appears I've referenced the wrong place. My intended question is: might this regular expression (`api\.github\.com/repos/[\w.-]+/[\w.-]+/.+`) be overly restrictive to match GitHub API URLs? It cannot match `api.github.com/repositories/<id>`.
https://github.com/ScoopInstaller/Scoop/blob/a7a7d340f948df68e42b02148f00b7bc61cd572b/bin/checkver.ps1#L158
https://github.com/ScoopInstaller/Scoop/blob/a7a7d340f948df68e42b02148f00b7bc61cd572b/bin/checkver.ps1#L168-L171

chawyehsu: I believe it already fit 99% of the needs, what's the reason for using `api.github.com/repositories/<id>`? It's not intuitive to use repo id.

z-Fng: > I believe it already fit 99% of the needs, what's the reason for using `api.github.com/repositories/<id>`? It's not intuitive to use repo id.

I agree. https://github.com/ScoopInstaller/Extras/pull/16758#issuecomment-3886568256

This was only suggested to maintain compatibility with the previous approach — there may still be some in the community bucket (https://github.com/ScoopInstaller/Extras/pull/16758#issuecomment-3642693909), but official bucket has already been fully migrated (https://github.com/ScoopInstaller/Extras/pull/17382).

z-Fng: I was wondering if we could merge this change soon, so that the main/extras/versions/nonportable buckets can make the corresponding changes to get `checkver` working again. There may be hundreds of manifests that need to be adjusted — and it's not as simple as just replacing a string. Many manifests currently have broken checkver status because the previous PR (#6641) actually changed the regex rules as well.

https://github.com/ScoopInstaller/Extras/pull/17746#issue-4371594469
> Some `checkver` may still not work. However, since there is still another breaking change (ScoopInstaller/Scoop#6653) that hasn't been merged yet, we may wait until it is merged before fixing this issue to avoid duplicating work.

chawyehsu: > I was wondering if we could merge this change soon

I'm thinking of adjusting the default regex for matching tags. I think adding the `-` symbol would be a good idea, I noticed that many version numbers use `-` after reviewing manifests, e.g. [imagemagick](https://github.com/ScoopInstaller/Main/blob/master/bucket/imagemagick.json).

bin/checkver.ps1
```diff
@@ -179,6 +176,17 @@ $Queue | ForEach-Object {
...
             $wc.Headers.Add('Authorization', "token $GitHubToken")
         }
+
+        if ($url -like 'https://api.github.com*') {
+            # Default jsonpath/regex in GitHub API mode
+            $jsonpath = '$.tag_name'
+            if (-not $regex) {
+                $regex = '(?:v|V)?([\d.-]+)'
+            }
```
z-Fng, line 185: Ran into a new issue: JSONPath is applied to the script output if `checkver.github` exists.

https://github.com/ScoopInstaller/Versions/blob/1f6e422157834a5177e9bd48d2398d45f96b5ca5/bucket/goneovim-nightly.json#L28-L37

```console
PS> .\bin\checkver.ps1 -App goneovim-nightly
ConvertFrom-Json: C:\Users\runneradmin\SCOOP\apps\scoop\current\lib\json.ps1:130
Line |
 130 |      $result = $json | ConvertFrom-Json -ea stop
     |                        ~~~~~~~~~~~~~~~~~~~~~~~~~
     | Conversion from JSON failed with error: Input string '20260203-4c89b4976f48723d5915e8f3a94376e5a0318c9f' is not
     | a valid number. Path '', line 1, position 49.
goneovim-nightly: couldn't find '$.tag_name' in the output of script
```

z-Fng: `checkver.script` must output JSON or it'll error out. e.g., https://github.com/ScoopInstaller/Versions/commit/138d6dad1e639264b654167d7ebcbbae8302c35a

Issue #6569, opened 2025-12-15
[Feature] Improve Search Ranking to Prioritize Core Packages (e.g., git)


doggy8088:
When I search for `git` on the website, the actual package only shows up on the second page.

<img width="1795" height="1300" alt="Image" src="https://github.com/user-attachments/assets/fdaed487-0bd9-418f-a8dc-48885c16ea07" />

When I use `scoop search git`, the real `git` package is listed as the 21st result.

<img width="962" height="900" alt="Image" src="https://github.com/user-attachments/assets/72fed644-7df3-415d-b7b7-55d82c0b1522" />

I'd like to suggest tweaking the search engine's ranking so that crucial packages appear sooner.

z-Fng:
> When I search for `git` on the website, the actual package only shows up on the second page.

We recently [optimized](https://github.com/ScoopInstaller/scoopinstaller.github.io-indexer/commit/ceb2a242464718a6f5602b260dbe0753a73cf7e5) the text weighting on our website's search service to [improve](https://scoop.sh/#/apps?q=git) the search experience.
<img width="1450" height="445" alt="Image" src="https://github.com/user-attachments/assets/facf18a1-ec41-408b-baed-f88de52954f9" />

Issue #6472, opened 2025-08-29
[Bug] Severe PowerShell 5.1 Performance Degradation When use_sqlite_cache is Enabled


I-AM-A-NOOB:
#### Current Behavior

After setting `scoop config use_sqlite_cache true` on Windows 10 22H2 with PowerShell 5.1, every PowerShell 5.1 instance experiences severe input and command execution latency (e.g., entering ls and waiting an abnormally long time for output to slowly appear line by line). The issue persists across process/System reboots and Windows updates. It is resolved only by setting `use_sqlite_cache` back to `false`. I installed Powershell 7  after this and it does not exhibit this problem.

#### Expected Behavior

Enabling the SQLite cache without causing significant degradation in the general responsiveness and command execution speed of the PowerShell host itself.

#### Additional context/output

I am using a modified version of scoop (with enhanced proxy support for faster GitHub connectivity in China), but I **suspect** this issue might originate from the original scoop mainline or Windows PowerShell 5.1.

The problem occurs specifically in PowerShell 5.1, but not in PowerShell 7.

**I have limited knowledge on this issue and couldn't find any existing information about it.**

#### Possible Solution

~`scoop config use_sqlite_cache false`~
N/A

System details
Windows 10 22H2 x64

PowerShell version: 5.1 (Issue occurs here). Also tested on PowerShell 7 (No issue).

Additional software: N/A

Scoop Configuration

```json
{
  "last_update": "2025-08-29T22:15:10.1237951+08:00",
  "scoop_repo": "https://gitee.com/scoop-installer/scoop",
  "scoop_branch": "master",
  "use_sqlite_cache": true,
  "aria2-warning-enabled": false
}
```
*similar configuration from another PC*

PR #6657, opened 2026-05-07
feat(search): Add binary-index cache for fast substring search

B67687:
## Summary

Adds a JSON-based binary-index cache to `scoop-search.ps1` for the non-SQLite code path. Three new functions, \~170 lines added, 2 files changed.

## Problem

Without `use_sqlite_cache`, `scoop search` reads and parses every manifest JSON on every invocation (\~3000ms). The SQLite cache path is faster (\~300ms) but requires a separate database dependency and isn't always enabled.

Additionally, the original code mutated `$query` from string to Regex mid-function, causing `search_remotes` to receive a Regex object (previously relied on `.ToString()` coincidence).

## How it works

```
scoop search <query>
  +-- SQLite cache enabled? -> SQLite DB (unchanged)
  +-- Has regex metacharacters? -> regex full-scan (unchanged)
  +-- Cache fresh? -> in-memory search (~200ms)
  +-- Cache stale/missing? -> regex full-scan -> build cache -> next time fast
```

## Performance

| Scenario | Before | After | Speedup |
|----------|--------|-------|---------|
| Literal query (warm) | \~3000ms | \~200ms | **15-25x** |
| Regex query | \~3000ms | \~3000ms | unchanged |
| Cache build (first run) | — | \~2500ms | one-time |

## Testing

10 queries verified identical to stock: google(9), firefox(9), git(86), node(28), python(40), vscode(10), eventstore(1), 7zip(5), curl(6), zig(3).

## Checklist

- [ ] I have updated the documentation accordingly. ***(Not applicable: internal search change, no user-facing docs)***
- [ ] I have updated the tests accordingly. ***(Not applicable: existing CI tests pass, cache path doesn't affect output)***

## Related
- #6569 
- #6472 

## Summary by CodeRabbit

* **New Features**
  * Indexed substring search with case-insensitive bin/app lookups and automatic fallback to a full scan for complex queries; cache is rebuilt as needed to keep results current.

* **Performance Improvements**
  * Local non-SQLite searches now use a JSON-based index for much faster substring queries (approx 15–20×) when the cache is fresh; stale caches trigger refreshes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@
 ### Performance Improvements
 
 - **shim:** Update kiennq-shim to v3.1.2 ([#6261](https://github.com/ScoopInstaller/Scoop/issues/6261))
+- **scoop-search:** Add JSON-based binary-index cache for 15-20x faster non-SQLite search ([#6657](https://github.com/ScoopInstaller/Scoop/pull/6657))
 
 ## [v0.5.2](https://github.com/ScoopInstaller/Scoop/compare/v0.5.1...v0.5.2) - 2024-07-26
 
--- a/libexec/scoop-search.ps1
+++ b/libexec/scoop-search.ps1
@@ -1,10 +1,11 @@
 # Usage: scoop search <query>
 # Summary: Search available apps
 # Help: Searches for apps that are available to install.
-#
-# If used with [query], shows app names that match the query.
 #   - With 'use_sqlite_cache' enabled, [query] is partially matched against app names, binaries, and shortcuts.
-#   - Without 'use_sqlite_cache', [query] can be a regular expression to match against app names and binaries.
+#   - Without 'use_sqlite_cache', [query] is matched against app names and binaries via:
+#       * A JSON-based binary-index cache (when fresh) for fast substring matching (~200ms), or
+#       * The original regex full-scan as fallback (~3s).
+#   - Queries containing regex metacharacters (.+*?|[](){}^$\ ) skip the cache and use regex directly.
 # Without [query], shows all the available apps.
 param($query)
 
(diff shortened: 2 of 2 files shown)


B67687:
## Architecture overview for reviewers

Three new functions inserted before `bin_match` (line 18), one else-block modified (line 310). Everything else is untouched.

### New functions

| Function | Lines | Purpose |
|----------|-------|---------|
| `init_search_cache` | 23-55 | Loads/staleness-checks the cache file. Returns `$true` if usable. |
| `build_search_cache` | 57-108 | Scans all manifests, builds JSON cache at `$scoopdir/search-cache.json`. |
| `search_by_index` | 110-173 | Zero-I/O search of cached app/binary names via `IndexOf`. Appends to `$list`. |

### Modified else-block (lines 310-339)

```
if regex metacharacters in query? → skip cache, do regex full-scan (original code, unchanged)
else if cache fresh? → search_by_index (fast)
if list still empty? → regex full-scan (original code, unchanged)
if cache wasn't fresh → build_search_cache
```

This is the only control-flow change. The original regex path runs identically when the cache is stale or a regex query is used.

### Key design decisions (not obvious from diff)

- **Staleness uses `Get-ChildItem -Recurse` instead of a lighter check** (line 31-34). Must match `build_search_cache` and `search_bucket` enumeration exactly — using a different method (`GetFiles`, non-recursive) could produce inconsistent counts for edge cases like symlinks. The 150ms overhead is acceptable for a 15-25x speedup.
- **Binary display matches stock** (lines 128-173): name-matched apps get `Binaries = ''`; only binary-only matches show binaries. Stock uses the same if/else pattern.
- **Array storage for cross-bucket duplicates** (line 76-77): if `git` exists in both `main` and `extras`, the cache stores both paths. Without this, removing one bucket's manifest silently drops the app until the next staleness-triggered rebuild.
- **`$query`/`$regex` separation** (line 322): original code mutated `$query` from string to `Regex` object, then passed it to `search_remotes` which expected a string. Fixed by using `$regex` for the compiled object and leaving `$query` as the original string.

### What to test if you want to verify manually

```powershell
# 1. Cold build (~3s first run)
scoop search node

# 2. Warm (should be instant)
scoop search node
scoop search git
scoop search "git|lfs"  # regex → cache skipped, uses regex path

# 3. Staleness: touch any manifest, re-search (should rebuild)
notepad $scoopdir\buckets\main\bucket\git.json  # add a space, save
scoop search git  # detects changed LastWriteTime, rebuilds cache

# 4. Remote search (unchanged, same as before)
scoop search nonexistent-app-that-exists-on-github
```

### Files NOT touched

`lib/buckets.ps1`, `lib/core.ps1`, `lib/database.ps1`, any other `libexec/` script. The change is confined to `scoop-search.ps1` + `CHANGELOG.md`.

B67687:
Closing due to inactivity. These PRs have been open for 3-4 months with no maintainer engagement. Happy to reopen if there's interest.

Issue #6029, opened 2024-06-24
Updating pwsh fails, because scoop always uses pwsh during install


lynx-xdg:
#### Current Behavior
While trying to update pwsh, the install fails because running the install script uses an instance of pwsh.

#### Expected Behavior
When updating pwsh.exe scoop should use powershell.exe instead of pwsh.exe.

#### Additional context/output
The shim for scoop will switches to pwsh.exe, even when run by powershell.exe.

#### Source of the problem
The current scoop shim:
```powershell
@rem C:\Users\rikvd\scoop\apps\scoop\current\bin\scoop.ps1
@echo off
where /q pwsh.exe
if %errorlevel% equ 0 (
    pwsh -noprofile -ex unrestricted -file "C:\Users\rikvd\scoop\apps\scoop\current\bin\scoop.ps1"  %*
) else (
    powershell -noprofile -ex unrestricted -file "C:\Users\rikvd\scoop\apps\scoop\current\bin\scoop.ps1"  %*
)
```

### System details

**Windows version:** 11

**OS architecture:** 64bit

**PowerShell version:** 
``` 
Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      22621  3672    
```

**Additional software:** `pwsh.exe` (installed via scoop) and `powershell.exe`

#### Scoop Configuration

```json
{
  "last_update": "2024-06-24T18:51:28.9137540+02:00",
  "scoop_branch": "master",
  "scoop_repo": "https://github.com/ScoopInstaller/Scoop"
}
```

lynx-xdg:
I've got it working by modifying `shims/scoop.cmd` to always use `powershell` and only saw the error message after the install succeeded. The proposed 'fix' in the error message also doesn't work, since the shim will get used anyways, thus using `pwsh` again.

niheaven:
Use Windows PowerShell while update pwsh.

jcotton42:
@niheaven why does Scoop try to use PS7 anyhow? It seems it has to support both 5.1 and 7 anyhow, so why not just always use 5 to avoid this issue?

At the very least, I think scoop should output a warning when this situation occurs, because this also happens when running `scoop update` from `cmd` (which I do since it starts a little faster than WinPS).

PR #6661, opened 2026-05-22
feat(scoop-update): add --allow-running flag to bypass running-process check

### LINKED ISSUE #6342: [Bug] Cannot update `pwsh` because Scoop uses `pwsh` to update
#### Current Behavior

`pwsh` is not updatable, because Scoop is using `pwsh` and cannot replace the `pwsh` binary, while it is in use. This devious circle blocks updating `pwsh` entirely.

#### Expected Behavior

`pwsh` is updated fine.

* Installing `pwsh` obviously works fine.
* I do not consider this a bug specific to the package, as it inherently is related to how Scoop itself operates.

#### Possible Solution

Add CLI option to specify path to executable of PowerShell of choice.

### System details

**Windows version:** [e.g. 7, 8, 10, 11]

11

**OS architecture:** [e.g. 32bit, 64bit, arm64]

amd64

```
Major  Minor  Patch  PreReleaseLabel BuildLabel
-----  -----  -----  --------------- ----------
7      5      0
```

#### Scoop Configuration

```json
{
  "last_update": "2025-04-24T11:30:12.2985148+00:00",
  "scoop_branch": "master"
}
```

B67687:
## Description

Adds `-r, --allow-running` flag to `scoop update` that temporarily sets `IGNORE_RUNNING_PROCESSES $true` for the duration of the update command, then restores the previous value.

This is safe for apps installed in versioned directories (e.g., pwsh, most portable apps), where the update creates a new version directory without touching the running binary. The existing `IGNORE_RUNNING_PROCESSES` config already handles this case — this flag simply exposes it as a one-shot CLI option so users don't need to toggle a global config.

```powershell
# Before: requires manual config toggle
scoop config IGNORE_RUNNING_PROCESSES $true
scoop update pwsh
scoop config IGNORE_RUNNING_PROCESSES $false

# After: single command
scoop update pwsh --allow-running
```

## Motivation and Context

Updating pwsh (or any scoop-managed app that is currently running) has required users to either:
1. Close all instances of the app first
2. Set the global `IGNORE_RUNNING_PROCESSES` config, update, then unset it

The scoop pwsh manifest itself notes: "Since Scoop uses pwsh.exe internally, to update PowerShell Core itself, run `scoop update pwsh` from Windows PowerShell." But even from Windows PowerShell, the running-process check can still block the update because the current pwsh.exe session matches the app directory.

For versioned-install apps like pwsh, the update is safe because scoop creates a new version directory (e.g., `7.6.2/`) without modifying the old directory (`7.6.1/`) where the running binary lives. The user merely needs to restart the app.

Closes #6342
Relates to #6029, #4713, #5799

## How Has This Been Tested?

- Verified that `scoop update pwsh` with `IGNORE_RUNNING_PROCESSES` set correctly creates a new version directory (`7.6.2/`) while the old binary (`7.6.1/`) continues running
- Verified that the `current` junction is updated to point to the new version
- Verified that the old `--skip-hash-check` and other flags still work alongside `--allow-running`
- Tested both single-app and bulk update paths (the config is set before the update loop and restored afterward)

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - **download|scoop-config:** Allow disabling automatic fallback to the default downloader when Aria2c download fails ([#6538](https://github.com/ScoopInstaller/Scoop/issues/6538))
 - **checkver:** Set GitHub mode default jsonpath and regex ([#6653](https://github.com/ScoopInstaller/Scoop/issues/6653))
 - **download:** Enhance Get-GitHubToken to include more token sources ([#6712](https://github.com/ScoopInstaller/Scoop/issues/6712))
+- **scoop-update:** Add `--allow-running` flag to bypass running-process check ([#6342](https://github.com/ScoopInstaller/Scoop/issues/6342), [#6029](https://github.com/ScoopInstaller/Scoop/issues/6029))
 
 ### Bug Fixes
 
--- a/libexec/scoop-update.ps1
+++ b/libexec/scoop-update.ps1
@@ -38,6 +38,7 @@ $use_cache = !($opt.k -or $opt.'no-cache')
 $quiet = $opt.q -or $opt.quiet
 $independent = $opt.i -or $opt.independent
 $all = $opt.a -or $opt.all
+$allowRunning = $opt.r -or $opt.'allow-running'
 
 # load config
 $configRepo = get_config SCOOP_REPO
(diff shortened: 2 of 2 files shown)


B67687:
Closing due to inactivity. These PRs have been open for 3-4 months with no maintainer engagement. Happy to reopen if there's interest.

PR #6712, opened 2026-08-10
feat(download): Enhance Get-GitHubToken to include more token sources

lisonge:
https://cli.github.com/manual/gh_help_environment

```text
> gh --help environment | grep '_TOKEN'
`GH_TOKEN`, `GITHUB_TOKEN` (in order of precedence): an authentication token that will be used when
`GH_ENTERPRISE_TOKEN`, `GITHUB_ENTERPRISE_TOKEN` (in order of precedence): an authentication
authenticated with, the stored credentials will be used. Otherwise, setting `GH_TOKEN` or
`GH_ENTERPRISE_TOKEN` is required, depending on the targeted host.
```

This pull request makes a small update to the `Get-GitHubToken` function in `lib/download.ps1` to improve how GitHub authentication tokens are retrieved.

- The function now checks two additional environment variables, `GH_TOKEN` and `GITHUB_TOKEN`, in addition to the previously checked variables, to find a valid GitHub token for authentication.


gitfool:
Better yet, check for and run `gh auth token`, as was suggested in https://github.com/ScoopInstaller/Scoop/issues/6716, which would be similar to what homebrew does.

niheaven:
Please target to develop branch

refack:
> Better yet, check for and run `gh auth token`, as was suggested in #6716, which would be similar to what homebrew does.

see #6725 
😊

